PR #6, opened 2026-01-07
feat: add public skills list endpoint

Dimillian:
Adds a public /api/skills endpoint (batch + limit) for highlighted/latest lists and exposes it via the schema routes.

Notes:
- Uses the existing skills.list query and returns lightweight fields for browsing.
- Schema dist artifacts were not regenerated in this PR.

### DIFF
--- a/convex/http.ts
+++ b/convex/http.ts
@@ -10,6 +10,7 @@ import {
   cliUploadUrlHttp,
   cliWhoamiHttp,
   getSkillHttp,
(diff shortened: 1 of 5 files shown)


steipete:
Shipped v1 public API on main (list/search/skill/version/file/download + publish/delete + OpenAPI at /api/v1/openapi.json). Closing this PR in favor of the new /api/v1 surface.

PR #12, opened 2026-01-17
fix: search mode navigation and state management

NACC96:
## Summary
- Fix "Explore search" button causing page refresh by using URL params instead of navigating to `/search`
- Enable `/search` URL deep linking via `beforeLoad` redirect
- Fix logo click not closing search mode by properly syncing state with URL params

## Root Cause
Search mode was managed as local React state that wasn't properly synced with URL params. The URL should be the source of truth for navigation state.

## Test plan
- [ ] Click "Explore search" button → should open search mode without page refresh
- [ ] Navigate directly to `/search` → should redirect to `/` with search mode open
- [ ] Click ClawdHub logo while in search mode → should close search and show landing page

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 ### Added
 - CLI: add `explore` command for latest updates, with limit clamping + tests/docs (thanks @jdrhyne, #14).
 
+### Fixed
+- Web: keep search mode navigation and state in sync (thanks @NACC96, #12).
+
 ## 0.2.0 - 2026-01-13
 
 ### Added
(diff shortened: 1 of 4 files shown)


NACC96:
Fixed in 060bdf9 - OnlyCrabsHome now preserves the `search` flag when `searchMode` is on (matching SkillsHome behavior).

steipete:
Landed via temp rebase onto main.\n\n- Gate: bun run lint && bun run build && bun run test\n- Land commit: 49e9c3c071888c5fc4b0a79789f2a83707aa75af\n- Merge commit: eaaa5e442387b876ad54ce1b0c20c8a260058446\n\nThanks @NACC96!

PR #40, opened 2026-01-26
fix: add rate limiting to download counter to prevent inflation

orlyjamie:
Summary                                                                                                     
                                                                                                              
  - Add IP-based rate limiting (5 downloads per skill per IP per hour) to prevent download count inflation    
  - Only trust cf-connecting-ip header - x-forwarded-for is trivially spoofable                               
  - Still serve files to rate-limited users, just don't count inflated downloads                              
                                                                                                              
  The Vulnerability                                                                                           
                                                                                                              
  The download endpoint at convex/downloads.ts:56 incremented download counts with no rate limiting. An       
  attacker could inflate any skill's download count by:                                                       
                                                                                                              
```
  for i in {1..1000}; do                                                                                      
    curl -H "X-Forwarded-For: 1.2.3.$i" \                                                                     
      "https://api.clawdhub.com/download?slug=target-skill" -o /dev/null                                      
  done    
```                                                                                                    
                                                                                                              
  Important Note                                                                                              
                                                                                                              
  Download counts are fundamentally gameable as trust metrics because downloads are anonymous. Even with    
  rate limiting, attackers can use proxies/VPNs/Tor.                                                          
                                                                                                              
  Recommendation: De-emphasize downloads in the UI. Stars and installs are better trust signals because they  
  require authenticated sessions.  

Reference: https://x.com/theonejvo/status/2015892980851474595

### DIFF
--- a/convex/downloads.ts
+++ b/convex/downloads.ts
@@ -1,8 +1,27 @@
 import { v } from 'convex/values'
 import { zipSync } from 'fflate'
-import { api } from './_generated/api'
+import { api, internal } from './_generated/api'
 import { httpAction, mutation } from './_generated/server'
 import { applySkillStatDeltas, bumpDailySkillStats } from './lib/skillStats'
+
+// Rate limit: 5 downloads per skill per IP per hour
+// NOTE: This is defense-in-depth only. Download counts are fundamentally ungameable
+// as a trust metric because:
+//   1. Downloads are anonymous (no auth required)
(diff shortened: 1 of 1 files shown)


steipete:
Superseded by #278, which is merged on `main` (merge commit `03cd710abc7754f613b975df594f1a69c2222fb4`).

That PR implemented the download-inflation mitigation with tests and docs (hourly unique identity counting using API-token user id fallback to IP). Closing this one to avoid duplicate/conflicting changes.

PR #74, opened 2026-01-31
fix(http): disable HTTP/2 to fix undici decompression issues

shamashel:
Fixes openclaw/openclaw#3637

## Summary

When undici fetches compressed content over HTTP/2, it fails to decompress the response, causing JSON parse errors:
```
✖ Unexpected token '',"..."... is not valid JSON
```

This change disables HTTP/2 (`allowH2: false`) to force HTTP/1.1 which handles compression correctly.

## Changes
- `packages/clawdhub/src/http.ts`: Changed `allowH2: true` → `allowH2: false`

## Testing
- [x] Verify clawhub CLI works with compressed responses
- [x] Check that skill search/install operations succeed

## Context

The undici HTTP/2 implementation does not properly decompress response bodies when servers return compressed content over HTTP/2. This causes the CLI to receive binary data instead of valid JSON.

See: https://github.com/openclaw/openclaw/issues/3637

### DIFF
--- a/e2e/clawdhub.e2e.test.ts
+++ b/e2e/clawdhub.e2e.test.ts
@@ -20,7 +20,7 @@ const REQUEST_TIMEOUT_MS = 15_000
 try {
   setGlobalDispatcher(
     new Agent({
-      allowH2: true,
+      allowH2: false,
       connect: { timeout: REQUEST_TIMEOUT_MS },
     }),
   )


steipete:
Follow-up (previous comment lost some inline code formatting due to shell backticks):

Closing as obsolete.

- main already removed the Undici allowH2 override (CHANGELOG.md -> 0.6.1 - 2026-02-13 -> Fixed -> "HTTP: remove allowH2 from Undici agent to prevent fetch failed on Node.js 22+").
- Undici v7 default is allowH2=false, so the effective fix is already in place by not setting it.
- PR was merge-conflicting; rebasing to land a redundant change doesn’t seem worth it.

If the original issue still reproduces on CLI >= 0.6.1 (2026-02-13), suggest: minimal repro + regression test around compressed JSON responses (HTTP/1.1 + HTTP/2).

Issue #156, opened 2026-02-06
Downloads & Star metrics not working


mastersyondgy:
Hi, seems starring skills is not working and when downloading skills using the CLI or the Download button on Clawhub website, it is not updating the metrics on that skill.

PR #176, opened 2026-02-08
fix: reduce skill stat processing interval from 15m to 1m

tanujbhaud:
This improves the responsiveness of download and star metrics on the website, addressing user feedback that these metrics were not updating in a timely manner (Issue #156).

<h2>Greptile Overview</h2>

<h3>Greptile Summary</h3>

This PR updates the Convex cron schedule for skill-stat processing to run much more frequently (from 15 minutes to 1 minute) to make download/star metrics update more responsively on the site (Issue #156). The change is localized to the Convex cron configuration (`convex/crons.ts`).

<h3>Confidence Score: 3/5</h3>

- This PR is low-risk if the cron syntax is correct, but as written it likely won’t run at all.
- Only one file changes and the intent is straightforward, but the new schedule string appears to use a non-cron duration format that typical Convex cron scheduling won’t accept, which would prevent the job from registering/executing.
- convex/crons.ts

<sub>(4/5) You can add custom instructions or style guidelines for the agent [here](https://app.greptile.com/review/github)!</sub>

**Context used:**

### DIFF
--- a/convex/crons.ts
+++ b/convex/crons.ts
@@ -24,9 +24,9 @@ crons.interval(
   { batchSize: 200, maxBatches: 5 },
 )
 
-crons.schedule(
+crons.interval(
   'skill-stat-events',
-  '*/1 * * * *',
+  { minutes: 1 },
   internal.skillStatEvents.processSkillStatEventsAction,
   {},
 )


tanujbhaud:
Closing this PR — reducing the cron interval from 15m to 1m doesn't address the root cause of issue #156 (downloads & star metrics not working at all). The problem is in the event recording/processing pipeline, not the scheduler frequency. This change would only increase Convex function invocations 15x without fixing the underlying bug.

PR #325, opened 2026-02-15
update

ianalloway:
<h3>Greptile Summary</h3>

This PR modifies the GitHub profile sync logic in the auth callback and the `syncGitHubProfile` action, but both files contain **duplicate code blocks** that appear to be the result of an incomplete merge or rebase — new code was inserted without removing the original logic it was intended to replace.

- **`convex/auth.ts`**: `handleDeletedUserSignIn` is called twice per sign-in, and `syncGitHubProfileAction` is scheduled unconditionally (bypassing the 6-hour rate-limit window) in addition to the existing conditional scheduling. Every sign-in will hit the GitHub API unnecessarily.
- **`convex/lib/githubAccount.ts`**: A new early-update block calls `syncGitHubProfileInternal` without the required `syncedAt` argument, which will cause a **runtime validation error** in Convex. The original correct call below it is still present, making this a duplicate that also breaks execution.

<h3>Confidence Score: 1/5</h3>

- This PR introduces a runtime-breaking validation error and duplicate logic that should not be merged as-is.
- The missing required `syncedAt` argument in `githubAccount.ts` will cause a Convex validation error at runtime whenever a user's GitHub username has changed, breaking the profile sync feature. Additionally, the duplicate `handleDeletedUserSignIn` calls and unconditional scheduler invocations in `auth.ts` bypass the intended rate-limiting, causing unnecessary GitHub API calls on every sign-in.
- Both `convex/auth.ts` and `convex/lib/githubAccount.ts` need attention — the duplicate code blocks in each file need to be resolved by removing either the old or new version of the logic.

<sub>Last reviewed commit: 331ba0d</sub>

**Context used:**

### DIFF
--- a/convex/users.ts
+++ b/convex/users.ts
@@ -76,26 +76,6 @@ export const syncGitHubProfileInternal = internalMutation({
     const user = await ctx.db.get(args.userId)
     if (!user || user.deletedAt || user.deactivatedAt) return
 
-    const updates: Record<string, unknown> = {
-      name: args.name,
-      updatedAt: Date.now(),
-    }
-
-    // Update handle if it was derived from the old username
-    if (user.handle === user.name) {
-      updates.handle = args.name
-    }
-
-    // Update displayName if it was derived from the old username
-    if (user.displayName === user.name || user.displayName === user.handle) {
-      updates.displayName = args.name
-    }
-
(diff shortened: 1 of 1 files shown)


steipete:
seems incomplete/unworkable

PR #588, opened 2026-03-03
feat(ui): add install command snippet and copy functionality to Skill…

moonrailgun:
…DetailPage

add this section which help people quick install

image


steipete:
Thanks. Closing this one.

We already have install UX on the skill page, and this adds a second, separate command surface. It also promotes `clawhub install <slug>`, while the current docs and UI prefer first-run-safe `npx` / `pnpm dlx` / `bunx` flows.

If we want copy-to-clipboard here, it should come from the existing canonical install component rather than a new card.

PR #640, opened 2026-03-09
Fix Convex env access in appMeta:getDeploymentInfo

agent-jackmax:
Fixes server error on ClawHub due to `process` being undefined in Convex query runtime. Uses safe env read via globalThis.process?.env.

### DIFF
--- a/convex/appMeta.ts
+++ b/convex/appMeta.ts
@@ -5,10 +5,17 @@ function normalizeEnv(value: string | undefined) {
   return normalized ? normalized : null
 }
 
+function readEnv(key: string): string | undefined {
+  // Convex functions don't always run in a Node.js environment.
+  // Accessing `process.env` directly can throw `ReferenceError: process is not defined`.
+  const env = (globalThis as any)?.process?.env as Record<string, string | undefined> | undefined
+  return env?.[key]
+}
+
 export const getDeploymentInfo = query({
   args: {},
   handler: async () => ({
-    appBuildSha: normalizeEnv(process.env.APP_BUILD_SHA),
-    deployedAt: normalizeEnv(process.env.APP_DEPLOYED_AT),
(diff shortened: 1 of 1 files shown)


agent-jackmax:
Repro: https://clawhub.ai and https://clawhub.ai/skills show 'Something went wrong'.\n\nBrowser console error:\n- [CONVEX Q(appMeta:getDeploymentInfo)] Server Error Called by client\n\nDirect Convex repro (returns 500):\n\n```bash\ncurl -s -X POST https://wry-manatee-359.convex.cloud/api/query \\n  -H 'Content-Type: application/json' \\n  -H 'Convex-Client: npm-1.31.7' \\n  --data '{"path":"appMeta:getDeploymentInfo","format":"convex_encoded_json","args":[{}]}'\n```\n\nRoot cause: Convex query runtime may not define Node's `process`, so `process.env.*` throws (ReferenceError).\n\nFix in this PR: read env via `globalThis.process?.env` so the query returns nulls instead of 500 when env isn't present.

ngutman:
Closing this as not needed.

The issue this PR is trying to address was not a durable `process is not defined` problem in `convex/appMeta.ts`. The real fix was upstream:

- `2be9b67` added the deploy pipeline that stamps `APP_BUILD_SHA` and `APP_DEPLOYED_AT` into Convex.
- `0aa702f` and `4d72506` hardened the deployment drift banner so a missing/failing query does not take down the app shell.

On current `main`, `appMeta:getDeploymentInfo` still uses `process.env`, and the live production query is returning real values, so the existing code path is working. This PR would just paper over env/config issues by returning `null`, which disables useful drift detection instead of fixing the underlying deployment contract.

Because the underlying issue has already been addressed upstream, there isn’t anything left to merge here. Thanks for digging into it.

PR #830, opened 2026-03-13
fix: align security scan summary text

geoffrey-xiao:
## Summary

Fix the security scan summary text alignment in the skill detail page.

## What Changed

- set the expandable security scan header to use full width
- left-align the header content
- left-align the wrapped summary text inside the OpenClaw analysis row

## Why

The security scan description could appear centered when the summary wrapped across multiple lines, which made the panel harder to scan and visually inconsistent with the rest of the detail page. This change keeps the description left-aligned without changing the interaction or adding truncation behavior.

## Files Changed

- `src/styles.css`

before:
image

after:
image

### DIFF
--- a/src/styles.css
+++ b/src/styles.css
@@ -3866,9 +3866,11 @@ html.theme-transition::view-transition-new(theme) {
   align-items: center;
   justify-content: space-between;
   gap: 12px;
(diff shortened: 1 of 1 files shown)


ImLukeF:

PR #833, opened 2026-03-13
fix: scroll bar in the fie view under light theme

geoffrey-xiao:
## Summary

This PR fixes two UI issues in the skill file viewer:

- corrects native scrollbar rendering in light theme
- aligns the file list and file preview panes to use the same max height

## Changes

- set the light theme `color-scheme` to `light` so browsers do not render dark native scrollbars while the app is in light mode
- keep dark mode controlled by the existing `[data-theme="dark"]` override
- introduce a shared file browser pane max-height token in `src/styles.css`
- apply that shared max height to both the left file list and the right file preview pane
- remove the redundant dark-theme override for the shared pane height token

before
image

after
image

## Files changed

- `src/styles.css`

## Notes

- no component logic changes included
- this is CSS-only

### DIFF
--- a/src/styles.css
+++ b/src/styles.css
@@ -4,7 +4,7 @@
 @import "@fontsource/ibm-plex-mono/latin.css";
 
 :root {
-  color-scheme: light dark;
+  color-scheme: light;
   --bg: #f8f2ed;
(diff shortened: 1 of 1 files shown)


ImLukeF:

PR #837, opened 2026-03-13
perf: split global stats recount into batched action

sethconvex:
## Summary
- The daily `updateGlobalStatsInternal` cron read all ~19K `skillSearchDigest` docs (17.8 MB) in a single mutation, exceeding the Convex bytes-read limit (3 errors in 72h)
- Switches to `updateGlobalStatsAction` — an action that pages through the table in ~1000-doc queries (~900 KB each), then writes the result in a separate mutation
- Keeps the old mutation as a deprecated fallback

## Context
Delta tracking via `adjustGlobalPublicSkillsCount` is already wired into all 17 skill mutation paths, so the count stays current in real-time. This cron is just a daily reconciliation — it just needs to not blow the read limit.

## Test plan
- [x] Pre-existing test failures on main (8 in search.test.ts / listPublicPageV2.test.ts) confirmed unrelated
- [x] `npx convex dev --once` typechecks (pre-existing type errors in search.ts/skills.ts unrelated)
- [ ] Deploy and verify cron runs without errors in insights

### DIFF
--- a/convex/lib/skillSearchDigest.ts
+++ b/convex/lib/skillSearchDigest.ts
@@ -68,7 +68,7 @@ export function digestToHydratableSkill(digest: Doc<'skillSearchDigest'>): Hydra
   }
 }
 
-/** Insert or update the digest row for a skill. */
+/** Insert or update the digest row for a skill. Skips the write when no fields changed. */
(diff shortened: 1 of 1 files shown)


magicseth:
@codex review

PR #903, opened 2026-03-15
fix: remove global stats full-table scan and shard rate limit counters

sethconvex:
## Summary

- **Remove `countPublicSkillsForGlobalStats`** which scanned all 19K `skillSearchDigest` rows (17MB) in a single mutation, exceeding the bytes-read limit. The `adjustGlobalPublicSkillsCount` fallback now initializes with `Math.max(0, delta)` — the daily cron reconciles within 24h.
- **Shard rate limit writes across 8 keys** to reduce OCC contention (3M retries / 43K permanent failures in 72h). The read-only query sums all shards + legacy key; the mutation writes to a random shard, reducing collision probability to 1/8.
- **Delete deprecated `updateGlobalStatsInternal`** mutation to prevent accidental manual invocations of the full-table scan.

## Test plan

- [x] `npx convex dev --once` — typechecks pass
- [ ] Deploy to prod: `npx convex deploy --yes`
- [ ] After deploy: `npx convex insights --prod --details` — verify OCC retries drop ~8×, no new errors
- [ ] Verify rate limiting still works (API requests not incorrectly rejected)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -87,7 +87,7 @@ bunx convex run --no-push devSeed:seedNixSkills
 bunx convex run --no-push devSeedExtra:seedExtraSkillsInternal
 
 # Refresh the cached skills count (required after seeding)
-bunx convex run --no-push statsMaintenance:updateGlobalStatsInternal
+bunx convex run --no-push statsMaintenance:updateGlobalStatsAction
 ```
 
 To reset and re-seed:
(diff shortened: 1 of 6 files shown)


vincentkoc:
Thanks for the patch. I landed the safe part directly on main as a narrower maintainer fix in 33334c5: the scan-heavy public skill count fallback is gone and local seed docs now use the paginated stats action.

I did not take the stale rate-limit sharding changes from this PR because that changes limiter enforcement semantics and needs a separate focused design/review.

PR #956, opened 2026-03-17
feat: switch /skills and homepage to listPublicPageV4

sethconvex:
## Summary
- Switch `/skills` browse page and homepage popular section from V3 to V4
- V4 uses `convex-helpers` `getPage()` for deterministic, cacheable index-key cursors
- Two users requesting the same page produce identical query args, enabling shared query caching
- Remove `useV4` flag and V3 fallback code — V4 is the only code path
- Guard against `hasMore=true` with `nextCursor=null` edge case

## Test plan
- [ ] Visit `/skills` — verify skills load, pagination, sort/filter/search all work
- [ ] Visit homepage — verify popular skills section loads
- [ ] Test pagination: load multiple pages, change sort, verify no duplicates

### DIFF
--- a/convex/lib/access.test.ts
+++ b/convex/lib/access.test.ts
@@ -2,6 +2,7 @@ import { describe, expect, it, vi } from 'vitest'
 
 vi.mock('@convex-dev/auth/server', () => ({
   getAuthUserId: vi.fn(),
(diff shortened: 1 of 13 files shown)


sethconvex:
@codex review

PR #963, opened 2026-03-17
perf: use digest for GitHub backup sync

sethconvex:
## Summary

- **`getGitHubBackupPageInternal`** now paginates `skillSearchDigest` (~800B/doc) instead of `skills` (~3-5KB) with `ctx.db.get()` joins to `skillVersions` and `users`
- Owner handle, version string, and publishedAt come directly from the digest — zero joins during the scan phase
- Version `files` are only fetched lazily via `getVersionByIdInternal` when a skill actually needs backing up (version mismatch with GitHub), which is rare in steady-state
- Stale cursor from the old `skills` table is handled gracefully (resets to page 1)

**Before:** 3 table reads per skill (skills + skillVersions + users) = ~7KB per row
**After:** 1 table read per skill (skillSearchDigest) = ~800B per row, ~9x reduction

## Test plan

- [x] `npx convex dev --once` typechecks pass
- [x] `npx vitest run convex/githubBackups` — 3 tests pass
- [ ] Deploy and verify backup sync cron runs without errors
- [ ] Verify bandwidth reduction in Convex dashboard

### DIFF
--- a/convex/githubBackups.test.ts
+++ b/convex/githubBackups.test.ts
@@ -4,64 +4,61 @@ import { getGitHubBackupPageInternal } from './githubBackups'
 const handler = (getGitHubBackupPageInternal as unknown as { _handler: Function })._handler
 
 describe('githubBackups page filtering', () => {
-  it('skips non-public skills (soft-deleted, hidden, removed)', async () => {
-    const activeSkill = {
-      _id: 'skills:active',
+  it('skips non-public digests (soft-deleted, hidden, removed)', async () => {
+    const activeDigest = {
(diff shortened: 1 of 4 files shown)


sethconvex:
@codex review

vincentkoc:
Thanks for the patch. I landed the useful direction directly on main in cf20e10 with an updated narrow implementation: GitHub backup sync now pages over `skillSearchDigest`, only loads version files when a backup is actually needed, and resets stale old cursors from the previous `skills` scan.

Closing this conflicted PR as superseded.

PR #977, opened 2026-03-17
fix: install card text overflow style

geoffrey-xiao:
## Summary

Fixes the install card layout on skill detail pages.

## Changes

- replace generic install entry markup with dedicated install-card structure
- improve spacing and typography for install entries
- make install commands wrap cleanly inside the card
- keep install metadata styling consistent with the surrounding panels

before
PixPin_2026-03-17_20-06-16

after
PixPin_2026-03-17_20-07-52

## Files changed

- `src/components/SkillInstallCard.tsx`
- `src/styles.css`

## Notes

- style-only change
- no backend or behavior changes


ImLukeF:

PR #981, opened 2026-03-17
refactor: skill card with tag

geoffrey-xiao:
## Summary

Moves skill card tags below the title to improve card hierarchy and readability in the skills grid.

## What Changed

- Updated [src/components/SkillCard.tsx](/Users/geoffrey/Desktop/AI/clawhub/src/components/SkillCard.tsx) so the skill title renders before the tag row.
- Kept the existing tag styling and behavior unchanged; this is a layout-only refactor.

before
image

after
image

## Why

The previous layout put badges/platform tags above the title, which made card scanning less natural. Rendering the title first gives the card a clearer reading order:

1. title
2. tags
3. summary
4. owner/stats

This works better for both promotional badges like `Highlighted` and technical tags like OS labels.

### DIFF
--- a/src/components/SkillCard.tsx
+++ b/src/components/SkillCard.tsx
@@ -20,6 +20,7 @@ export function SkillCard({ skill, badge, chip, platformLabels, summaryFallback,
 
   return (
     <Link to={link} className="card skill-card">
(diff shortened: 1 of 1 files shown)


ImLukeF:

PR #1029, opened 2026-03-19
Add owner skill delete flow

neeravmakwana:
## Summary
- add an owner-facing soft-delete action to the existing skill owner tools panel
- require users to click delete first, then type `delete` before enabling the destructive confirm action
- reuse the existing Convex soft-delete behavior on the backend and cover the new UI flow with a focused test

## Test plan
- [x] `bun run test src/__tests__/skill-detail-page.test.tsx`
- [x] `bun run lint`
- [ ] Sign in as the skill owner on a deployed preview and verify `Delete skill` appears under owner tools
- [ ] Click `Delete skill`, confirm the text box only appears after the click, and verify `Confirm delete` stays disabled until `delete` is entered
- [ ] Confirm the skill is removed from the dashboard/listing after deletion

Made with [Cursor](https://cursor.com)

### DIFF
--- a/convex/skills.ts
+++ b/convex/skills.ts
@@ -4911,6 +4911,16 @@ export const mergeOwnedSkillIntoCanonical = mutation({
   },
 })
 
+export const deleteOwnedSkill = mutation({
+  args: {
+    slug: v.string(),
+  },
+  handler: async (ctx, args) => {
+    const { user } = await requireUser(ctx)
+    return setSkillSoftDeletedByActor(ctx, user._id, args.slug, true)
+  },
+})
+
 export const renameOwnedSkillInternal = internalMutation({
   args: {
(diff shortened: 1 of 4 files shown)


neeravmakwana:
Addressed the review feedback in follow-up commit `0705786`.

Changes made:
- updated the post-delete redirect to `navigate({ to: '/dashboard', replace: true })` so the deleted skill URL is replaced in history instead of remaining on the back stack
- removed the redundant `if (isSubmitting) return` guard from the delete cancel handler, since the button is already disabled while submitting
- updated the focused test assertion to expect the `replace: true` navigation call

Validation re-run:
- `bun run test src/__tests__/skill-detail-page.test.tsx`
- `bun run lint`

PR #1130, opened 2026-03-22
fix(search): boost exact slug match over containing slug

QuinnH496:
## Bug

When searching for `self-improving-agent`, the exact match does not rank #1. Instead, longer slugs like `xiucheng-self-improving-agent` appear first.

## Root Cause

`getLexicalBoost()` uses `matchesAllTokens()` to check if every query token appears in the slug tokens. For query "self-improving-agent" (tokens: `["self", "improving", "agent"]`), both:
- `self-improving-agent` → tokens match → **1.4 boost**
- `xiucheng-self-improving-agent` → tokens match → **1.4 boost**

Both get the same `SLUG_EXACT_BOOST`, so the vector similarity score (which favors the longer, more "semantically similar" slug) determines ranking.

## Fix

Added `EXACT_SLUG_BOOST = 2.5` that fires when the entire normalized query equals the slug exactly. This gives exact matches a decisive scoring edge.

```typescript
const normalizedQuery = queryTokens.join("-");
if (normalizedQuery === slug) {
  boost += EXACT_SLUG_BOOST;  // 2.5 — highest possible lexical boost
} else if (matchesAllTokens(...)) {
  boost += SLUG_EXACT_BOOST;  // 1.4 — tokens match but slug is longer
}
```

## Testing

- All 17 tests pass (16 existing + 1 new)
- New test: `boosts exact full slug over slug that merely contains all query tokens`

## Files Changed

- `convex/search.ts` — Add `EXACT_SLUG_BOOST` constant and exact slug check in `getLexicalBoost()`
- `convex/search.test.ts` — Add regression test

Fixes #52034

### DIFF
--- a/convex/search.test.ts
+++ b/convex/search.test.ts
@@ -359,6 +359,26 @@ describe("search helpers", () => {
     expect(exactScore).toBeGreaterThan(looseScore);
   });
 
+  it("boosts exact full slug over slug that merely contains all query tokens", () => {
+    const queryTokens = tokenize("self-improving-agent");
+    const exactScore = __test.scoreSkillResult(
+      queryTokens,
+      0.5,
+      "Self Improving Agent",
+      "self-improving-agent",
+      10,
+    );
+    const containingScore = __test.scoreSkillResult(
+      queryTokens,
+      0.6,
+      "Self Improving Agent",
+      "xiucheng-self-improving-agent",
+      100,
+    );
+    // Exact slug should beat a longer slug even with lower vector score and fewer downloads
+    expect(exactScore).toBeGreaterThan(containingScore);
+  });
+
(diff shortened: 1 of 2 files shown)


steipete:
Landed this on `main` in 7c5b8b2 (`fix(search): boost exact slug matches`).

What landed:
- Added a dedicated full-slug boost before token-level slug matching.
- Renamed the existing token-level slug boost for clarity.
- Added regression coverage proving `self-improving-agent` outranks `xiucheng-self-improving-agent` even with lower vector score and fewer downloads.
- Changelog credits this PR and @QuinnH496.

Verified locally:
- `bun run test`
- `bun run coverage`
- `bun run lint`
- `bun run build`

Closing this PR as superseded by the landed commit. Thanks for the clear fix and test case.

Issue #1182, opened 2026-03-23
Publish/Import fails: Convex pagination error in syncSkillSearchDigestsForOwnerPublisherId


ivo-gos:
Attempting to publish v1.4.2 of battlecard-competitive-intelligence fails across all three methods (CLI, Upload, Import). Error: "This query or mutation function ran multiple paginated queries. Convex only supports a single paginated query in each function." Stack trace points to syncSkillSearchDigestsForOwnerPublisherId. Skill has 8 existing versions which may be triggering the pagination limit. Account: @ivo-gos

vincentkoc:
Thanks for the report. This was the same backend failure as #1195: owner-publisher digest rebuilds could hit Convex's single-paginated-query limit.

Fixed on `main` in 9b27c1a by changing the package and skill owner-publisher digest jobs to process one page per internal mutation and continue via scheduled cursor handoff.

Validation run:
- `bun run format:check -- convex/functions.ts convex/functions.test.ts convex/tsconfig.json convex/lib/httpRateLimit.ts`
- `bun run test -- convex/functions.test.ts`
- `bun run lint`
- `bunx tsc -p convex/tsconfig.json --noEmit`
- `bunx tsc -p packages/schema/tsconfig.json --noEmit`
- `bunx tsc -p packages/clawhub/tsconfig.json --noEmit`
- `bunx tsc --noEmit`

This will take effect in production after the next manual deploy from `main`.

Issue #1179, opened 2026-03-23
clawhub publish + Web UI: Server Error on users:ensure and ensurePersonalPublisherForUser (first-time publisher)


rasimme:
First-time publisher. Both CLI `clawhub publish` and the Web UI fail with Convex server errors. Profile cannot be saved, skill cannot be published.

## CLI Error
clawhub publish ./my-skill --slug my-skill --name "My Skill" --version 1.0.0

Error: This query or mutation function ran multiple paginated queries.
Convex only supports a single paginated query in each function.
at syncSkillSearchDigestsForOwnerPublisherId (../../convex/functions.ts:223:17)
at ensurePersonalPublisherForUser (../../convex/lib/publishers.ts:169:14)

## Web UI Console Errors (clawhub.com/settings)
[CONVEX M(users:ensure)] [Request ID: 7a1983ae7adb28a5] Server Error
[CONVEX M(users:updateProfile)] [Request ID: 8e03b7fb6f37fa2b] Server Error
[CONVEX M(users:updateProfile)] [Request ID: f9a2005527fd2c39] Server Error

Saving display name on settings page does nothing — no error shown in UI, but console shows the above.

## Account
- GitHub-linked, dashboard shows correct avatar + info
- `clawhub whoami` → ✔ (shows "unknown" — likely because profile can't be saved)
- No prior publishes

## Environment
- OpenClaw: v2026.3.22
- ClawHub CLI: bundled with OpenClaw
- OS: Linux arm64

## Notes
Possibly related to #477 and #614. The root cause appears to be `ensurePersonalPublisherForUser` calling `syncSkillSearchDigestsForOwnerPublisherId` which runs multiple paginated Convex queries in one mutation — violating Convex's single-pagination constraint.

steipete:
Thanks for the detailed repro, and sorry this blocked publishing.

This report was hitting the same root cause as the Convex `multiple paginated queries` publish failures: owner-publisher digest sync ran pagination inside the publish/first-publisher mutation path. Current main/prod code now pages one batch and schedules continuation work instead, so the `ensurePersonalPublisherForUser` path no longer trips Convex's single-pagination limit.

I’m closing this duplicate as resolved. If a current publish still hits the same Convex error, please open a fresh issue with the current CLI version, account, slug, and timestamp.

Issue #1176, opened 2026-03-23
publish/sync fails for all skills: "multiple paginated queries" Convex error


hubtiger123:
## Bug

All `clawhub publish` and `clawhub sync` operations fail with the following error for account `hubtiger123`:

```
Uncaught Error: This query or mutation function ran multiple paginated queries. Convex only supports a single paginated query in each function.
    at async syncSkillSearchDigestsForOwnerPublisherId (../../convex/functions.ts:223:17)
    at async ensurePersonalPublisherForUser (../../convex/lib/publishers.ts:169:14)
```

## Steps to reproduce

```bash
clawhub publish skills/qstar-video-ecom --version 1.0.7 --changelog "..."
# or
clawhub sync --all
```

Both commands fail immediately at the "Preparing" step with the same error.

## Impact

Unable to publish any skill updates. Have a bug fix pending for an actively used skill (`qstar-video-ecom`) that cannot be shipped.

## Environment

- clawhub CLI v0.7.0
- Account: hubtiger123
- Started failing: 2026-03-23

steipete:
Thanks for the detailed repro, and sorry this blocked publishing.

This root cause is fixed in current main/prod code. `syncSkillSearchDigestsForOwnerPublisherId` / owner-publisher digest sync now pages one batch at a time and schedules continuation work instead of running multiple paginated queries inside the same Convex function. That removes the Convex `multiple paginated queries` crash from the publish/first-publisher path.

Code proof checked now in `convex/functions.ts`; related shipped fix is `fix(convex): page owner-publisher digest syncs`. I’m closing this duplicate report as resolved. If a current publish still hits the same Convex error, please open a fresh issue with the new CLI version, account, slug, and timestamp.

Issue #1175, opened 2026-03-23
Publish fails: Convex pagination error in ensurePersonalPublisherForUser


msaleme:
## Bug

`clawhub publish` consistently fails with a server-side Convex pagination error when publishing an updated skill version.

## Steps to Reproduce

```bash
npx clawhub auth whoami  # confirms: msaleme
npx clawhub publish /path/to/skill --slug agent-security-harness --version 3.3.0
```

## Error

```
- Preparing agent-security-harness@3.3.0
✖ Uncaught Error: This query or mutation function ran multiple paginated queries.
  Convex only supports a single paginated query in each function.
    at async syncSkillSearchDigestsForOwnerPublisherId (../../convex/functions.ts:223:17)
    at async <anonymous> (../../convex/functions.ts:339:1)
    ...
    at async ensurePersonalPublisherForUser (../../convex/lib/publishers.ts:169:14)
```

## Context

- Skill `agent-security-harness` was originally published as v3.2.0 earlier today (2026-03-23 ~01:11 UTC) successfully.
- Attempting to publish v3.3.0 with updated SKILL.md content (added AIUC-1 crosswalk section).
- SKILL.md parses correctly (valid YAML frontmatter, confirmed with Python yaml.safe_load).
- Tried multiple times over ~10 minutes, same error every time.
- `clawhub` version: 0.9.0 (via npx)

## Expected

v3.3.0 publishes successfully, updating the listing at clawhub.ai/msaleme/agent-security-harness.

steipete:
Thanks for the detailed repro, and sorry this blocked publishing.

This root cause is fixed in current main/prod code. `syncSkillSearchDigestsForOwnerPublisherId` / owner-publisher digest sync now pages one batch at a time and schedules continuation work instead of running multiple paginated queries inside the same Convex function. That removes the Convex `multiple paginated queries` crash from the publish/first-publisher path.

Code proof checked now in `convex/functions.ts`; related shipped fix is `fix(convex): page owner-publisher digest syncs`. I’m closing this duplicate report as resolved. If a current publish still hits the same Convex error, please open a fresh issue with the new CLI version, account, slug, and timestamp.

PR #1183, opened 2026-03-23
fix(convex): replace paginate with collect to fix multi-pagination error

### LINKED ISSUE #1170: ClawHub Cli publish bug
CLI version: `0.7.0`, `0.9.0`

## Issue
Cli is throwing error while publishing skill. I have checked cli is logged in `clawhub whomai` is working.

`
✖ Uncaught Error: This query or mutation function ran multiple paginated queries. Convex only supports a single paginated query in each function.
    at async syncSkillSearchDigestsForOwnerPublisherId (../../convex/functions.ts:223:17)
    at async <anonymous> (../../convex/functions.ts:339:1)
    at async <anonymous> (../../node_modules/.bun/convex-helpers@0.1.114+9cc8149e1e4a55ae/node_modules/convex-helpers/server/triggers.js:208:13)
    at async <anonymous> (../../node_modules/.bun/convex-helpers@0.1.114+9cc8149e1e4a55ae/node_modules/convex-helpers/server/triggers.js:224:12)
    at async withLock [as withLock] (../../node_modules/.bun/convex-helpers@0.1.114+9cc8149e1e4a55ae/node_modules/convex-helpers/server/triggers.js:39:8)
    at async _execThenTrigger (../../node_modules/.bun/convex-helpers@0.1.114+9cc8149e1e4a55ae/node_modules/convex-helpers/server/triggers.js:219:12)
    at async insert [as insert] (../../node_modules/.bun/convex-helpers@0.1.114+9cc8149e1e4a55ae/node_modules/convex-helpers/server/triggers.js:175:59)
    at async ensurePersonalPublisherForUser (../../convex/lib/publishers.ts:169:14)
`

imwyvern:
## Problem

All publish operations fail with:

```
Error: This query or mutation function ran multiple paginated queries.
```

The `publishers` trigger in `convex/functions.ts` (line 337-340) calls both:
1. `syncPackageSearchDigestsForOwnerPublisherId` — uses `paginate()`
2. `syncSkillSearchDigestsForOwnerPublisherId` — uses `paginate()`

Convex only allows **one** paginated query per mutation, so the second call always fails.

## Solution

Replace `paginate()` loops with `collect()` in both functions. This is safe because:
- A single publisher's packages/skills are bounded (typically <100)
- `collect()` stays well within Convex's document limits for this use case
- No behavioral change — same documents are processed, just fetched in one call instead of pages

## Changes

- `convex/functions.ts`: Replace paginate loops with collect in both `syncPackageSearchDigestsForOwnerPublisherId` and `syncSkillSearchDigestsForOwnerPublisherId`

## Testing

- Manually verified the logic change is equivalent (same index, same iteration)
- The fix resolves the root cause for all 7 related issues

Fixes #1170, #1171, #1173, #1175, #1176, #1179, #1182

### DIFF
--- a/convex/functions.ts
+++ b/convex/functions.ts
@@ -170,18 +170,13 @@ export async function syncPackageSearchDigestsForOwnerPublisherId(
   ownerPublisherId: Id<"publishers"> | null | undefined,
 ) {
   if (!ownerPublisherId) return;
-  let cursor: string | null = null;
   try {
-    while (true) {
-      const page = await ctx.db
-        .query("packages")
-        .withIndex("by_owner_publisher", (q) => q.eq("ownerPublisherId", ownerPublisherId))
-        .paginate({ cursor, numItems: 100 });
-      for (const pkg of page.page) {
-        await syncPackageSearchDigest(ctx, pkg);
-      }
-      if (page.isDone) break;
-      cursor = page.continueCursor;
+    const packages = await ctx.db
+      .query("packages")
+      .withIndex("by_owner_publisher", (q) => q.eq("ownerPublisherId", ownerPublisherId))
+      .collect();
+    for (const pkg of packages) {
+      await syncPackageSearchDigest(ctx, pkg);
     }
   } catch (error) {
     if (isMissingTableError(error, "packages")) return;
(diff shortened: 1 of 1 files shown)


ImLukeF:

PR #1189, opened 2026-03-23
fix: replace paginate with collect in publisher trigger sync functions

jiweiyuan:
## Problem

The `publishers` trigger in `convex/functions.ts` calls both `syncPackageSearchDigestsForOwnerPublisherId` and `syncSkillSearchDigestsForOwnerPublisherId` sequentially. Each used a `.paginate()` while-loop, causing Convex to reject the mutation with:

```
Uncaught Error: This query or mutation function ran multiple paginated queries.
```

This breaks:
- `clawhub publish` — fails when `ensurePersonalPublisherForUser` inserts a new publisher record
- `clawhub` sync — same error
- Web publishing — results in 'Personal publisher not found' since the publisher record is never created

## Fix

Replace `.paginate()` loops with `.collect()` in all three sync-by-owner functions:
- `syncPackageSearchDigestsForOwnerUserId`
- `syncPackageSearchDigestsForOwnerPublisherId`
- `syncSkillSearchDigestsForOwnerPublisherId`

`.collect()` does not count against Convex's single-pagination-per-mutation limit, so the `publishers` trigger can call multiple sync functions without error.

## Testing

The `getPreferredFallbackPackageRelease` function retains `.paginate()` since it's only called from the `packageReleases` trigger (no multi-pagination conflict).

### DIFF
--- a/convex/functions.ts
+++ b/convex/functions.ts
@@ -146,18 +146,13 @@ export async function syncPackageSearchDigestsForOwnerUserId(
   ownerUserId: Id<"users"> | null | undefined,
 ) {
   if (!ownerUserId) return;
-  let cursor: string | null = null;
   try {
-    while (true) {
-      const page = await ctx.db
-        .query("packages")
-        .withIndex("by_owner", (q) => q.eq("ownerUserId", ownerUserId))
-        .paginate({ cursor, numItems: 100 });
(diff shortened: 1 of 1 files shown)


ImLukeF:

PR #1196, opened 2026-03-23
fix(convex): schedule package+skill digest workers to avoid dual paginate

kevinksaji:
## Summary
Fix publisher-triggered digest syncing by moving package and skill owner-wide sync into scheduled paginated workers.

## What this changes
- Replaces inline `publishers` trigger owner-wide sync with two scheduled internal workers:
  - `syncPackageSearchDigestsForOwnerPublisherIdInternal`
  - `syncSkillSearchDigestsForOwnerPublisherIdInternal`
- Each worker processes one page and self-schedules with `continueCursor` until complete.
- `publishers` trigger now enqueues both workers instead of running owner-wide pagination inline.

## Why
- Prevents the dual-paginate-in-one-execution Convex error.
- Preserves scalability by keeping cursor pagination (no `.collect()` bulk reads).
- Avoids TDZ issues by declaring workers with `rawInternalMutation`.

## Additional clarity improvements
- Uses explicit single-page helper naming (`...PageForOwnerPublisherId`).
- Uses consistent `null` sentinel on non-page paths.
- Adds inline comments explaining scheduler continuation and Convex pagination constraints.

## File changed
- `convex/functions.ts`

### DIFF
--- a/convex/functions.ts
+++ b/convex/functions.ts
@@ -1,6 +1,8 @@
 import { customCtx, customMutation } from "convex-helpers/server/customFunctions";
 import { Triggers } from "convex-helpers/server/triggers";
+import { v } from "convex/values";
 import semver from "semver";
+import { internal } from "./_generated/api";
 import type { DataModel, Doc, Id } from "./_generated/dataModel";
 import {
   mutation as rawMutation,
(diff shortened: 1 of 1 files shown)


ImLukeF:

Issue #1185, opened 2026-03-23
Published skill not indexed in search (ai-clipping)


andygogogo2025:
# Summary
This skill (https://clawhub.ai/wayinvideo/ai-clipping) can be accessed directly via its URL or through clawhub inspect, but it does not show up in any search results—neither on the CLI nor on the website.

# Environment
clawhub CLI: v0.7.0
skill: https://clawhub.ai/wayinvideo/ai-clipping

# Steps to Reproduce

✔ Verify skill exists:
> clawhub inspect ai-clipping

The output:
ai-clipping  AI Clipping
Summary: AI-powered highlight extraction that identifies the most engaging moments and generates viral-ready clips. Ideal for social media content creation, highlight...
Owner: wayinvideo
Created: 2026-03-19T11:47:26.987Z
Updated: 2026-03-20T19:55:53.252Z
Latest: 1.0.4
Tags: latest=1.0.4

✔ Visit direct URL:
https://clawhub.ai/wayinvideo/ai-clipping → ✅ Page loads, content renders correctly

× Search for it:

clawhub search "clipping"        # → empty results
clawhub search "ai-clipping"       # → empty results
clawhub search "ai clipping" # → 10 results, skill NOT included

Website search at https://clawhub.ai/skills?q=clipping → "No skills match that filter"

steipete:
Fixed on `main` in `456f4db` (`fix(search): widen lexical fallback coverage`).

I verified the reported production state before the fix: the `ai-clipping` skill detail API resolves, while production search for `clipping` currently returns no results. The landed fix expands the lexical fallback window and also scans recently created skill digests, so newly published skills can be found before embeddings rank well.

Verification on the landed `main` commit:
- `bun run test convex/search.test.ts`
- `bunx tsc --noEmit --pretty false`
- `bun run lint`
- `bun run test`
- `bun run coverage`
- `bunx tsc -p packages/schema/tsconfig.json --noEmit --pretty false`
- `bunx tsc -p packages/clawhub/tsconfig.json --noEmit --pretty false`
- `bun run build`
- GitHub CI and secret scan are green.

Production deploys are manual-only in this repo, so the fix will be visible on clawhub.ai after the next production deploy from `main`.

PR #1200, opened 2026-03-23
fix(search): improve lexical fallback coverage for recently published skills (#1185)

thirumaleshp:
## Summary
- Raise `FALLBACK_SCAN_LIMIT` from 500 to 2000 to cover more skills in the lexical fallback
- Add a secondary scan by `createdAt` alongside the existing `updatedAt` scan, so recently published skills are found even if not recently updated
- Deduplicate results across both scan windows

## Problem
Published skills can be invisible in search when their embedding has low vector similarity AND the skill falls outside the 500 most recently updated digests in the lexical fallback scan. This was reported in #1185 where a published skill was accessible via direct URL but didn't appear in any search queries.

## How it works
The lexical fallback previously only scanned `skillSearchDigest` by `by_active_updated` (descending). On an active registry, a newly published skill that hasn't been modified since publication could fall outside the top 500, making it invisible to the fallback path.

This PR:
1. Raises the scan limit from 500 → 2000 (~1.6MB at ~800 bytes/digest, within Convex's 8MB limit)
2. Adds a parallel scan by `by_active_created` to catch recently created skills missed by the `updatedAt` window
3. Deduplicates across both scans using the existing `seenSkillIds` set

## Test plan
- [x] Added test: skill found only via `createdAt` scan is included in results
- [x] Added test: skills appearing in both scans are deduplicated
- [x] All 18 existing + new search tests pass

### DIFF
--- a/convex/search.test.ts
+++ b/convex/search.test.ts
@@ -576,6 +576,51 @@ describe("search helpers", () => {
     expect(result).toHaveLength(0);
   });
 
+  it("finds recently created skill missed by updatedAt scan via createdAt scan (#1185)", async () => {
+    // Skill was recently created but not recently updated — only appears in by_active_created.
+    const newSkill = makeSkillDoc({
+      id: "skills:new",
+      slug: "ai-clipping",
+      displayName: "AI Clipping",
+    });
+    const ctx = makeLexicalCtx({
+      exactSlugSkill: null,
+      recentSkills: [], // updatedAt scan returns nothing
+      recentByCreated: [newSkill], // createdAt scan finds it
+    });
+
+    const result = await lexicalFallbackSkillsHandler(ctx, {
+      query: "clipping",
+      queryTokens: ["clipping"],
+      limit: 10,
+    });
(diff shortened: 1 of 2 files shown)


steipete:
Thanks for the fix. I landed the same behavior on `main` in `456f4db` (`fix(search): widen lexical fallback coverage`), adapted on top of the current search code.

What landed:
- `FALLBACK_SCAN_LIMIT` increased from 500 to 2000.
- Lexical fallback now scans both `skillSearchDigest.by_active_updated` and `skillSearchDigest.by_active_created` with indexed bounded queries.
- Results are deduplicated across both fallback windows.
- Added regression coverage for the #1185 `ai-clipping` shape and cross-window dedupe.
- Added changelog credit for @thirumaleshp.

Verification on the landed `main` commit:
- `bun run lint`
- `bun run test`
- `bun run coverage`
- `bun run build`
- GitHub CI and secret scan are green.

Closing this PR as superseded by the landed main commit.

PR #1202, opened 2026-03-23
fix: replace .paginate() with .collect() in publisher trigger sync (#1201)

### LINKED ISSUE #1201: CLI publish fails: Convex pagination error in ensurePersonalPublisherForUser
## Bug Report

**CLI version:** ClawHub CLI v0.7.0 (8e5ef402)
**Account:** @liveneon (first-time publisher)
**OS:** macOS, Apple Silicon

## Steps to reproduce

```bash
clawhub login --token <token> --no-browser
# ✔ OK. Logged in as @liveneon.

clawhub --workdir . publish skills/ollama-herd \
  --slug ollama-herd \
  --name "Ollama Herd" \
  --version 1.1.0 \
  --tags "ollama,fleet-management"
```

## Error output

```
- Preparing ollama-herd@1.1.0
✖ Uncaught Error: This query or mutation function ran multiple paginated queries.
    at async syncSkillSearchDigestsForOwnerPublisherId (../../convex/functions.ts:223:17)
    at async <anonymous> (../../convex/functions.ts:339:1)
    ...
    at async ensurePersonalPublisherForUser (../../convex/lib/publishers.ts:186:14)
```

## Analysis

The error occurs in `ensurePersonalPublisherForUser` — it appears to trigger when a user publishes for the first time and the system tries to create their publisher profile. The `syncSkillSearchDigestsForOwnerPublisherId` function runs multiple paginated queries within a single Convex mutation, which Convex doesn't support.

Existing accounts that have already published (e.g., `@driftsbot`) don't hit this because their publisher profile already exists.

## Workaround

Publishing under an account that has previously published works fine. New accounts are blocked.

thirumaleshp:
## Summary

- Fixes #1201 — CLI publish fails with Convex pagination error in `syncSkillSearchDigestsForOwnerPublisherId`
- The `publishers` trigger calls both `syncPackageSearchDigestsForOwnerPublisherId` and `syncSkillSearchDigestsForOwnerPublisherId` sequentially, each using a `while(true) { .paginate() }` loop — violating Convex's single-pagination-per-function constraint
- Replace `.paginate()` loops with `.collect()` in all three owner-scoped sync functions (`syncPackageSearchDigestsForOwnerUserId`, `syncPackageSearchDigestsForOwnerPublisherId`, `syncSkillSearchDigestsForOwnerPublisherId`). These queries are index-scoped to a single owner so the result set is always small.

## Test plan

- [x] Updated existing test to use `.collect()` mock instead of `.paginate()` mock
- [x] Added new test verifying both package and skill sync can run sequentially without pagination conflict (simulates the publishers trigger)
- [x] All 6 `convex/functions.test.ts` tests pass

### DIFF
--- a/convex/functions.test.ts
+++ b/convex/functions.test.ts
@@ -445,7 +445,11 @@ describe("package digest sync", () => {
       updatedAt: 2,
       verification: undefined,
     };
-    const take = vi.fn().mockResolvedValueOnce([pkg]);
+    const paginate = vi.fn().mockResolvedValueOnce({
+      page: [pkg],
+      isDone: true,
+      continueCursor: "",
+    });
     const ctx = {
       db: {
         get: vi.fn(async (id: string) => {
@@ -456,7 +460,7 @@ describe("package digest sync", () => {
           if (table === "packages") {
             return {
               withIndex: vi.fn(() => ({
-                take,
+                paginate,
               })),
             };
           }
@@ -489,7 +493,7 @@ describe("package digest sync", () => {
       "users:owner" as never,
     );
 
-    expect(take).toHaveBeenCalled();
+    expect(paginate).toHaveBeenCalled();
     expect(ctx.db.insert).toHaveBeenCalledWith(
       "packageSearchDigest",
       expect.objectContaining({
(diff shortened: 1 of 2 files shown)


ImLukeF:

PR #1230, opened 2026-03-24
fix: re-enable theme switching

whoisheartbreak:
Closes #1228

## Summary
- re-enable switching between system, light, and dark themes from the top-right header control
- keep the original Radix `ToggleGroup` pill UI and restore both click and keyboard-driven theme changes
- default first-time ClawHub visits to `system` by ignoring the old `clawdhub-theme` storage key
- add regressions covering desktop theme switching, keyboard navigation, and the first-visit `system` default

## Why this approach
This keeps the original desktop theme toggle structure and active-state styling, restores the broken interaction path, preserves keyboard accessibility on the controlled `ToggleGroup`, and removes the legacy storage fallback that could incorrectly preselect `dark` on a first ClawHub visit.

## Testing
- `bun run lint`
- `bun run test`
- `bun run build`

## AI assistance
- AI-assisted change
- Human verified locally by running the commands above

## Screenshot
- Local visual verification completed for the header theme toggle update
Bildschirmfoto 2026-03-24 um 12 42 01 
Bildschirmfoto 2026-03-24 um 13 23 14

PR #1234, opened 2026-03-24
fix: simplify desktop theme toggle and drop legacy storage key

### LINKED ISSUE #1228: Theme toggle: re-enable switching between system, light, and dark modes
## Summary
The top-right theme toggle stopped switching between system, light, and dark modes.

## Fix
Re-enable the theme switcher so selecting system, light, or dark updates the active theme correctly again.

## Notes
The user-visible goal is to restore working theme switching across all three modes in the ClawHub header.

obrutjack:
Closes #1228

## What was wrong

The desktop theme toggle (system / light / dark) wasn't responding to clicks. The root cause is that Radix's `ToggleGroup` fires `onValueChange` with an empty string when you click the already-selected item — and the old `if (!value) return` guard was silently eating the event for all clicks.

### `Header.tsx`
I moved from `onValueChange` on the `ToggleGroup` to `onClick` on each individual `ToggleGroupItem`. This sidesteps the empty-string problem entirely. No extra refs or dedup logic needed — `startThemeTransition` already bails out when `currentTheme === nextTheme`, so clicking the active pill is naturally a no-op.

### `theme.ts`
Dropped the legacy `clawdhub-theme` localStorage fallback. It was causing some first-time visitors to land on dark mode from stale data instead of getting the `system` default.

### Tests
- `header.test.tsx`: Added a test that clicks a theme button and checks the full chain fires (`startThemeTransition` → `applyTheme` → `setMode`). Kept the mock minimal — just forwards `onClick`, nothing fancy.
- `theme.test.tsx`: Split the legacy key check into its own test case confirming it's now ignored.

## How this differs from #1230

PR #1230 uses a `pointerThemeRef` to coordinate between `onClick` and `onValueChange` so they don't double-fire. That works, but it depends on browser event ordering between the two handlers, which could get flaky across Radix versions. I just removed `onValueChange` entirely so there's only one code path.

The test mock is also much smaller — #1230 rebuilds the full ToggleGroup with `role="radio"`, `data-state`, `aria-checked`, and keyboard nav simulation (~50 lines). I only need `onClick` forwarding (~10 lines) since that's what I'm actually testing.

## Verification

- `bun run lint` — 0 warnings, 0 errors
- `bun run test` — 140 files, 993 tests passed
- `bun run build` — clean

### DIFF
--- a/src/__tests__/header.test.tsx
+++ b/src/__tests__/header.test.tsx
@@ -1,9 +1,21 @@
 /* @vitest-environment jsdom */
 
-import { render, screen } from "@testing-library/react";
+import { fireEvent, render, screen } from "@testing-library/react";
 import type { ReactNode } from "react";
-import { describe, expect, it, vi } from "vitest";
+import { beforeEach, describe, expect, it, vi } from "vitest";
 import Header from "../components/Header";
+
+// I hoist these mocks so they're available inside vi.mock() factories.
+const { applyThemeMock, setModeMock, startThemeTransitionMock } = vi.hoisted(() => ({
+  applyThemeMock: vi.fn(),
+  setModeMock: vi.fn(),
+  // This fake calls setTheme right away, so I can assert that
+  // applyTheme and setMode got the right value downstream.
+  startThemeTransitionMock: vi.fn(
+    ({ setTheme, nextTheme }: { setTheme: (v: string) => void; nextTheme: string }) =>
+      setTheme(nextTheme),
+  ),
+}));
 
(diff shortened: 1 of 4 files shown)


ImLukeF:
Thanks for the fix here. I'm going to close this one because we already have an overlapping theme-toggle PR open that covers the same bug with the more complete path, including the keyboard interaction side as well: https://github.com/openclaw/clawhub/pull/1230

That keeps this issue tracked in one place instead of splitting review across two very similar fixes.

PR #1251, opened 2026-03-25
fix: avoid double JSON encoding in transfer, rename, and merge requests

bennyzhe:
## Summary

  This fixes a CLI bug where some `POST` requests were JSON-encoded twice before being sent.

  I hit this while trying to transfer an existing skill:

  ```bash
  clawhub transfer request <skill name> <github handle>
 ```

  The CLI showed the confirmation prompt correctly, but the server returned:

  toUserHandle required

  ## Root Cause

  Some command handlers were doing:

  ```ts
  body: JSON.stringify(...)

  before passing the payload to apiRequest().

  But apiRequest() already serializes args.body for JSON POST requests.

  That meant the server received a JSON string literal instead of a JSON object.

  ## Changes

  - Update transfer requests to pass plain objects to apiRequest()
  - Update rename requests to pass plain objects to apiRequest()
  - Update merge requests to pass plain objects to apiRequest()
  - Make http.ts tolerate already-serialized string bodies to avoid double-encoding regressions
  - Update command tests to assert object payloads instead of pre-stringified bodies
  - Add an http.ts test covering pre-serialized JSON passthrough

  ## Why this approach

  Fixing only transfer would leave the same bug pattern in other command handlers.

  This change fixes the affected call sites and also hardens the shared HTTP layer so accidental pre-
  serialized bodies are not double-encoded in the future.

  ## Validation

  Ran:

  bunx vitest run \
    packages/clawdhub/src/cli/commands/transfer.test.ts \
    packages/clawdhub/src/cli/commands/ownership.test.ts \
    packages/clawdhub/src/http.test.ts

  Result:

  - 3 test files passed
  - 31 tests passed

  ## User impact

  This fixes broken CLI flows for requests that expect JSON object bodies, including:

  - skill transfer requests
  - skill rename requests
  - skill merge requests

### DIFF
--- a/packages/clawdhub/src/cli/commands/ownership.test.ts
+++ b/packages/clawdhub/src/cli/commands/ownership.test.ts
@@ -67,8 +67,8 @@ describe("ownership commands", () => {
       }),
       expect.anything(),
     );
-    const requestArgs = mockApiRequest.mock.calls[0]?.[1] as { body?: string };
-    expect(requestArgs.body).toContain('"newSlug":"demo-new"');
+    const requestArgs = mockApiRequest.mock.calls[0]?.[1] as { body?: unknown };
+    expect(requestArgs.body).toEqual({ newSlug: "demo-new" });
   });
 
   it("merge calls merge endpoint", async () => {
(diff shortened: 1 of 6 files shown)


ImLukeF:

PR #1291, opened 2026-03-26
fix: fallback to lexical search when embedding generation fails

goulonghui:
## Problem

When `generateEmbedding()` throws (e.g. embedding API unavailable), both `searchSkills` and 
`searchSouls` immediately return an empty array. However, the lexical fallback path 
(`lexicalFallbackSkills`) does not depend on embeddings — it uses token matching against 
skill slugs, names, and summaries.

## Solution

- Changed `vector` from `number[]` to `number[] | null`
- The vector search `while` loop is now guarded by `vector &&`, so it is skipped when 
  embedding is unavailable
- For `searchSkills`, this naturally falls through to `lexicalFallbackSkills`
- For `searchSouls`, the vector loop is similarly skipped

## Changes

- `convex/search.ts`: Modified `searchSkills` and `searchSouls` to gracefully handle 
  embedding failures
- `convex/search.test.ts`: Added test case verifying lexical fallback is invoked when 
  embedding generation fails

All 17 tests pass.

### DIFF
--- a/convex/search.test.ts
+++ b/convex/search.test.ts
@@ -65,6 +65,36 @@ describe("search helpers", () => {
     );
   });
 
+  it("falls back to lexical search when embedding generation fails", async () => {
+    generateEmbeddingMock.mockRejectedValueOnce(new Error("API unavailable"));
+    const fallback = [
+      {
+        skill: makePublicSkill({ id: "skills:orf", slug: "orf", displayName: "ORF" }),
+        version: null,
+        ownerHandle: "steipete",
+        owner: null,
+      },
+    ];
+    const runQuery = vi.fn().mockResolvedValueOnce(fallback); // lexicalFallbackSkills
+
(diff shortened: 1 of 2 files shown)


steipete:
Landed the useful part of this on `main` in e2cb7df (`fix(search): fall back when embeddings fail`).

What landed:
- `searchSkills` now falls through to `lexicalFallbackSkills` when embedding generation fails instead of returning empty results.
- Added regression coverage that proves vector search is skipped and lexical fallback is invoked on embedding failure.
- Left `searchSouls` behavior unchanged because there is still no soul lexical fallback path; it already catches the embedding failure and returns empty results accurately.
- Changelog credits this PR and @goulonghui.

Verified locally:
- `bun run test`
- `bun run coverage`
- `bun run lint`
- `bun run build`

Closing this PR as superseded by the landed commit. Thanks for the fix.

PR #1330, opened 2026-03-28
fix: remove duplicate search navigation item

kevinw-openai:
## Summary

Removes the duplicate `Search` navigation item from the header.

The current `Search` nav entry routes users back into the same skills browse flow, so it appears redundant with `Skills`.

## Testing

Before:
image

After:
image

- `bun run test -- src/__tests__/header.test.tsx src/__tests__/search-route.test.ts`
- `bunx oxlint --type-aware --tsconfig ./tsconfig.oxlint.json src/components/Header.tsx src/__tests__/header.test.tsx`
- `bun run build`
- `bun run check:peers`
- `bunx tsc -p packages/clawdhub/tsconfig.json --noEmit`

## Notes

- `bun run lint` fails locally due to an existing repo issue referencing missing `convex/lib/moderationEngine.test.ts`
- `bun run test` has existing Windows-specific failures in `packages/clawdhub`

### DIFF
--- a/src/__tests__/header.test.tsx
+++ b/src/__tests__/header.test.tsx
@@ -2,8 +2,10 @@
 
 import { render, screen } from "@testing-library/react";
 import type { ReactNode } from "react";
-import { describe, expect, it, vi } from "vitest";
+import { beforeEach, describe, expect, it, vi } from "vitest";
 import Header from "../components/Header";
+
+let mockSiteMode: "souls" | "skills" = "souls";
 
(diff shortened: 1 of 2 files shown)


ImLukeF:
Thanks for the cleanup here. I’m going to close this for now because we should revisit this navigation decision alongside the broader website overhaul and the larger staging/design work that’s already in flight.

I appreciate the thoughtful pass on the header. Once the overhaul settles, we can re-evaluate this in the new navigation structure.

PR #1331, opened 2026-03-28
fix: normalize github sign-in button spacing and fix cursor 

kevinw-openai:
## Why

The header GitHub sign-in CTA was rendered as two separate text spans, so the spacing between `Sign in` and `with GitHub` came from the shared button flex gap instead of normal text spacing.

That created two small but visible UI issues:
- the space between `in` and `with` looked wider than the rest of the label
- hovering the button showed the default arrow cursor instead of a pointer

## What Changed

- added a dedicated `sign-in-button` class to the header auth CTA
- tuned the spacing between the two text spans so the label reads like normal text again
- added `cursor: pointer` to the shared `.btn` base style so button hover behavior matches expectations
- added a focused header test covering the sign-in button structure

## Before
image

- `Sign in` and `with GitHub` were separated by the shared flex gap, which looked too wide for plain text
- the sign-in button did not show a pointer cursor on hover

## After
image

- `Sign in with GitHub` reads with natural-looking spacing
- the sign-in button shows a pointer cursor on hover
- the header test locks in the dedicated sign-in button structure

## Verification

Passed:
- `bunx vitest run src/__tests__/header.test.tsx`
- `bunx oxfmt --check src/components/Header.tsx src/__tests__/header.test.tsx src/styles.css`
- `bun run build`
- `bun run lint`

### DIFF
--- a/src/__tests__/header.test.tsx
+++ b/src/__tests__/header.test.tsx
@@ -33,8 +33,13 @@ vi.mock("../lib/theme", () => ({
 }));
 
 vi.mock("../lib/theme-transition", () => ({
-  startThemeTransition: ({ setTheme, nextTheme }: { setTheme: (value: string) => void; nextTheme: string }) =>
-    setTheme(nextTheme),
+  startThemeTransition: ({
+    setTheme,
+    nextTheme,
+  }: {
+    setTheme: (value: string) => void;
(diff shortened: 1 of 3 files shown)


ImLukeF:

PR #1332, opened 2026-03-28
fix: align skills page width with site layout

kevinw-openai:
## Summary

Fix the `Skills` page so it uses the same content width as other top-level pages like `About` and `Plugins`.

## Why

The `Skills` page was visibly narrower than other pages because it wrapped the toolbar and results in a page-specific `.skills-container` with `max-width: 900px`.

Other top-level pages already rely on the shared `.section` layout width, so the extra wrapper made `/skills` look inconsistent.

## What Changed

- Removed the extra `.skills-container` wrapper from the skills index page
- Removed the page-specific `max-width: 900px` CSS constraint
- Added a regression test to ensure the skills page is not wrapped in the narrow container again

## Before
image

- `/skills` content was constrained to a narrower width than `/about`
- The toolbar and results table looked more compressed than the rest of the site

## After
image

- `/skills` now uses the shared page width from the main section layout
- The toolbar and results align with the same width system used elsewhere

## Verification

- `bun run test -- src/__tests__/skills-index.test.tsx`
- `bun run build`
- Ran the app locally and verified the updated `/skills` layout in the browser
- Verified locally with Playwright that `/skills` and `/about` now share the same main content width

## AI Assistance

This PR was prepared with Codex and manually verified locally.

### DIFF
--- a/src/__tests__/skills-index.test.tsx
+++ b/src/__tests__/skills-index.test.tsx
@@ -63,6 +63,13 @@ describe("SkillsIndex", () => {
     );
   });
 
+  it("does not wrap the skills page content in a narrow width container", async () => {
+    render(<SkillsIndex />);
+    await act(async () => {});
+
+    expect(document.querySelector(".skills-container")).toBeNull();
+  });
+
   it("renders an empty state when no skills are returned", async () => {
     render(<SkillsIndex />);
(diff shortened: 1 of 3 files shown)


ImLukeF:

PR #1337, opened 2026-03-28
feat: add publish skill CTA to skills toolbar

kevinw-openai:
## Summary

Add a `Publish Skill` CTA to the `/skills` browse toolbar so the skills page matches the plugin catalog and homepage entry points.

## Why

The skills index page currently lets users browse, search, filter, and sort skills, but it does not expose a direct publishing action.

That creates an inconsistency in the product:
- the homepage already has a `Publish Skill` button
- the plugins page already has a `Publish Plugin` button in the toolbar
- the skills page is missing the equivalent CTA

Adding the button makes publishing easier to discover for users who land directly on the skills catalog.

## What Changed

- Added a `Publish Skill` button to the right side of the skills toolbar
- Reused the existing `/publish-skill` route and existing `btn btn-primary` styling
- Kept the layout consistent with the current toolbar structure and the plugins page CTA placement
- Added a regression test to verify the publish action is rendered on the skills page
- Updated the related test `Link` mocks to preserve `className`
- Tightened one existing sorting test so it asserts against skill result rows instead of assuming the first page link is a result item

## Before
image
image

- `/skills` had search/filter/sort controls only
- users browsing the skills catalog had no direct publish CTA on that page

## After

image

- `/skills` shows a `Publish Skill` button in the toolbar
- clicking the button navigates to `/publish-skill`

## Testing

- `bun run test -- src/__tests__/skills-index.test.tsx src/__tests__/skills-index-load-more.test.tsx`
- `bun x oxlint src/routes/skills/-SkillsToolbar.tsx src/__tests__/skills-index.test.tsx src/__tests__/skills-index-load-more.test.tsx`
- `bun run build`

## Verification

- Ran the app locally
- Verified `http://localhost:3000/skills?sort=downloads` shows the new `Publish Skill` CTA in the toolbar
- Verified clicking the button navigates to `/publish-skill`

## AI Usage

This PR was made by Codex, with local verification applied before submission.

### DIFF
--- a/src/__tests__/skills-index-load-more.test.tsx
+++ b/src/__tests__/skills-index-load-more.test.tsx
@@ -19,7 +19,11 @@ vi.mock("@tanstack/react-router", () => ({
     useSearch: () => searchMock,
   }),
   redirect: (options: unknown) => ({ redirect: options }),
-  Link: (props: { children: ReactNode }) => <a href="/">{props.children}</a>,
+  Link: (props: { children: ReactNode; className?: string }) => (
+    <a href="/" className={props.className}>
+      {props.children}
+    </a>
+  ),
 }));
 
 vi.mock("convex/react", () => ({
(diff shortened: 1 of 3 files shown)


ImLukeF:

PR #1339, opened 2026-03-28
fix(ui): align publish-skill header with publish-plugin and remove custom background

kevinw-openai:
## Why

The `/publish-skill` page had a custom square background that felt visually off, especially compared with the cleaner and more restrained header on `/publish-plugin`.

## What changed

- removed the custom publish-skill hero/background treatment
- switched the page header to the same simple title/subtitle structure used on the plugin publishing page
- kept the existing two-column form layout by scoping the grid override to a dedicated `publish-skill-grid` class

## Before
image

- the page showed a long tinted rectangle-like background behind the title area
- the top section felt heavier and visually inconsistent with the plugin publishing flow

## After
image

- the odd colored header block is gone
- the page now uses a plain title + subtitle header, matching the plugin page more closely
- the form and upload interaction remain unchanged

## Verification

- `bunx vitest run src/__tests__/upload.route.test.tsx`
- `bun run build`
- `bun run lint`

### DIFF
--- a/src/routes/publish-skill.tsx
+++ b/src/routes/publish-skill.tsx
@@ -432,17 +432,17 @@ export function Upload() {
   }
 
   return (
-    <main className="section upload-page">
-      <header className="upload-page-header">
-        <div>
-          <h1 className="upload-page-title">Publish a {contentLabel}</h1>
-          <p className="upload-page-subtitle">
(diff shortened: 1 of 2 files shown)


ImLukeF:

PR #1355, opened 2026-03-28
fix: add a proper 404 page

kevinw-openai:
## Why

Visiting a non-existent route currently falls back to TanStack Router's default `Not Found` paragraph.
That leaves the page looking unstyled and broken, with no clear recovery action for the user.

## What changed

- registered an app-level 404 handler via `defaultNotFoundComponent`
- added a dedicated `NotFoundPage` component with real layout and styling
- kept the final UI intentionally minimal:
  - `404 • Page not found`
  - main explanatory copy
  - a single `Return home` CTA
- added tests to cover both the router-level registration and the rendered 404 UI

## Before
chrome-capture-2026-03-28 (1)

- unknown routes rendered as a bare `Not Found` string
- no meaningful content container
- no clear path back to the homepage

## After
chrome-capture-2026-03-28

- unknown routes render inside the normal app shell
- the page has a styled 404 container consistent with the site
- users get a single, obvious `Return home` action

## Verification

- `bunx vitest run src/components/NotFoundPage.test.tsx src/router.test.tsx`
- `bunx tsc --noEmit`
- `bunx tsc -p packages/clawdhub/tsconfig.json --noEmit`
- `bun --bun vite build`
- verified locally in browser at `/asdfd`
- verified with Playwright against the local page

This PR is assisted by codex.

### DIFF
--- a/src/components/NotFoundPage.test.tsx
+++ b/src/components/NotFoundPage.test.tsx
@@ -1,0 +1,26 @@
+/* @vitest-environment jsdom */
+
+import { render, screen } from "@testing-library/react";
+import type { ReactNode } from "react";
+import { describe, expect, it, vi } from "vitest";
+import { NotFoundPage } from "./NotFoundPage";
+
+vi.mock("@tanstack/react-router", () => ({
+  Link: ({ children, to, className }: { children: ReactNode; to: string; className?: string }) => (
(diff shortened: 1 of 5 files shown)


ImLukeF:

PR #1365, opened 2026-03-29
fix: remove double JSON.stringify in transfer request

brunobuddy:
## Summary

`clawhub transfer request <slug> <handle>` always fails with `toUserHandle required` because the request body is double-stringified.

In `transfer.ts`, `cmdTransferRequest` wraps the body in `JSON.stringify()` before passing it to `apiRequest`. But `apiRequest` in `http.ts` also calls `JSON.stringify(args.body)` for POST requests (line 74). The server receives a JSON string containing a string instead of an object, so it can't extract `toUserHandle`.

## Fix

Remove the redundant `JSON.stringify()` in `transfer.ts` so the body is passed as a plain object, matching the contract of `apiRequest` (and matching how other POST callers like `apiRequestForm` work).

## Reproduction

```bash
clawhub transfer request my-skill @someuser --yes
# Error: toUserHandle required
```

## Test plan

- [x] Updated existing test to assert on the raw object instead of a pre-stringified string
- [ ] CI passes

### DIFF
--- a/packages/clawdhub/src/cli/commands/transfer.test.ts
+++ b/packages/clawdhub/src/cli/commands/transfer.test.ts
@@ -82,8 +82,8 @@ describe("transfer commands", () => {
       }),
       expect.anything(),
     );
-    const requestArgs = mockApiRequest.mock.calls[0]?.[1] as { body?: string };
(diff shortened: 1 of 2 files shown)


ImLukeF:

PR #1372, opened 2026-03-29
feat: add ability to delete version tags from skill detail page

tjefferson:
## Problem

Version tags on skills accumulate over time with no way to remove them. Each publish or manual tag update only appends/merges tags — old tags are never cleaned up. This leaves skill pages cluttered with stale tags pointing to old versions.

## Solution

Add a `deleteTags` mutation and a delete button (×) on each tag in the skill detail page.

### Changes

- **`convex/skills.ts`**: New `deleteTags` mutation — accepts a list of tag names to remove. Protects the `latest` tag from deletion. Same permission model as `updateTags` (owner or moderator).
- **`src/components/SkillDetailPage.tsx`**: Wire up `deleteTags` mutation and pass `onTagDelete` prop to `SkillHeader`.
- **`src/components/SkillHeader.tsx`**: Show a × delete button on each tag (only visible to skill owner/moderator, hidden for `latest` tag).
- **`src/styles.css`**: Minimal `.tag-delete` button styles.

### Screenshots

Delete button appears inline on each tag (except `latest`), visible only to the skill owner:

```
[A-shares v1.1.2 ×] [Hong-Kong v1.1.2 ×] [latest v1.1.3]
```

### Notes

- The `latest` tag is protected and cannot be deleted.
- This does not change the existing `updateTags` or `insertVersion` merge behavior — that could be addressed separately if desired.

### DIFF
--- a/convex/skills.deleteTags.test.ts
+++ b/convex/skills.deleteTags.test.ts
@@ -1,0 +1,169 @@
+import { describe, expect, it, vi } from "vitest";
+
+vi.mock("@convex-dev/auth/server", () => ({
+  getAuthUserId: vi.fn(),
+  authTables: {},
+}));
+
+import { deleteTags } from "./skills";
+
+type WrappedHandler<TArgs, TResult = unknown> = {
+  _handler: (ctx: unknown, args: TArgs) => Promise<TResult>;
+};
+
+const deleteTagsHandler = (
+  deleteTags as unknown as WrappedHandler<{
+    skillId: string;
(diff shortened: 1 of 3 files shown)


ImLukeF:
@tjefferson Thanks for putting this feature together. The delete-tags work itself was solid and it ended up landing via #1380 after a small amount of follow-up test-harness and typecheck cleanup to get it through CI cleanly. Appreciate you pushing on the stale-tag cleanup problem.

PR #1429, opened 2026-03-31
fix: raise vector candidate pool minimum to prevent search misses

### LINKED ISSUE #1375: bug: search misses highest-scoring skill when limit is small (vector candidate set too small)
## Bug Description

The highest-scoring skill for a query can be completely absent from search results when `limit` is small (e.g. the CLI default of 10), but appears as **#1** when `limit` is larger (e.g. the web UI default of 25).

## Reproduction

```bash
# CLI default (limit=10) — stock-price-query is MISSING
clawhub search 'stock price'
# Results: tencent-finance-stock-price (3.467), jarvis-stock-price (3.378), ...

# With limit=25 — stock-price-query is #1
clawhub search 'stock price' --limit 25
# Results: stock-price-query (3.543), tencent-finance-stock-price (3.467), ...
```

`stock-price-query` scores **3.543** (highest), yet doesn't appear in the top 10. It outscores every result in the `limit=10` response.

## Root Cause

In `convex/search.ts`, the vector search candidate set size is derived from `limit`:

```typescript
const limit = args.limit ?? 10;
const candidateLimit = Math.min(Math.max(limit * 3, 50), 256);  // limit=10 → 50 candidates
```

The loop at line 201 breaks early once `exactMatches.length >= limit`:

```typescript
if (exactMatches.length >= limit || results.length < candidateLimit) {
    break;
}
```

When `limit=10`, only 50 vector candidates are fetched. If the target skill's **embedding similarity** doesn't rank in the top 50 (even though its **total score** would be highest after lexical + popularity boosts), it never enters the candidate pool. Meanwhile, 10 other skills with lower total scores but higher raw vector scores fill the quota and trigger the early break.

With `limit=25`, `candidateLimit` becomes 75, which is large enough to include the skill. After lexical boost (+2.5 for exact slug + name match) and popularity boost, it becomes #1.

## Impact

- **CLI users** see degraded search quality (CLI doesn't pass `limit`, so backend defaults to 10)
- **Web UI** happens to work because it passes `limit=25` (`pageSize` in `useSkillsBrowseModel.ts`)
- Any skill with high lexical relevance but moderate vector similarity can be silently dropped

## Suggested Fix

Decouple the vector candidate set size from the return `limit`. The candidate pool should always be large enough to capture skills that score high after lexical/popularity boosting:

```typescript
// Option A: Use a fixed minimum candidate size regardless of limit
const candidateLimit = Math.min(Math.max(limit * 3, 200), 256);

// Option B: Always start at maxCandidate
const candidateLimit = Math.min(Math.max(limit * 10, 200), 256);
```

Alternatively, the CLI could pass a reasonable default `limit` (e.g. 25) instead of relying on the backend default of 10.

tjefferson:
## Problem

The highest-scoring skill for a query can be completely absent from search results when `limit` is small (e.g. the CLI default of 10), but appears as **#1** when `limit` is larger (e.g. the web UI default of 25).

**Root cause:** In `convex/search.ts`, the initial vector candidate pool size was `max(limit * 3, 50)`. When `limit=10`, only 50 vector candidates were fetched. If a skill's embedding similarity didn't rank in the top 50 (even though its **total score** would be highest after lexical + popularity boosts), it never entered the candidate pool.

## Solution

Two changes that work together:

### 1. Backend: Raise candidate pool floor (`convex/search.ts`)

```diff
- let candidateLimit = Math.min(Math.max(limit * 3, 50), 256);
+ let candidateLimit = Math.min(Math.max(limit * 3, 200), 256);
```

The initial candidate pool now always starts at **200** (instead of 50), regardless of `limit`. This ensures skills with moderate vector similarity but high lexical/popularity boosts are not silently dropped.

### 2. CLI: Default to limit=25 (`packages/clawdhub/src/cli/commands/skills.ts`)

```diff
- if (typeof limit === "number" && Number.isFinite(limit)) {
-   url.searchParams.set("limit", String(limit));
- }
+ const effectiveLimit = typeof limit === "number" && Number.isFinite(limit) ? limit : 25;
+ url.searchParams.set("limit", String(effectiveLimit));
```

CLI `search` now defaults to `limit=25` (matching the web UI `pageSize`), instead of omitting the parameter and falling back to the server default of 10.

## Tests

- Updated the incremental hydration test in `convex/search.test.ts` to reflect the new candidate pool size
- Added two new CLI tests verifying the default limit (25) and explicit limit passthrough
- All 21 backend tests pass ✅
- All 27 CLI tests pass ✅

## Impact

- **CLI users** will now see the same search quality as web users
- **No behavioral change** for web UI (already uses `limit=25`)
- The candidate pool increase from 50→200 has minimal performance impact since Convex vector search is already optimized for up to 256 candidates

Fixes #1375

### DIFF
--- a/convex/search.test.ts
+++ b/convex/search.test.ts
@@ -732,14 +732,20 @@ describe("search helpers", () => {
   it("only hydrates new embedding IDs on subsequent iterations (incremental)", async () => {
     generateEmbeddingMock.mockResolvedValueOnce([0, 1, 2]);
 
-    // limit=10 → candidateLimit starts at 50, maxCandidate=200.
-    // First iteration must return exactly candidateLimit (50) to trigger expansion.
-    const firstBatch = Array.from({ length: 50 }, (_, i) => ({
+    // Use limit=100 so candidateLimit starts at 300 (capped to 256) and
+    // maxCandidate = min(1000,256) = 256.  Since 256 == 256 there is only
+    // one iteration, but we still need two iterations to test incremental
+    // hydration, so we use a lower limit.
+    //
+    // With the raised minimum (200), use limit=50 → candidateLimit =
+    // max(50*3,200) = 200, maxCandidate = min(500,256) = 256.
+    // First iteration returns exactly 200 to trigger expansion to 256.
+    const firstBatch = Array.from({ length: 200 }, (_, i) => ({
       _id: `skillEmbeddings:e${i}`,
       _score: 0.5 - i * 0.001,
     }));
-    // Second iteration returns 60 results (50 old + 10 new).
-    // 60 < next candidateLimit (100), so the loop breaks.
+    // Second iteration returns 210 results (200 old + 10 new).
+    // 210 < next candidateLimit (256 is the expanded limit), so the loop breaks.
     const secondBatch = [
       ...firstBatch,
       ...Array.from({ length: 10 }, (_, i) => ({
@@ -779,12 +785,12 @@ describe("search helpers", () => {
 
     await searchSkillsHandler(
       { vectorSearch: vectorSearchMock, runQuery },
-      { query: "test", limit: 10 },
+      { query: "test", limit: 50 },
     );
 
     // Should have been called twice, but second call should only have new IDs
     expect(hydrateCalls).toHaveLength(2);
-    expect(hydrateCalls[0]).toHaveLength(50);
+    expect(hydrateCalls[0]).toHaveLength(200);
     expect(hydrateCalls[1]).toHaveLength(10);
     // Verify no overlap between the two hydrate calls
     const firstSet = new Set(hydrateCalls[0]);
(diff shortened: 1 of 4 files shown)


tjefferson:
Addressed the Greptile review feedback in the second commit (b4c37e6):

- **`searchSouls` candidate pool floor**: Applied the same `50 → 200` fix to `searchSouls` (line 446), which had the identical bug pattern as described in the P1 comment.

All CI checks (build, secret scan, Vercel Agent Review) are passing. The Vercel deploy fail is the standard authorization-required for external contributors — same as previous PRs from this fork.

steipete:
Landed the current-path version of this fix on `main` in 5797057 (`fix(search): widen vector candidate pool`).

What landed:
- `convex/search.ts`: initial candidate pool floor raised to 200 for both skill search and soul search.
- `packages/clawhub/src/cli/commands/skills.ts`: CLI search now sends the web-aligned default `limit=25` when no explicit limit is provided.
- Regression coverage updated/added in `convex/search.test.ts` and `packages/clawhub/src/cli/commands/skills.test.ts`.
- Changelog credits this PR and @tjefferson.

Verified locally:
- `bun run test`
- `bun run coverage`
- `bun run lint`
- `bun run verify` in `packages/clawhub`
- `bun run build`

Closing this PR as superseded by the landed commit. Thanks for the original fix and detailed repro.

PR #1466, opened 2026-04-01
fix: improve admin user search coverage

ImLukeF:
## Summary

Admin user search was only scanning a recent slice of users when a query was present. That meant an older existing account could be missing from the admin Users page even when the search text matched exactly.

This PR changes searched admin listings to scan the full ordered user set, while keeping the default non-search listing capped by `limit`. It also adds a regression test covering an older matching user outside the previous recent scan window.

## Testing

Not run locally in this environment because `bun` is unavailable.

### DIFF
--- a/convex/users.test.ts
+++ b/convex/users.test.ts
@@ -1075,6 +1075,81 @@ describe("users.list", () => {
     expect(result.items[0]?.handle).toBe("alice");
   });
 
+  it("includes an exact older handle match outside the bounded scan", async () => {
+    vi.mocked(requireUser).mockResolvedValue({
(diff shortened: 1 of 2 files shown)


ImLukeF:
Rebased onto current `main`, resolved the conflicts, and reran the local gates.

What’s on the branch now is the bounded-scan version plus exact handle / personal publisher lookup for older exact matches. It does **not** do an unbounded full-table collect on admin search.

Local verification after rebase:
- `bun x vitest run convex/users.test.ts src/__tests__/package-detail-route.test.tsx`
- `bun run lint`
- `bun run build`

Issue #1483, opened 2026-04-02
Allow to view security from CLI


ibaraki-douji:
Could it be possible to add in the CLI a way to see the security status of a skill ?

Currently when i ask claw to install package, it's like 70% of the time where the pacakge is not "Benign".

So if we could ask him to install package that are "Benign" only that would be a huge + for skill installation.

i've checked with `clawhub package inspect` and `clawhub inspect` with text and json output, but no mention of any security.

vincentkoc:
Closed as completed on main in https://github.com/openclaw/clawhub/commit/1461d0fcfcf1db373b1a56fdc802f57d64bfe36f.

`clawhub inspect` now includes skill moderation metadata in plain text and JSON output; version security metadata was already present.

PR #1488, opened 2026-04-02
[fix] show moderation status on plain clawhub inspect

edenfunf:
## What's the problem?

Running `clawhub inspect <skill>` without any flags never shows the security status, even though the API always returns it.

If you try with `--version 1.2.3` you'll see the security block — but that requires knowing a version number upfront. For a quick sanity check, you shouldn't have to do that.

Related to #1483.

## Why does it happen?

The existing `printSecuritySummary` call is gated behind `versionResult?.version`:

```ts
if (shouldPrintMeta && versionResult?.version) {
  printVersionSummary(versionResult.version);
  printSecuritySummary(versionResult.version);   // never runs on plain inspect
}
```

`versionResult` is only populated when `--version`, `--tag`, `--files`, or `--file` is passed. Without those flags it stays `null`, and the skill's `moderation` object — which the API always returns — is fetched but never displayed.

## What I changed

Added a `printModerationSummary` function that reads the skill-level `moderation` object (`verdict`, `isSuspicious`, `isMalwareBlocked`, `updatedAt`, `engineVersion`) and prints it in the `else if (shouldPrintMeta)` branch, so the security line is always visible on a basic inspect:

```
$ clawhub inspect some-skill
some-skill  Some Skill
Owner: someone
...
Security: SUSPICIOUS
Suspicious: yes
Checked: 2024-01-15T10:30:00.000Z
Engine: v2
```

When no moderation data comes back from the API the block is a no-op, so existing behaviour for older responses is unchanged.

## How I verified it

- Added three unit tests to `inspect.test.ts`:
  1. Suspicious skill with full moderation data → all fields printed correctly
  2. Clean skill → `Security: CLEAN`, no suspicious/malware lines
  3. No moderation field in response → security block not printed at all
- Ran the full clawdhub test suite: **176/176 passed**

### DIFF
--- a/packages/clawhub/src/cli/commands/inspect.test.ts
+++ b/packages/clawhub/src/cli/commands/inspect.test.ts
@@ -218,6 +218,36 @@ describe("cmdInspect", () => {
     expect(securityCalls).toHaveLength(0);
   });
 
+  it("falls back to boolean flags when verdict is absent", async () => {
+    httpMocks.apiRequest.mockResolvedValueOnce({
+      skill: {
+        slug: "demo",
+        displayName: "Demo",
+        summary: null,
+        tags: { latest: "1.0.0" },
+        stats: {},
+        createdAt: 1,
+        updatedAt: 2,
+      },
+      latestVersion: { version: "1.0.0", createdAt: 3, changelog: "init", license: null },
+      owner: null,
+      moderation: {
+        isSuspicious: true,
+        isMalwareBlocked: false,
+        // verdict intentionally absent
+        reasonCodes: [],
+        updatedAt: null,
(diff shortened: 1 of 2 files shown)


ImLukeF:
Thanks for the thoughtful fix here — the behavior you pointed out makes sense.

We’re going to close this one for now because we don’t want to add more default output to plain `clawhub inspect` unless it’s clearly necessary. For the common case, always surfacing the moderation/security block would add noise to what is meant to be a quick metadata view.

So this is less about the implementation and more about the product choice: we’d rather keep the default inspect output lean unless we decide to introduce a more explicit security-focused view later.

Appreciate the contribution.

PR #1531, opened 2026-04-05
fix: use indexed query for soul list to avoid over-fetching

goulonghui:
The souls.list query previously used take(limit * 5) followed by JS filtering on softDeletedAt, reading up to 5× more documents than needed.

- For the ownerUserId branch: use Convex .filter() on the by_owner index to skip deleted rows before truncating.
- For the browse branch: use the by_active_updated compound index with eq(softDeletedAt, undefined) to skip deleted rows at the index level.

This reduces document reads and bandwidth, especially as the souls table grows.

### DIFF
--- a/convex/souls.ts
+++ b/convex/souls.ts
@@ -125,24 +125,27 @@ export const list = query({
     const limit = args.limit ?? 24;
     const ownerUserId = args.ownerUserId;
     if (ownerUserId) {
+      // No compound index on (ownerUserId, softDeletedAt) exists, so use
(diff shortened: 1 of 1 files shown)


ImLukeF:
Thanks for working on this. The underlying over-fetch issue is real, but this branch was blocked by the missing `souls.by_active_updated` schema index.

We picked up the fix in a smaller follow-up PR that includes the missing index and focused verification:
https://github.com/openclaw/clawhub/pull/1637

Closing this one in favor of that version so we keep one merge path open.

PR #1532, opened 2026-04-05
fix: replace Convex .filter() with indexed queries in skill transfers

goulonghui:
getPendingTransferBySkillAndUserInternal and
getPendingTransferBySkillAndFromUserInternal used .filter() on toUserId/fromUserId after the by_skill_status index, which scans every pending transfer for the skill and reads documents that are thrown away.

Switch to the by_to_user_status and by_from_user_status indexes respectively, then filter by skillId in JS.  A user will have at most a handful of pending transfers, so the JS check is effectively free while the index lookup is O(1).


ImLukeF:

PR #1533, opened 2026-04-05
refactor: deduplicate clampInt utility across convex modules

goulonghui:
Extract the clampInt helper into convex/lib/math.ts and replace 11 duplicate definitions across the codebase.  The previous copies used five different rounding strategies (Math.round, Math.trunc, Math.floor, Number.isFinite guard, or none).  The canonical version uses Math.trunc with a Number.isFinite guard, matching the most defensive variant from maintenance.ts.

Files updated:
  - convex/souls.ts
  - convex/commentModeration.ts
  - convex/leaderboards.ts
  - convex/users.ts
  - convex/statsMaintenance.ts
  - convex/githubBackups.ts
  - convex/githubBackupsNode.ts
  - convex/githubSoulBackups.ts
  - convex/githubSoulBackupsNode.ts
  - convex/maintenance.ts
  - convex/skills.ts

### DIFF
--- a/convex/commentModeration.ts
+++ b/convex/commentModeration.ts
@@ -4,6 +4,7 @@ import type { Id } from "./_generated/dataModel";
 import type { ActionCtx, MutationCtx } from "./_generated/server";
 import { action, internalAction, internalMutation, internalQuery } from "./functions";
 import { assertRole, requireUserFromAction } from "./lib/access";
+import { clampInt } from "./lib/math";
 import {
   buildCommentScamBanReason,
   isCertainScam,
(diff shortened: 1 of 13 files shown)


vincentkoc:
Thanks for the cleanup branch. I am closing this rather than landing it.

The shared `clampInt` refactor changes behavior for existing `skills.ts` and `souls.ts` pagination paths (`Math.round` to `Math.trunc`) while touching a broad set of Convex modules. That is too much semantic risk for a dedupe cleanup.

If we revisit this later, it should preserve each call site’s current rounding semantics or split helpers by rounding behavior.

PR #1560, opened 2026-04-06
fix: default workdir to ~/.openclaw instead of cwd

gregb100:
## Problem

clawhub install defaults to cwd, which means where you run it from determines where skills land. Skills from two completely different sources (OpenClaw team bundled skills vs custom clawhub skills) end up mixed in one directory with no way to tell them apart.

## Solution

Change the default workdir fallback from `cwd` to `~/.openclaw`. Skills installed via clawhub now land in `~/.openclaw/<skill>` by default instead of `cwd/skills/`.

## Priority order (unchanged except for last item)

1. Explicit `--workdir` flag
2. `CLAWHUB_WORKDIR` / `CLAWDHUB_WORKDIR` env vars
3. `.clawhub` marker directory in cwd
4. clawdbot workspace (if set)
5. `~/.openclaw` (was `cwd`)

### DIFF
--- a/packages/clawhub/src/cli.ts
+++ b/packages/clawhub/src/cli.ts
@@ -104,7 +104,9 @@ async function resolveWorkdir(explicit?: string) {
   if (hasMarker) return cwd;
 
   const clawdbotWorkspace = await resolveClawdbotDefaultWorkspace();
-  return clawdbotWorkspace ? resolve(clawdbotWorkspace) : cwd;
+  if (clawdbotWorkspace) return resolve(clawdbotWorkspace);
+  // Default to ~/.openclaw instead of cwd to avoid mixing npm skills with custom skills
+  return resolve(process.env.HOME ?? "", ".openclaw");
 }
 
 async function hasClawhubMarker(workdir: string) {


momothemage:
Thanks for the contribution! We really appreciate your effort and the work you've put into this PR. 

A small question: what is the actual reason for separating skills into two distinct paths for management? It seems unnecessary to manage the same type of resource in two separate paths.

gregb100:
no worries.
~/.npm-global/lib/node_modules/openclaw/skills/ — 55 bundled skills (OpenClaw team), none used
• ~/workspaces/otto/skills/ — 27 custom clawhub-installed skills, all in use
• The two paths are completely separate by design, but clawhub install defaults to cwd, so skills land wherever you ran the command

Why the split exists: OpenClaw's skill loader searches multiple directories (openclaw-extra, openclaw-workspace, openclaw-user), and the workspace path lets different agents have different skill sets. clawhub defaults to cwd for portability.

Why it's messy:

1. clawhub install scatters skills wherever you happen to run it from
2. The symlink clara → ~/projects/clara/skills triggers the sandboxing warning because it resolves outside the workspace root — that's the Skipping skill path that resolves outside its configured root WARN you see 2x every ~30s
3. No clean separation between "OpenClaw-bundled" vs "custom-installed" skills

My proposed fix: ~/.openclaw/ as clawhub's default would give a predictable, centralized location for custom skills that's:

• Not workspace-dependent
• Not mixed with npm-bundled skills
• Easy to back up / manage as a unit

ImLukeF:
Thanks for the thoughtful write-up here — the underlying pain makes sense.

We’re going to close this one for now because changing the default install location is a broader product decision than a small CLI fallback tweak. It affects default behavior for everyone, and the problem it’s addressing is tied pretty closely to OpenClaw’s overall local skill-path/layout conventions rather than a narrow ClawHub bug.

So this is less “bad idea” and more “needs a larger design decision if we ever do it.” We don’t want to change the default install location right now.

Appreciate the contribution and the clear explanation of the problem.

PR #1562, opened 2026-04-06
fix: allow skill owner to move skill to an org acct while publishing existing skill

### LINKED ISSUE #1431: Cannot change an existing skill's publisher/owner to an organization
When publishing a skill, the UI shows a dropdown to select the owner (personal account vs. organization). However, this only works for new skills. If you try to publish a new version of an existing skill and change the owner to an organization you belong to, you get:                                                                                                   
                  
  **▎ Slug is already taken. Choose a different slug. Existing skill: /pushmatrix/shop**     

This is currently happening to me as I am trying to change ownership from skill shop from my personal acct (pushmatrix) to my org account (shopify)                                  
  
**Potential Root cause:**                                                                                                                
                  
  insertVersion in convex/skills.ts compares the existing skill's ownerPublisherId against the one being published with. If  
  they differ, it treats it as a slug collision — even if the same user owns both the skill and the target org.
                                                                                                                             
  if (skill && skill.ownerPublisherId && skill.ownerPublisherId !== ownerPublisherId) {
    throw new ConvexError(buildSlugTakenErrorMessage(skill, owner));                                                         
  }

**Steps to reproduce**
- Create an org account
- Publish a new skill under your personal publisher
- Go to edit the skill and change the owner to the org account
- You get the message Slug is already taken. Choose a different slug. Existing skill: /personalacct/skillname
                                                                                                                                                                                          
**Expected behavior:**                                                                                                         
                  
If a user owns a skill and is a member of a target org (with publisher role or above), they should be able to move the skill to that org — either through the publish flow or a dedicated action.

pushmatrix:
Fixes: https://github.com/openclaw/clawhub/issues/1431

## Summary

- When republishing a skill under a different publisher (e.g. moving from personal to org), `insertVersion` incorrectly threw "Slug is already taken" even when the caller owns the skill
- The check at `convex/skills.ts:5820` treated any `ownerPublisherId` mismatch as a conflict with no escape hatch for the actual owner
- Now, if the caller is the `ownerUserId`, the skill's publisher is updated instead of throwing
- Slug aliases are also updated to stay in sync (matching the pattern used by the transfer flow)

## Test plan

- [x] Added test: owner can move skill from personal publisher to org publisher
- [x] Added test: non-owner is still blocked from claiming a skill with a different publisher  
- [x] Added test: slug aliases are updated when moving skill to a different publisher
- [ ] Manual: republish an existing skill under a different org via the web UI

### DIFF
--- a/convex/skills.rateLimit.test.ts
+++ b/convex/skills.rateLimit.test.ts
@@ -1357,4 +1357,305 @@ describe("skills anti-spam guards", () => {
       }),
     );
   });
+
+  it("allows owner to move skill from personal publisher to org publisher", async () => {
+    const patch = vi.fn(async () => {});
+    const insert = vi.fn(async (table: string) => {
+      if (table === "skillVersions") return "skillVersions:1";
+      if (table === "skillEmbeddings") return "skillEmbeddings:1";
+      if (table === "embeddingSkillMap") return "embeddingSkillMap:1";
+      if (table === "skillVersionFingerprints") return "skillVersionFingerprints:1";
+      throw new Error(`unexpected insert table ${table}`);
+    });
+    const existingSkill = {
+      _id: "skills:1",
+      slug: "shop",
+      displayName: "Shop",
+      summary: "A skill",
+      ownerUserId: "users:owner",
+      ownerPublisherId: "publishers:personal",
+      latestVersionId: undefined,
+      tags: {},
+      softDeletedAt: undefined,
+      badges: { redactionApproved: undefined, highlighted: undefined, official: undefined, deprecated: undefined },
+      moderationStatus: "active",
+      moderationReason: "pending.scan",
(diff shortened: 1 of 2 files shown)


pushmatrix:
@momothemage good catch! Should be all fixed now

pushmatrix:
@momothemage oops, I see the types were failing. That's fixed now. Might running CI again?

ImLukeF:
Thanks for digging into this. The underlying problem is real, but we’re going to handle ownership changes through the dedicated transfer flow instead of teaching the publish path to move ownership as a side effect.

We’ve got the broader transfer work open here:
https://github.com/openclaw/clawhub/pull/1635

I’m going to close this one off in favor of that approach so we keep ownership changes explicit and easier to reason about.

PR #1599, opened 2026-04-09
fix: flag destructive troubleshooting delete commands

ImLukeF:
## Summary

This updates skill moderation to catch the `#1592` class of issue: destructive troubleshooting commands in public `SKILL.md` docs that an agent could execute without an explicit user confirmation gate.

## What changed

- add a new static moderation rule for destructive delete commands in markdown docs
- flag `rm -rf` only when it appears in troubleshooting/reinstall-style guidance
- require the delete target to be a sensitive absolute/home-style path
- suppress the finding when the docs clearly require user confirmation first
- do not flag routine project cleanup like `rm -rf node_modules dist`
- do not flag temp-directory cleanup like `/tmp/...`
- bump the moderation engine version so the rule is reflected in scan metadata

## Why

Issue #1592 showed that `stt-simple` includes a destructive reinstall path in its Troubleshooting section:

```bash
rm -rf /root/.openclaw/venv/stt-simple
/root/.openclaw/workspace/skills/stt-simple/install.sh
```

In an agent context, troubleshooting sections are executable guidance. Destructive reinstall steps against real user/system paths should be gated behind explicit user confirmation.

## Validation

- `bun test convex/lib/moderationEngine.test.ts`

## Notes

`bunx tsc -p tsconfig.json --noEmit` is currently blocked in this clean worktree by baseline type-resolution issues (`vite/client`, `vitest/globals`), not by this change.

### DIFF
--- a/convex/lib/moderationEngine.test.ts
+++ b/convex/lib/moderationEngine.test.ts
@@ -144,6 +144,291 @@ describe("moderationEngine", () => {
     expect(result.status).toBe("clean");
   });
 
+  it("flags destructive rm -rf troubleshooting commands without confirmation", () => {
+    const result = runStaticModerationScan({
+      slug: "stt-simple",
+      displayName: "STT Simple",
+      summary: "Simple speech-to-text setup",
+      frontmatter: {},
+      metadata: {},
+      files: [{ path: "SKILL.md", size: 512 }],
+      fileContents: [
+        {
+          path: "SKILL.md",
+          content: [
+            "## Troubleshooting",
+            "",
+            "```bash",
+            "# Reinstall",
+            "rm -rf /root/.openclaw/venv/stt-simple",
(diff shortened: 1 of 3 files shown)


vincentkoc:
Thanks for the report and patch. This class is now covered on main by direct maintainer fixes:

- b90a43adcbc91e3207ad14f3fcae1c0c30c7bbc2 added `suspicious.destructive_delete_command` coverage for the `stt-simple` pattern.
- 4a9ae92 tightened the confirmation-gate heuristic so loose prose like “Confirm the backup exists before proceeding” does not suppress the finding.

I also kept the reviewer gap around stale scan-version fixtures covered with the pending scan queue test.

Validation on main:

- `bun run test -- convex/lib/moderationEngine.test.ts convex/skills.pendingScanQueue.test.ts`
- `bun run lint`
- `git diff --check`

Closing this PR as superseded by the direct main fixes.

PR #1602, opened 2026-04-09
fix: transfer request fails with toUserHandle required

East5RingRoad-kyle:
## Problem
CLI `transfer request` command fails with "toUserHandle required" error due to double JSON.stringify in the request body.

## Root Cause
1. `transfer.ts` was using `JSON.stringify()` on the body
2. `apiRequest` also serializes the body
3. Result: server received stringified string instead of object, `toUserHandle` became undefined

## Solution
1. **transfer.ts**: Remove `JSON.stringify()`, pass plain object to `apiRequest`
2. **transfer.test.ts**: 
   - Fix undefined `mockApiRequest` → use `httpMocks.apiRequest`
   - Update assertion to expect object instead of stringified JSON

## Testing
bash
✓ 5 pass, 0 fail
✓ All transfer command tests passing

## Related
- Fixes test failures blocking PR #973
- Unblocks skill ownership transfer workflow
- Resolves issue where CLI transfer command fails with "toUserHandle required"

### DIFF
--- a/packages/clawhub/src/cli/commands/transfer.test.ts
+++ b/packages/clawhub/src/cli/commands/transfer.test.ts
@@ -64,8 +64,9 @@ describe("transfer commands", () => {
       }),
       expect.anything(),
     );
-    const requestArgs = httpMocks.apiRequest.mock.calls[0]?.[1] as { body?: string };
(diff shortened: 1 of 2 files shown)


ImLukeF:

Issue #1621, opened 2026-04-10
Cannot signin. First time trying to publish


ImMrLucky:
I am trying to login to ClawHub. 

I did get redirected to github and authorized it. ClawdHub shows as authorized in my github account.

Github username:
immrlucky

Now every time I click on Signin with Github button it just refreshes the page. I have tried using an incognito browser window as well. 

I see an a __convexAuthOAuthVerifier_httpswrymanatee359convexcloud
Key in Localstorage with a value

As well as a session storage and cookie.

Any help would be appreciated

raydoomed:
挂了，登录不了

ImMrLucky:
can anyone please help me create an account and login?

NoamBenS:
Seconded - same exact issue, also a first-time publisher.

Vitaliy-Gureyev:
I'm getting the same issue. I tried logging in through GitHub and switched browsers too, but no luck. The UI just shows "Sign in failed. Please try again." For context, this is first time trying to log into ClawHub.

Patrick-Erichsen:
Closing as no longer reproducible in the supported GitHub OAuth flow.

Maintainer verified a clean incognito browser sign-in on 2026-05-19 PT, and we also verified the CLI browser-login path by running:

```sh
clawhub logout
clawhub login
clawhub whoami
```

That completed successfully and returned the expected authenticated handle. Please open a fresh issue with current browser/network details if this still reproduces on the current production site.

Issue #1619, opened 2026-04-10
Attempting to publishing a skill. Stuck on "Sign in failed. Please try again."


jtcchan:
I'm trying to publish a skill but can't login in. I'm getting a generic - "Sign in failed. Please try again." Happens to both logging in from the web UI and running `clawdhub login` from the CLI.

Cleared cached, tried Chrome and Safari. Revoked token from Github and reconnected again. 

User is:
https://clawhub.ai/u/jtcchan

<img width="596" height="203" alt="Image" src="https://github.com/user-attachments/assets/5e054607-5b37-47df-acb0-8e1be32e3eb7" />

JonatanCosta:
Same here

lwch88690:
The same problem. I thought it was an account anomaly at first.

<img width="1078" height="268" alt="Image" src="https://github.com/user-attachments/assets/efa3e896-e436-41a1-b28a-635d006fc790" />

kevin3cx:
same here; fails with: chrome, chromium, safari. fails with a 5-year old account and a 5-hour old account.

cannot login to clawhub, cannot publish

ImMrLucky:
I am having a similar issue. I am trying to login for the first time with ClawHub and publish a skill.

vlbeta:
Same error in Chrome 146 on macOS, in incognito with zero extensions — reproduces every time, not Safari-specific, not extension-related, not stale-state-related.

Captured a HAR of the full post-authorize round trip. The failure happens in the Convex callback and matches the exact technical symptom in #1552:

**Request** (after clicking `Authorize` on github.com):

```
GET https://wry-manatee-359.convex.site/api/auth/callback/github?code=<gh_code>&iss=https%3A%2F%2Fgithub.com%2Flogin%2Foauth

referer: https://github.com/
sec-fetch-site: cross-site
sec-fetch-mode: navigate
sec-fetch-dest: document
(no Cookie header)
```

**Response:**

```
302 Found
location: https://clawhub.ai/publish-skill
content-length: 0
(no Set-Cookie)
(no ?code= or error param in the Location)
```

Key observations from the HAR:

1. **The callback request has zero cookies.** The browser sent no `Cookie` header at all. The PKCE state/verifier cookie that Convex Auth sets on `wry-manatee-359.convex.site` during the earlier `/api/auth/signin/github` hop isn't being returned to the callback endpoint.

[...]

PR #1623, opened 2026-04-11
Add issuer to GitHub authentication configuration

hpsin:
Rather than waiting on convex to pick up the fix from nextjs here: https://github.com/nextauthjs/next-auth/pull/13410

This pulls the issuer config up to your convex profile.  It's untested, I'm afraid, but happy to figure out how to build and deploy if it helps. 

This fixes the issues reported here: 
* https://github.com/openclaw/clawhub/issues/1621
* https://github.com/openclaw/clawhub/issues/1619

The root cause of those issues can be read about here: https://github.com/langfuse/langfuse/issues/13091

The net is that we rolled out a new security feature, but this uncovered a bug / unexpected behavior in nextJS and its component libraries. Apps attempted to validate the issuer value in the code response, but nextJS gives them nothing to validate it against, causing the validation to fail. By providing the issuer in your config, your app will now successfully validate the issuer value that we send (when we start sending it again, in the future).

### DIFF
--- a/convex/auth.ts
+++ b/convex/auth.ts
@@ -74,7 +74,7 @@ export const { auth, signIn, signOut, store, isAuthenticated } = convexAuth({
   providers: [
     GitHub({
       clientId: process.env.AUTH_GITHUB_ID ?? "",
-      clientSecret: process.env.AUTH_GITHUB_SECRET ?? "",,
+      clientSecret: process.env.AUTH_GITHUB_SECRET ?? "",
       issuer: "https://github.com/login/oauth",
       profile(profile) {
         return {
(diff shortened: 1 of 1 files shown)


ImLukeF:
Thanks for the fix and for digging into the upstream context here.

We’re going to close this one for now because GitHub sign-in is currently working on ClawHub, and we don’t want to add workaround auth configuration unless we can reproduce an active issue in the current app.

If this starts failing again or we can clearly reproduce the issuer-validation problem on current main, this is the sort of change we can revisit quickly. Appreciate you putting the investigation together.

hpsin:
Cheers, makes sense @ImLukeF . We'll be rolling this out again in about a month, we believe, after some more proactive community discussion to ensure that apps using auth.js can get updates and test themselves.

ImLukeF:
@hpsin Thanks mate, We are also undergoign a pretty big overhaul too, so we can probably look into it again in the next week, unless you think its worthing changing while we undergo it?

PR #1625, opened 2026-04-11
fix: enhance dark mode styles in SkillsToolbar component

geoffrey-xiao:
### Summary

The first two filter chips on the skills browse toolbar (“Staff Picks” and “Clean only”) only applied light-mode surface styles (white background and light-theme border). In dark mode they stayed bright and looked wrong next to the category and sort controls, which already use the shared dark surface tokens.

### Changes

- Update `FilterChip` in `src/routes/skills/-SkillsToolbar.tsx` with the same dark-mode border and background as other toolbar controls (`dark:border-[rgba(255,255,255,0.12)]`, `dark:bg-[rgba(14,28,37,0.84)]`).

### Testing

- Open the skills browse page in dark theme and confirm the two chips match the “All categories” / sort surfaces when inactive.
- Toggle each chip on and confirm the active (accent) state still reads clearly.

### Screenshots

before
image

after
image

### DIFF
--- a/src/routes/skills/-SkillsToolbar.tsx
+++ b/src/routes/skills/-SkillsToolbar.tsx
@@ -260,7 +260,7 @@ function FilterChip({
       type="button"
       aria-pressed={active}
       onClick={onClick}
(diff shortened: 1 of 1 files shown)


ImLukeF:

PR #1627, opened 2026-04-11
fix: Implement theme-aware syntax highlighting and adjust markdown styles

geoffrey-xiao:
## Summary

Fixes markdown code blocks on light theme by aligning Shiki’s syntax theme with the app’s resolved light/dark mode, and centralizes “which theme is active?” in `theme.ts`.

## Problem

`MarkdownPreview` always used Shiki with `github-dark`. That theme emits light token colors meant for dark backgrounds. On light mode, those colors sat on the light `.markdown pre` styling and were nearly unreadable.

## Changes

- **`src/components/MarkdownPreview.tsx`** — Chooses `github-light` vs `github-dark` based on the resolved theme and re-runs async highlighting when the theme changes.
- **`src/lib/theme.ts`** — Adds `getResolvedTheme()` and `useResolvedTheme()` that read `document.documentElement.dataset.theme` (what `applyTheme` sets) and subscribe via `MutationObserver` so theme toggles update consumers without a global React context.

## How to test

1. Open any page that renders `MarkdownPreview` with fenced code (e.g. skill detail **Files** tab / `SKILL.md`).
2. Use light theme: code inside fences should be clearly readable.
3. Switch to dark theme: highlighting should switch to the dark palette and remain readable.
4. Toggle back to light: markdown should re-highlight without a full reload.

## Screenshot
before
image

after
image

### DIFF
--- a/src/components/MarkdownPreview.tsx
+++ b/src/components/MarkdownPreview.tsx
@@ -1,6 +1,7 @@
 import { parse } from "@create-markdown/core";
 import { blocksToHTML, renderAsync, shikiPlugin } from "@create-markdown/preview";
 import { useEffect, useRef, useState } from "react";
+import { useResolvedTheme } from "../lib/theme";
 import { cn } from "../lib/utils";
 
 interface MarkdownPreviewProps {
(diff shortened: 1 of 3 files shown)


ImLukeF:

PR #1637, opened 2026-04-12
fix: reduce souls browse overfetch

ImLukeF:
## Summary
- add the missing `souls.by_active_updated` index so browse queries can target active souls directly
- update `souls.list` browse mode to read `limit` active rows instead of `limit * 5` rows filtered in JS
- add focused regression coverage for the browse-path query shape
- rename a shadowed local in `src/routes/cli/auth.tsx` so lint passes cleanly

## Why
The previous browse listing path over-read soul documents and then filtered soft-deleted rows in application code. That meant requests like `/souls` with `limit=500` could read up to 2500 rows to return 500 visible items.

This patch keeps the behavior the same while making the browse path ask Convex for active rows directly.

## Verification
- `bun run lint`
- `bun x vitest run convex/souls.test.ts src/routes/cli/-auth.test.ts`
- `bun run build`

## Notes
- This supersedes the approach in #1531 with a smaller patch that includes the missing schema index.
- `bunx convex codegen` is blocked in this shell because no `CONVEX_DEPLOYMENT` is configured here.

### DIFF
--- a/convex/schema.ts
+++ b/convex/schema.ts
@@ -384,7 +384,8 @@ const souls = defineTable({
   .index("by_slug", ["slug"])
   .index("by_owner", ["ownerUserId"])
   .index("by_owner_publisher", ["ownerPublisherId"])
-  .index("by_updated", ["updatedAt"]);
+  .index("by_updated", ["updatedAt"])
+  .index("by_active_updated", ["softDeletedAt", "updatedAt"]);
 
 const skillVersions = defineTable({
   skillId: v.id("skills"),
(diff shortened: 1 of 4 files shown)


ImLukeF:
Thanks. I checked the ordering concern before merging.

For `/souls`, the route fetches the browse set and then applies client-side sorting, so the server-side browse order is not what drives that screen. Separately, the paginated public listing path already uses `by_updated`, so using the active+updated index here is acceptable and consistent with the broader soul listing direction.

I’m treating the owner-branch overfetch as follow-up scope, not a blocker for this patch.

PR #1657, opened 2026-04-13
feat: detect phantom dependencies in skill bundle files

I3eg1nner:
## Summary

Adds a new async security scanner that parses dependency files bundled in skills (`requirements.txt`, `package.json`, `Cargo.toml`, `pyproject.toml`) and verifies every referenced package exists on its public registry (PyPI / npm / crates.io). Non-existent packages are reported as a **Dependency Confusion via Phantom Package** supply-chain risk.

## What changed and why

Skills can bundle dependency files that trigger automatic package installation. An attacker could reference phantom (non-existent) package names in these files, then register those names on the public registry with malicious code — a classic Dependency Confusion attack. This scanner detects that vector at publish time.

**New file — `convex/depRegistryScan.ts`:**
- Parsers for `requirements.txt` (incl. `-dev`/`-test` variants), `package.json`, `Cargo.toml`, `pyproject.toml`
- Registry existence check via public APIs with retry/exponential backoff, 8s timeout, and abort after 3 consecutive failures
- `depRegistryCache` table deduplicates lookups across skills (30-day TTL for existing packages, 7-day for missing)
- `rescanErrorDepRegistryVersions` action for retrying network-failure scans

**Modified files:**
- `convex/schema.ts` — `depRegistryAnalysis` field on `skillVersions`, new `depRegistryCache` table
- `convex/lib/moderationReasonCodes.ts` — `DEP_NOT_FOUND` reason code
- `convex/skills.ts` — `updateVersionDepRegistryAnalysisInternal` mutation: stores results, injects a single consolidated critical finding into `staticScan`, triggers moderation rebuild
- `convex/lib/skillPublish.ts` — schedules the scan after publish
- `src/components/SkillSecurityScanResults.tsx` — hides `:0` suffix when finding line is zero

## Test commands

```bash
bun run lint       # 0 warnings, 0 errors
bun run test       # 965/965 passed
bun run build      # successful
bun run --cwd packages/clawhub verify   # all passed
```

## Screenshot

Manual end-to-end test: seeded a skill with `requirements.txt`, `package.json`, `Cargo.toml`, and `pyproject.toml` containing 6 phantom packages across 3 registries — all correctly detected, consolidated into one critical finding.

Manual test screenshot

## AI-Generated Code

This PR was generated with AI assistance (Claude). The implementation:
- Was designed through iterative review of the existing scanner architecture (VT, LLM eval) to follow established `internalAction` → `internalMutation` → `patchStructuredModerationFromVersion` patterns
- All four CI-equivalent checks pass with zero regressions
- Was manually end-to-end tested against live PyPI/npm/crates.io registries
- The author understands and can maintain the code

## Why create the feat?

I have been in contact with ClawHub's security email address, and at Peter's suggestion, I am participating in the remediation efforts for this supply chain risk. Below is the background:

1. Findings

After scanning all dependency files (requirements.txt, package.json, Cargo.toml, etc.) across every skill and verifying each package name against official registry APIs, I found that out of 2,847 unique dependencies, 153 do not exist on their respective registries (PyPI/npm/crates.io).

2. Risks

These non-existent package names stem from two causes, but both lead to the same threat:

First, vibe coding hallucinations. Many skills are AI-generated, and LLMs fabricate plausible but non-existent package names — such as baoyu-chrome-cdp and windows-rpa — or mistakenly put standard library modules (json, math) and PyPI package names (Pillow, opencv-python) into package.json.

Second, deliberate attacker construction. Some skills may be intentionally designed to reference non-existent package names, after which the attacker registers the same name as a malicious package on the registry — a classic Dependency Confusion attack.

Regardless of the cause, the risk is the same: anyone can claim these vacant package names and inject malicious code. As proof, I successfully registered a harmless placeholder package under the vacant name pinyin-box (https://pypi.org/project/pinyin-box/0.0.1/) — the entire process took less than one minute.

As for the severity: taking PyPI as an example, an attacker only needs to plant arbitrary code in setup.py's install hook (e.g., a reverse shell, or exfiltrating environment variables and secrets). The code runs automatically when a user executes pip install — no explicit import required. Since PyPI has no pre-publication review process, a malicious package becomes installable immediately after upload. Although PyPI's malware detection and community reporting typically take it down within 0.5 to 1 day, that window is more than enough for an effective attack — especially when the skill has active install users. The same applies to npm.

No confirmed exploitation has been found so far, but the audit revealed high-download skills that reference non-existent npm package names. Once an attacker claims these names, all new installations and updates will be immediately affected. Every additional day this window remains open adds to the accumulated risk. I recommend taking action as soon as possible.

### DIFF
--- a/convex/depRegistryScan.ts
+++ b/convex/depRegistryScan.ts
@@ -82,9 +82,10 @@ function parsePackageJson(content: string, path: string): DepEntry[] {
       const deps = pkg[field];
       if (deps && typeof deps === "object" && !Array.isArray(deps)) {
         for (const name of Object.keys(deps as Record<string, unknown>)) {
-          // Skip scoped packages pointing to local paths/URLs
+          // Skip non-registry version specifiers (local, git, URL, workspace)
           const ver = (deps as Record<string, string>)[name] ?? "";
-          if (ver.startsWith("file:") || ver.startsWith("link:")) continue;
+          const NON_REGISTRY = ["file:", "link:", "git+", "git://", "github:", "bitbucket:", "gist:", "http:", "https://", "workspace:"];
+          if (NON_REGISTRY.some((p) => ver.startsWith(p))) continue;
           entries.push({ name: name.toLowerCase(), registry: "npm", source: path });
         }
       }
@@ -204,7 +205,12 @@ async function checkRegistryExists(
   registry: SupportedRegistry,
   packageName: string,
 ): Promise<{ exists: boolean; httpStatus: number } | { exists: null; httpStatus: null }> {
-  const url = REGISTRY_ENDPOINTS[registry].replace("{name}", encodeURIComponent(packageName));
+  // npm scoped packages (@scope/name) need the @ literal, only the rest encoded.
+  const encodedName =
+    registry === "npm" && packageName.startsWith("@")
+      ? `@${encodeURIComponent(packageName.slice(1))}`
+      : encodeURIComponent(packageName);
+  const url = REGISTRY_ENDPOINTS[registry].replace("{name}", encodedName);
   const headers: Record<string, string> = {
     Accept: "application/json",
   };
@@ -353,6 +359,7 @@ export const checkDependencyRegistries = internalAction({
       name: string;
       exists: boolean;
       httpStatus?: number;
+      networkError?: boolean;
     }> = [];
     let consecutiveFailures = 0;
     let abortedDueToErrors = false;
@@ -382,9 +389,9 @@ export const checkDependencyRegistries = internalAction({
       const check = await checkRegistryExists(dep.registry, dep.name);
 
       if (check.exists === null) {
-        // Network error — count towards abort threshold
+        // Network error — count towards abort threshold but do NOT mark as missing.
         consecutiveFailures++;
-        results.push({ registry: dep.registry, name: dep.name, exists: false });
+        results.push({ registry: dep.registry, name: dep.name, exists: true, networkError: true });
         if (consecutiveFailures >= CONSECUTIVE_FAILURE_ABORT) {
           console.warn(
             `depRegistryScan: aborting after ${CONSECUTIVE_FAILURE_ABORT} consecutive failures for version ${args.versionId}`,
@@ -416,7 +423,7 @@ export const checkDependencyRegistries = internalAction({
     }
 
     const notFoundPackages = results
-      .filter((r) => !r.exists)
+      .filter((r) => !r.exists && !r.networkError)
       .map((r) => `${r.name} (${r.registry})`);
 
     let status: "clean" | "suspicious" | "error";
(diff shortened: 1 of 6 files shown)


I3eg1nner:
Thanks for the review @clawsweeper. Confirming that each of the 8 concerns flagged for maintainer attention is already covered by the current branch — citing specific locations in case it helps the human review:

| Concern | Implementation |
|---|---|
| **Scoped npm packages** | `checkRegistryExists` keeps the `@` literal and only encodes the inner path (`depRegistryScan.ts:328-330`) — fixes the `%40scope%2Fname` 404 bug |
| **Transient registry failures** | Per-request 8s `AbortController`, exponential backoff retry up to 3 attempts on 429/5xx (`MAX_RETRIES=3`, `BACKOFF_BASE_MS=2000`); on exhausted retries, packages are **excluded from results** rather than marked missing (`depRegistryScan.ts:506-518`); 3 consecutive failures abort the scan and yield `status=\"error\"` for retry |
| **npm aliases** | `npm:` prefix added to `NON_REGISTRY` skip-list in `parsePackageJson` (`depRegistryScan.ts:89`) |
| **URL/workspace specs** | Same skip-list covers `file:`, `link:`, `git+`, `git://`, `github:`, `bitbucket:`, `gist:`, `http:`, `https://`, `workspace:`, `npm:` (`depRegistryScan.ts:89`); `requirements.txt` parser also skips `pkg @ git+...` direct-URL form via `if (/\s@\s/.test(line)) continue;` (`depRegistryScan.ts:64`) |
| **Optional dependencies** | `parsePackageJson` iterates `[\"dependencies\", \"devDependencies\", \"optionalDependencies\"]` (`depRegistryScan.ts:83`) |
| **pyproject formats** | `parsePyprojectToml` handles both PEP 621 (`dependencies = [...]`) and Poetry tables (`[tool.poetry.dependencies]`, `[tool.poetry.dev-dependencies]`, `[tool.poetry.group.dev.dependencies]`) (`depRegistryScan.ts:131-176`) |
| **Vendored dependency files** | `VENDORED_PATH_PATTERNS` skips files under `node_modules/`, `vendor/`, `__pycache__/`, `.venv/`, `venv/`, `target/`, `.cargo/` (`depRegistryScan.ts:198-214`) |
| **Indexed retry/backfill** | `depRegistryCache` table indexed on `[registry, name]` for O(1) lookup; `rescanErrorDepRegistryVersions` action drains error-state versions in batches with 2s stagger (`depRegistryScan.ts:619-668`). The maintenance query uses JS-filter on a nested field — annotated with rationale at line 636-639 since it's a low-frequency retry path |

Additional capabilities not in your list but added during review:
- npm scope registration check: when `@scope/name` is missing, also verifies whether `@scope` itself is registered (free-text search + client-side `@scope/` prefix filter, since v1 search doesn't honor `scope:`); unregistered scope is reported as **higher risk** in the consolidated finding (`depRegistryScan.ts:248-313`)
- All not-found packages are merged into a **single critical** staticScan finding titled \"Dependency Confusion via Phantom Package\" rather than per-package noise (`skills.ts:4604-4625`)

Test commands (all green on this branch):
\`\`\`
bun run lint                              # 0 warnings, 0 errors
bun run test                              # 965/965 passed
bun run build                             # successful
bun run --cwd packages/clawhub verify     # 178/178 passed
\`\`\`

End-to-end verified: vendored phantoms in \`vendor/somepkg/package.json\` and \`node_modules/somelib/package.json\` are correctly ignored; \`@deffinitely-not-real-scope-xyz\` is correctly identified as unregistered while real scope \`@types\` is not flagged.

vincentkoc:
thanks for raising the dependency-confusion scanner work. i landed a maintainer rebuild directly on main as a535da6 with safer unresolved-lookup semantics, generated Convex refs, and indexed retry state, so i am closing this PR to avoid keeping a duplicate implementation lane open.

Issue #1717, opened 2026-04-17
Bug: /api/auth/signin/github returns HTTP 500 — GitHub OAuth broken, no new tokens can be generated


jeffvsutherland:
## Summary
GET /api/auth/signin/github returns HTTP 500 from both the Vercel frontend and the Convex backend.
This breaks `clawhub login` completely — no new clh_ tokens can be generated.
## Reproduce
curl -i https://clawhub.ai/api/auth/signin/github        # 500
curl -i https://wry-manatee-359.convex.site/api/auth/signin/github  # 500
clawhub login → browser opens → circular redirect loop (500 causes retry loop)
## Suspected cause
AUTH_GITHUB_ID / AUTH_GITHUB_SECRET missing or expired in the Convex deployment.
## Impact
All authenticated CLI operations blocked: publish, whoami, sync.
Read-only operations (search, explore, inspect) still work without auth.
## Request
Check/revalidate the GitHub OAuth app credentials in your Convex deployment.
Environment: clawhub v0.9.0, macOS 15.4, 2026-04-17

Patrick-Erichsen:
Fresh live repro from production on 2026-05-19 17:01 PT:

```sh
curl -sS -D - -o /tmp/clawhub-auth-signin-body.txt https://clawhub.ai/api/auth/signin/github
curl -sS -D - -o /tmp/clawhub-auth-signin-convex-body.txt https://wry-manatee-359.convex.site/api/auth/signin/github
```

Both the canonical site and Convex site returned `HTTP/2 500` with an empty response body. This matches the reported failure mode and should be treated as the representative production auth incident for the broader login-loop bucket until the sign-in endpoint returns a normal OAuth redirect again.

Patrick-Erichsen:
Closing as not reproducible from the supported login flow.

`/api/auth/signin/github` is a Convex Auth route mounted by `auth.addHttpRoutes`, but it is not intended to be called directly without the verifier generated by the `auth:signIn` action. A bare `curl https://clawhub.ai/api/auth/signin/github` returning 500 is not the same as the browser/CLI sign-in flow failing.

Verified locally on 2026-05-19 PT:

```sh
clawhub logout
clawhub login
clawhub whoami
```

Result: browser login completed, a new token was stored, and `whoami` returned `patrick-erichsen-2`.

PR #1720, opened 2026-04-17
fix(auth): fail fast when AUTH_GITHUB_ID/SECRET env vars are missing

jeffvsutherland:
## Problem

`/api/auth/signin/github` returns HTTP 500, breaking `clawhub login` for all users. Tracked in #1717.

## Root Cause

In `convex/auth.ts`, missing env vars silently fall back to empty strings:

```typescript
clientId: process.env.AUTH_GITHUB_ID ?? "",
clientSecret: process.env.AUTH_GITHUB_SECRET ?? "",
```

Auth.js receives empty credentials, fails internally, and Convex surfaces this as an unhandled HTTP 500 with no useful diagnostic information.

## Fix

Replace the `?? ""` fallbacks with a `missingEnvVar()` helper that throws immediately with a clear message pointing to the Convex dashboard:

```
Error: Missing required environment variable: AUTH_GITHUB_ID.
Set it in the Convex dashboard under Settings → Environment Variables.
```

This makes misconfigured deployments fail loudly at startup instead of silently at request time.

## To Fix the Live Deployment

In addition to merging this PR, the `AUTH_GITHUB_ID` and `AUTH_GITHUB_SECRET` values need to be set in the Convex dashboard:

1. Go to **Convex Dashboard → your deployment → Settings → Environment Variables**
2. Add `AUTH_GITHUB_ID` and `AUTH_GITHUB_SECRET` from your GitHub OAuth App
3. Redeploy

The GitHub OAuth App credentials can be found at: **GitHub → Settings → Developer Settings → OAuth Apps**

### DIFF
--- a/.env.test
+++ b/.env.test
@@ -1,0 +1,4 @@
+# Stub values so convex/auth.ts module initialization doesn't throw during tests.
+# These are not real credentials — GitHub OAuth is not exercised in the test suite.
+AUTH_GITHUB_ID=test-github-client-id
+AUTH_GITHUB_SECRET=test-github-client-secret
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,9 +27,15 @@ jobs:
 
       - name: Test
         run: bun run test
+        env:
+          AUTH_GITHUB_ID: test-github-client-id
+          AUTH_GITHUB_SECRET: test-github-client-secret
 
       - name: Coverage
         run: bun run coverage
+        env:
+          AUTH_GITHUB_ID: test-github-client-id
+          AUTH_GITHUB_SECRET: test-github-client-secret
 
       - name: ClawHub CLI Verify
         run: bun run --cwd packages/clawhub verify
--- a/convex/auth.ts
+++ b/convex/auth.ts
@@ -70,11 +70,18 @@ export async function handleDeletedUserSignIn(
   throw new ConvexError(DELETED_ACCOUNT_REAUTH_MESSAGE);
 }
 
+function missingEnvVar(name: string): never {
+  throw new Error(
+    `Missing required environment variable: ${name}. ` +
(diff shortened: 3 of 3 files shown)


jeffvsutherland:
Good catch. You're right — the `?? missingEnvVar()` calls fire at module init time, not at request time, so importing `./auth` in tests crashes immediately when the env vars aren't set.

Updated the PR with **test-safe env setup** (your suggested path):

- Added `.env.test` with stub values — Vitest auto-loads this file, so `bun run test` works locally without needing real credentials
- For CI (`.github/workflows/ci.yml`), the Test and Coverage steps need `AUTH_GITHUB_ID` and `AUTH_GITHUB_SECRET` added as `env:` entries (same stub values). I couldn't push that change directly — my PAT lacks `workflow` scope — but the diff is straightforward:

```yaml
- name: Test
  run: bun run test
  env:
    AUTH_GITHUB_ID: test-github-client-id
    AUTH_GITHUB_SECRET: test-github-client-secret

- name: Coverage
  run: bun run coverage
  env:
    AUTH_GITHUB_ID: test-github-client-id
    AUTH_GITHUB_SECRET: test-github-client-secret
```

Happy to apply that directly if you grant write access, or you can apply it yourself.

jeffvsutherland:
CI step env vars pushed in 17f0671 — now has `workflow` scope. PR should be complete.

vincentkoc:
Thanks for taking this on. I am closing this branch rather than landing it as-is.

The current approach validates GitHub OAuth env at `convex/auth.ts` module load, which makes imports and non-auth test paths brittle. Adding `.env.test`/CI stubs works around that symptom but keeps the production behavior coupled to module initialization.

We should solve this with an import-safe/request-time diagnostic or a deploy/env validation path, not an eager throw in the auth module.

PR #1753, opened 2026-04-20
fix(api): use listPublicPageV4 in v1 skills endpoint

### LINKED ISSUE #1722: `/api/v1/skills` returns empty `{items:[]}` for all callers — `clawhub explore` affected
## Summary

`GET /api/v1/skills` currently returns an identical empty response (`{"items":[],"nextCursor":null}`, 30 bytes) for every request, regardless of auth state. Official CLI `clawhub@0.9.0` `explore` prints "No skills found." Sibling endpoints are fine — the same token can call `/api/v1/whoami` and `/api/v1/search` successfully.

## Reproduction

Tested against `https://clawhub.ai` on 2026-04-17.

| Auth | ratelimit-remaining | status | body |
|---|---|---|---|
| no Authorization header | 178/180 | 200 | `{"items":[],"nextCursor":null}` |
| `Authorization: Bearer deliberately-wrong-xxx` | 177/180 | 200 | `{"items":[],"nextCursor":null}` |
| `Authorization: Bearer <valid CLI token>` | 899/900 | 200 | `{"items":[],"nextCursor":null}` |
| same valid token → `/api/v1/search?q=file` | 898/900 | 200 | returns 10+ scored results |

All of the following query variants produce the same empty body:

- `?limit=5`
- `?limit=100&nonSuspiciousOnly=false`
- `?limit=5&nonSuspicious=false` (legacy alias)
- every `--sort` value accepted by `clawhub explore` (newest / downloads / rating / installs / installsAllTime / trending)

## Why this looks like a server-side issue (not client)

1. The byte-identical response across anonymous / invalid-token / valid-token requests rules out a per-key quota or auth problem.
2. The official `clawhub@0.9.0` CLI `explore` command hits the same endpoint and also returns empty — so it's not a client bug.
3. `/api/v1/search` works on the same token, so network and auth are fine.
4. `ratelimit-remaining` is far from zero, so this isn't 429.

## Related

[openclaw/openclaw#52949](https://github.com/openclaw/openclaw/issues/52949) (closed) mentioned the same `items:[]` symptom as a separate sub-issue from its 429 part. Re-filing here because the list-endpoint behavior still reproduces independently of that issue.

## Ask

- Is `/api/v1/skills` intended to browse the full public catalog (as `clawhub explore` implies), or is it scoped to the authenticated user's own published skills?
- If the former, what changed server-side? If the latter, the `explore` command and the OpenAPI spec's lack of scope annotation are misleading.

Linux2010:
Fixes #1722

## Problem
The `/api/v1/skills` endpoint returns empty `{"items":[],"nextCursor":null}` for all callers because it calls `listPublicPage` which is a deprecated stub.

## Root Cause
In `convex/httpApiV1/skillsV1.ts:474`, the handler calls `api.skills.listPublicPage` which always returns empty.

## Solution
Changed to call `listPublicPageV4` which has the actual implementation with proper filtering and pagination.

## Changes
1. Call `listPublicPageV4` instead of `listPublicPage`
2. Map V1 sort params to V4 format
3. Map V4 return format (page -> items)

## Impact
- `clawhub explore` CLI command will work again
- Public skills listing via API will return actual results

### DIFF
--- a/convex/httpApiV1/skillsV1.ts
+++ b/convex/httpApiV1/skillsV1.ts
@@ -470,13 +470,19 @@ export async function listSkillsV1Handler(ctx: ActionCtx, request: Request) {
     url.searchParams.get("nonSuspiciousOnly"),
     url.searchParams.get("nonSuspicious"),
   );
+  // Map V1 sort to V4 sort (V4 uses "newest" instead of "updated")
+  const v4Sort = sort === "updated" ? "newest" :
+                 sort === "installsCurrent" ? "installs" :
+                 sort === "installsAllTime" ? "installs" : sort;
 
-  const result = (await ctx.runQuery(api.skills.listPublicPage, {
-    limit,
+  const v4Result = await ctx.runQuery(api.skills.listPublicPageV4, {
     cursor,
-    sort,
+    numItems: limit,
+    sort: v4Sort,
     nonSuspiciousOnly: nonSuspiciousOnly || undefined,
-  })) as ListSkillsResult;
+  });
+
+  const result = { items: v4Result.page ?? [], nextCursor: v4Result.nextCursor ?? null } as ListSkillsResult;
 
   // Batch resolve all tags in a single query instead of N queries
   const resolvedTagsList = await resolveTagsBatch(


steipete:
Thanks for tracking this down. Fixed on `main` in `dd111ca`.

I used the same core direction, with a couple of corrections before landing:
- `/api/v1/skills` now calls `listPublicPageV4` for cursor-backed sorts.
- `updated` keeps updated-at ordering instead of being mapped to creation date.
- `trending` routes through a leaderboard-backed query instead of passing an unsupported sort into V4.

Verified with `bun run test convex/httpApiV1.handlers.test.ts` and `bun run lint`. Closing this PR as superseded by the landed main fix.

Issue #1759, opened 2026-04-21
Bug: search loses vector scores and can completely miss highly relevant skills due to three interacting defects


LinPower:
## Summary

The `searchSkills` action in `convex/search.ts` has three interacting bugs that together can cause highly relevant skills to be **completely absent** from search results — not just poorly ranked. The bugs form a chain: overly permissive filtering fills the result quota with partially-matching skills, which triggers early loop termination and skips the lexical fallback, making the target skill unreachable by any path.

## Reproduction

1. Search for a multi-word query like `"baidu yijian"` on clawhub.ai/skills
2. The skill "Baidu Yijian Vision" does **not** appear in results at all
3. Previously, this skill appeared as #1 when loading enough results (limit ~75), but as the skill registry grew, it disappeared entirely
4. The ranking and visibility of skills changes depending on the `limit` parameter passed to the search API — a frontend pagination concern that should not affect search correctness

### Bug 1: `matchesExactTokens` uses `some` instead of `every`

```typescript
// convex/lib/searchText.ts, line 152-154
return queryTokens.some((queryToken) =>
  textTokens.some((textToken) => textToken.startsWith(queryToken)),
);
```

For query `"baidu yijian"` (tokens: `["baidu", "yijian"]`), this passes any skill whose displayName/slug/summary matches **at least one** of the tokens. Skills matching only "baidu" or only "yijian" are treated as equally valid as skills matching both.

This is too permissive as a filter. It causes `exactMatches` to fill up with partially-matching results, which triggers the early exit condition in Bug 2.

### Bug 2: Early loop exit + skipped lexical fallback (consequence of Bug 1)

```typescript
// convex/search.ts, inside the while loop
if (exactMatches.length >= limit || results.length < candidateLimit) {
  break;  // ← exits the candidate expansion loop early
}

// After the loop:
const fallbackMatches =
  primaryMatches.length >= limit
    ? []   // ← lexical fallback is completely skipped!
    : await ctx.runQuery(internal.search.lexicalFallbackSkills, {...});
```

Because Bug 1 makes `matchesExactTokens` too permissive, `exactMatches` quickly reaches `limit` (e.g., 25) with partially-matching skills. This triggers two problems:

1. **Early loop exit**: The candidate expansion loop (50 → 100 → 200 → 256) stops before the target skill enters the vector candidate pool
[...]

PR #1762, opened 2026-04-21
fix(search): preserve vector scores across loop iterations and tighte…

LinPower:
## Problem

The `searchSkills` and `searchSouls` actions have a bug that causes vector scores to be lost across candidate expansion iterations, leading to inconsistent and incorrect search rankings.

### Bug: `scoreById` overwritten each loop iteration

```typescript
// convex/search.ts (inside while loop)
scoreById = new Map<Id<"skillEmbeddings">, number>(
  results.map((result) => [result._id, result._score]),
);
```

The entire map is **replaced** on each iteration rather than merged. Meanwhile, `hydrated` entries are correctly accumulated across iterations:

```typescript
hydrated = [...hydrated, ...newEntries]; // accumulated ✅
```

Skills found in earlier iterations lose their vector scores (treated as 0 via `scoreById.get(entry.embeddingId) ?? 0`), causing severe ranking degradation. This also makes search results inconsistent — the same query produces different rankings depending on the `limit` parameter.

**Why ranking changes with `limit`:** When the frontend requests a larger `limit` (e.g., 75), the initial `candidateLimit` starts higher (225), so fewer loop iterations are needed and `scoreById` is overwritten fewer times. In the extreme case where only one iteration is needed, no scores are lost.

### `convex/search.ts`
- Changed `scoreById = new Map(...)` → `for (result) scoreById.set(...)` in both `searchSkills` and `searchSouls`

### `convex/search.test.ts`
- Added: `preserves vector scores across candidate expansion iterations`

## Test

- Vector scores preserved across loop iterations — early-found skills retain their scores

### DIFF
--- a/convex/search.test.ts
+++ b/convex/search.test.ts
@@ -866,6 +866,83 @@ describe("search helpers", () => {
     expect(merged).toHaveLength(2);
     expect(merged.map((entry) => entry.skill._id)).toEqual(["skills:1", "skills:2"]);
   });
+
+  it("preserves vector scores across candidate expansion iterations", async () => {
+    // Regression test for scoreById overwrite bug.
+    //
+    // Setup:
+    //   limit=10  →  candidateLimit starts at 50, maxCandidate=200
+    //   Iteration 1: vectorSearch returns exactly 50 results (= candidateLimit)
+    //                → results.length < candidateLimit is false → loop continues
+    //   Iteration 2: vectorSearch returns 2 results (< 100) → loop exits
+    //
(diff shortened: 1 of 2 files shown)


steipete:
Landed on `main` in `e3c772d`.

I kept the vector-score merge from this PR and also fixed the related exact-token filter from #1759 so partial one-token matches cannot fill the quota and crowd out better multi-token results.

Validation:

- `bun run test convex/search.test.ts convex/lib/searchText.test.ts`
- `bun run lint`
- `git diff --check`

Closing this PR as superseded by the main commit.

PR #1763, opened 2026-04-21
fix(ui): trim version input before semver validation

### LINKED ISSUE #1739: Publish skill page: version field rejects valid minor/major bumps with "must be valid semver" error
## Summary

The new-version form on the publish-skill page auto-populates the version field with a patch bump (e.g. current published version `1.5.1` → field pre-fills `1.5.2`). Changing that value to a minor or major bump (e.g. `1.6.0`) — by typing or pasting — fires a client-side validation error:

> Version must be valid semver (e.g. 1.0.0).

`1.6.0` **is** valid semver. The auto-populated `1.5.2` is accepted. Only the user-edited value is rejected.

## Reproduction

1. Go to the publish-new-version page for any skill that already has a published version (in my case `@axonflow/governance-policies` at 1.5.1).
2. The Version field auto-fills with the next patch (e.g. `1.5.2`).
3. Change it to `1.6.0` — either by typing or pasting.
4. The form shows: **Version must be valid semver (e.g. 1.0.0).**
5. Revert to `1.5.2` — accepted.

## Expected

Any valid semver string >= the current published version should be accepted. `1.6.0` after a `1.5.1` publish is a clean minor bump and should pass.

## Actual

Only the auto-populated patch bump is accepted. Any manual edit — including trivially-valid semver like `1.6.0`, `2.0.0`, or even the same value retyped — trips the validator.

## Likely root cause

Suspicion: the validator isn't actually checking semver format — it's checking equality with the auto-populated suggestion, then surfacing the generic "must be valid semver" message on mismatch. Or it's doing a strict-successor check and rejecting anything that isn't exactly `current + patch`.

If the intent is to enforce "must be a patch-only bump from the latest published version," the error message is misleading — should be something like "Version must be a patch-level increment from 1.5.1 (e.g. 1.5.2)." If the intent is just semver validation, the regex needs to actually accept all valid semver strings.

## Impact

- Skill maintainers cannot cut minor or major releases from the web UI.
- Forces either:
  - Workaround: release everything as a patch bump (breaks semver meaning for consumers reading the skill CHANGELOG).
  - Workaround: publish via CLI instead, bypassing the form entirely.
- The `(e.g. 1.0.0)` hint in the error tells users the format is the problem, so they'll try things like `1.6.0`, `v1.6.0`, `1.6`, etc. — none work, because the actual rule is different from what the message says.

## Workaround I'm using

Releasing the v1.6.0 skill content as `1.5.2` instead. The CHANGELOG entry in-skill will reference the real version number (1.6.0) but the ClawHub listing will be tagged 1.5.2 until this lands.

## Environment

- ClawHub publish page: https://clawhub.ai/ (current deployment as of 2026-04-19)
- Skill in question: `@axonflow/governance-policies`
- Current published: 1.5.1
- Attempted: 1.6.0 (rejected)
- Accepted: 1.5.2 (auto-populated default)

Linux2010:
Fixes #1739

## Problem
The version field on the publish-skill page rejects valid minor/major bumps (e.g. `1.6.0` after `1.5.1`) because user input may contain leading/trailing whitespace.

`semver.valid()` returns `null` for strings with whitespace, triggering the "Version must be valid semver (e.g. 1.0.0)" error even though the trimmed value would be valid.

## Root Cause
- `version` state variable was not trimmed before validation
- Other fields (`slug`, `displayName`, `changelog`) are trimmed, but `version` was missed

## Solution
Added `trimmedVersion = version.trim()` and used it in:
- `semver.valid()` checks (line 219, 288)
- `publishVersion` call (line 435)
- Toast success message (line 446)
- Changelog preview generation (line 241)
- All useMemo dependencies

The Input value remains untrimmed for user visibility.

## Testing
Users can now manually enter:
- `1.6.0` (no whitespace) ✓
- `  1.6.0` (leading space) ✓
- `1.6.0  ` (trailing space) ✓

All will be trimmed and accepted as valid semver.

## Impact
- Skill maintainers can now publish minor/major releases from the web UI
- No need to workaround by publishing via CLI or using patch-only bumps

### DIFF
--- a/src/routes/publish-skill.tsx
+++ b/src/routes/publish-skill.tsx
@@ -167,6 +167,7 @@ export function Upload() {
   const trimmedSlug = slug.trim();
   const trimmedName = displayName.trim();
   const trimmedChangelog = changelog.trim();
+  const trimmedVersion = version.trim();
   const slugAvailability = useQuery(
     api.skills.checkSlugAvailability,
     !isSoulMode && isAuthenticated && trimmedSlug && SLUG_PATTERN.test(trimmedSlug)
@@ -215,7 +216,7 @@ export function Upload() {
     if (changelogTouchedRef.current) return;
     if (trimmedChangelog) return;
     if (!trimmedSlug || !SLUG_PATTERN.test(trimmedSlug)) return;
-    if (!semver.valid(version)) return;
+    if (!semver.valid(trimmedVersion)) return;
     if (!hasRequiredFile) return;
     if (files.length === 0) return;
 
@@ -228,7 +229,7 @@ export function Upload() {
     const requiredFile = files[requiredIndex];
     if (!requiredFile) return;
 
-    const key = `${trimmedSlug}:${version}:${requiredFile.size}:${requiredFile.lastModified}:${normalizedPaths.length}`;
+    const key = `${trimmedSlug}:${trimmedVersion}:${requiredFile.size}:${requiredFile.lastModified}:${normalizedPaths.length}`;
     if (changelogKeyRef.current === key) return;
     changelogKeyRef.current = key;
 
@@ -240,7 +241,7 @@ export function Upload() {
         if (changelogRequestRef.current !== requestId) return null;
         return generateChangelogPreview({
           slug: trimmedSlug,
-          version,
+          trimmedVersion,
           readmeText: text.slice(0, 20_000),
           filePaths: normalizedPaths,
         });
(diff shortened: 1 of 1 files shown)


steipete:
Thanks for the fix. This is now fixed on `main` in `dd111ca`.

I landed the intended trimming behavior while preserving the action payload shape: the UI validates and publishes with `version: trimmedVersion`, and the changelog preview still receives the expected `version` field.

Verified with `bun run test convex/httpApiV1.handlers.test.ts` and `bun run lint`. Closing this PR as superseded by the landed main fix.

PR #1803, opened 2026-04-23
fix: reduce env scan false positives

### LINKED ISSUE #1790: False positives when environment variables suggested for security purposes
Skills using environment variables as a security mechanism to avoid plain storage of API keys, are wrongly flagged as risky due to environment variables sent through network.
Example: 
https://clawhub.ai/salomonhenao/veritier-fact-checking

deepujain:
## Summary

Fixes #1790.

- stop flagging `suspicious.env_credential_access` when code sends only declared env vars to the expected API
- keep flagging undeclared env usage and broad `process.env` scraping
- keep existing exfiltration coverage for file read plus network send
- add focused regression tests for the safe and unsafe cases
- bump the moderation engine version for the scanner behavior change

## Root cause

The static moderation scan treated any `process.env` access plus any network call as suspicious. That caught legitimate skills which declare an API key in metadata and then send that key to the provider they integrate with.

## Follow-up

The full-suite validation initially found one stale expectation in `convex/skills.pendingScanQueue.test.ts` after the engine-version bump. Commit `5fdedaf` updates that expectation so current-version scans are not requeued as stale.

## Validation

- `git diff --check`

### DIFF
--- a/convex/lib/moderationEngine.test.ts
+++ b/convex/lib/moderationEngine.test.ts
@@ -38,13 +38,18 @@ describe("moderationEngine", () => {
     expect(result.status).toBe("suspicious");
   });
 
-  it("flags process.env + fetch as suspicious (not malicious)", () => {
+  it("does not flag declared env vars sent to the intended API", () => {
     const result = runStaticModerationScan({
       slug: "todoist",
       displayName: "Todoist",
       summary: "Manage tasks via the Todoist API",
       frontmatter: {},
-      metadata: {},
+      metadata: {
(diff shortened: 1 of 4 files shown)


deepujain:
Ran the same clean-worktree sweep and found one failing full-suite expectation in convex/skills.pendingScanQueue.test.ts after the moderation engine version bump. Pushed 5fdedaf to align that stale/current expectation; oxlint, full vitest, repo/package typechecks, vite build, and asset copy are now green locally.

PR #1804, opened 2026-04-23
fix: separate historical tags on skill detail page

deepujain:
## Summary

- show only tags that point at the latest skill version in the public detail-page tag surfaces
- move older tag mappings into a separate manager-only "Historical tags" section so they are no longer mixed with current metadata
- add focused detail-page tests covering both the public filtering behavior and the manager-only historical tag view

## Root Cause

The skill detail page rendered every entry from `skill.tags`, even though that map can intentionally retain tag pointers for older versions. After metadata refreshes, the page mixed current tags with historical tags because it did not filter by `latestVersionId`.

## User Impact

Publishers and visitors now see the current/latest tag set on the detail page instead of a noisy blend of old and new metadata. Managers still keep visibility into historical tag mappings and can delete them from the same page.

## Validation

- `git diff --check`

### DIFF
--- a/PR_1784_body.md
+++ b/PR_1784_body.md
@@ -1,0 +1,17 @@
+## Summary
+
+- show only tags that point at the latest skill version in the public detail-page tag surfaces
+- move older tag mappings into a separate manager-only "Historical tags" section so they are no longer mixed with current metadata
+- add focused detail-page tests covering both the public filtering behavior and the manager-only historical tag view
+
+## Root Cause
+
+The skill detail page rendered every entry from `skill.tags`, even though that map can intentionally retain tag pointers for older versions. After metadata refreshes, the page mixed current tags with historical tags because it did not filter by `latestVersionId`.
+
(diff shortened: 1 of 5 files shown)


deepujain:
Ran a clean-worktree local validation sweep before marking this ready: oxlint, vitest run, repo tsc, packages/schema tsc, packages/clawhub tsc, vite build, and the OG asset copy script. The current branch stayed green locally.

steipete:
Landed the clean version on `main` in `d5776f8`.

I kept the implementation and regression coverage, but excluded the accidental root `PR_1784_body.md` artifact from the PR patch. Validation run locally:

- `bun run lint`
- `bun run build`

The focused React test file is still blocked locally by the existing duplicate-React invalid-hook-call failure that affects all tests in `src/__tests__/skill-detail-page.test.tsx`, so I could not use that file as a passing signal. Closing this PR as superseded by the main commit.

PR #1805, opened 2026-04-23
fix: keep skill moderation tied to latest version

### LINKED ISSUE #1795: early versions with issues cause all subsequent versions to show failed even when they show clean.
https://clawhub.ai/encryptshawn/openclaw-emergency-rollback

My skill is clean but earlier releases wernt and clawhub has permenantly marked it bad, even showing suspciious for virus too but clicking in shows its clean

This is a critical skill for any remote openclaw admin, but also it has scripts, its important to be trusted, clawhub makes it hard to see that it is clean with these old artifacts poisoning the new.  I probably will jsut create a new one but I dont think I can delete the old one so it will also cause further confusion for users either way.

deepujain:
## Summary

Fixes #1795.

- stop hash-based VT callbacks from overwriting skill-level moderation when they target an older version
- keep skill-level moderation sourced from the current `latestVersionId`
- add a paginated moderation resync path for stale rows whose `moderationSourceVersionId` still points at an older version
- add focused regression tests for both the prevention path and the repair path

## Root cause

`approveSkillByHashInternal` and `escalateByVtInternal` updated the parent `skills` row for any matching version hash, even when that hash belonged to an older release. If an older suspicious release was rescanned after a clean newer release existed, the skill-level badge could be re-poisoned from the stale version.

## Follow-up

The broader validation pass found that older test fixtures without `latestVersionId` were being skipped too aggressively. Commit `aead53f` preserves the latest-version guard while allowing legacy fixtures/documents that do not yet have `latestVersionId`.

## Repair path

After deploy, admins can run:

- `npx convex run maintenance:backfillLatestSkillModeration --prod`

That backfill only touches scanner-managed rows whose `moderationSourceVersionId` no longer matches `latestVersionId`, so it repairs stale skill badges without broad moderation churn.

## Validation

- `git diff --check`

### DIFF
--- a/convex/maintenance.test.ts
+++ b/convex/maintenance.test.ts
@@ -23,6 +23,7 @@ vi.mock("./_generated/api", () => ({
       nominateEmptySkillSpammersInternal: Symbol("nominateEmptySkillSpammersInternal"),
     },
     skills: {
+      backfillLatestSkillModerationInternal: Symbol("skills.backfillLatestSkillModerationInternal"),
       getVersionByIdInternal: Symbol("skills.getVersionByIdInternal"),
       getOwnerSkillActivityInternal: Symbol("skills.getOwnerSkillActivityInternal"),
     },
--- a/convex/maintenance.ts
+++ b/convex/maintenance.ts
@@ -1965,6 +1965,20 @@ export const backfillLatestVersionSummaryInternal = internalMutation({
   },
 });
 
+// Repair stale skill-level moderation that was sourced from a non-latest version.
+// Run once after deploying the latest-version moderation fix:
+//   npx convex run maintenance:backfillLatestSkillModeration --prod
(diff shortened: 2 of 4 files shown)


deepujain:
Re-ran the clean-worktree sweep on current head aead53f: oxlint, full vitest, repo/package typechecks, vite build, and asset copy all pass locally.

steipete:
Landed a tightened version on `main` in `e7a1e99`.

Kept the latest-version guards for hash-based VT approval/escalation and the repair action, with one extra safety change: the backfill now only re-syncs scanner-managed rows and skips manual/non-scanner moderation reasons so it does not churn unrelated active skills.

Validation run locally:

- `bun run test convex/skills.rateLimit.test.ts convex/maintenance.test.ts`
- `bun run lint`
- `git diff --check`

Closing this PR as superseded by the main commit.

PR #1806, opened 2026-04-23
feat(cli): add per-skill pinning

### LINKED ISSUE #1768: Feature request: native per-skill pinning to block update/--force/--all
## Problem

`clawhub update`, `clawhub update --force`, and `clawhub update --all` give operators no way to mark a specific installed skill as "do not touch" while still keeping the rest of the lockfile fresh. This matters most during an active scanner-flag window: when ClawHub flags a skill SUSPICIOUS but disposition is "wait for moderator", the locally-installed bytes are the last known-good copy. A routine `update --force` (from a cron, a security-check script, or another shell session) will replace those bytes with whatever the registry currently serves — exactly the wrong direction.

Concrete trigger: on 2026-04-21 the scanner flagged four skills (`bear-notes`, `slack`, `trello`, `nano-banana-pro`) as SUSPICIOUS pending moderator review. To prevent any unattended `update --force` from overwriting them, we built a wrapper that blocks the relevant subcommands on a per-slug pin file. It works, but every consumer (humans, crons, security-check skill, sibling Claude sessions) has to remember to call the wrapper instead of `clawhub`. A native flag would close that gap.

## Proposed surface

Subcommands:

- `clawhub pin <slug> [--reason <text>]` — mark a slug as pinned in the lockfile.
- `clawhub unpin <slug>` — remove the pin.
- `clawhub list` — show a `pinned` indicator next to pinned slugs.

Lockfile semantics (option A, single file):

```json
{
  "version": 2,
  "skills": {
    "bear-notes": {
      "version": "1.0.0",
      "installedAt": 1773698478099,
      "pinned": true,
      "pinReason": "scanner-flagged 2026-04-21, awaiting moderator"
    }
  }
}
```

Or option B, sidecar `pinned.json` next to `lock.json`, which is what our wrapper uses today.

Update behavior:

- `clawhub update <slug>` where `<slug>` is pinned → exit non-zero with a clear message ("skill `<slug>` is pinned; run `clawhub unpin <slug>` first"). Do not silently no-op — operators asking for a specific slug should hear back.
- `clawhub update --all` → silently skip pinned slugs and report them in the summary ("skipped 4 pinned skills: …"). Continue updating the rest. Do not fail the run just because pinned slugs exist.
- `clawhub update --force` (with or without a slug) → must still honor the pin. `--force` is about ignoring local-changes detection, not about ignoring an explicit operator decision to freeze a slug.
- `clawhub update --force --all` → same as `--all`: skip pinned, update the rest.

The only way to update a pinned skill is to unpin it first. That keeps the audit trail in lockfile diffs (or pin-file diffs).

## Prior art / current workaround

We ship a thin wrapper at `~/.openclaw/bin/clawhub-safe` that reads `~/.openclaw/workspace/.clawhub/pinned.json` and refuses `update`/`update --force`/`update --all` for any listed slug; other subcommands pass through to `clawhub` unchanged. It also fail-closes `--all`/`--force` if the pin file is missing, so a deleted pin file can't silently re-enable mass updates. That's adequate for one operator who controls every shell, but doesn't generalize — once you have crons, scheduled jobs, and multiple agents all touching the same install, "remember to call the wrapper" stops being a reliable safeguard. Native CLI support would.

Happy to provide more detail on the wrapper's exit-code conventions or pin-file shape if that's useful for design discussion.

## Why this is worth it

The supply-chain pattern this defends against is already happening (steipete-cluster flag on 2026-04-21, ClawHavoc earlier in April). Pin-then-investigate is the natural operator response, and it deserves a first-class CLI affordance rather than a bolt-on wrapper at every install.

deepujain:
## Summary

- add native `clawhub pin` and `clawhub unpin` commands backed by lockfile metadata
- block direct updates for pinned skills, skip them during `update --all`, and surface the pinned state in `clawhub list`
- extend the CLI/schema tests and docs for the new pinning workflow

## Problem

Operators had no native way to freeze a single installed skill during a scanner-review window. `update`, `update --force`, and `update --all` could still replace the local last-known-good bytes unless every caller remembered to go through an external wrapper.

## Solution

Store optional `pinned` and `pinReason` metadata in the CLI lockfile and teach the CLI to honor it everywhere that matters:
- `pin`/`unpin` update the lockfile for installed skills
- `update <slug>` fails fast with an explicit unpin instruction
- `update --all` skips pinned slugs and reports them in the summary
- `list` shows which installed skills are pinned and why

## Follow-up

This branch was rebuilt on current `origin/main` and force-pushed at `739a7b2` so the PR contains only the skill-pinning work. Earlier unrelated moderation changes are no longer part of this diff.

## Validation

- `git diff --check`
- `npm run --workspace packages/clawhub verify:build`

Closes #1768

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Changes
 
+- CLI: add per-skill pinning so installed skills can be frozen against direct updates, bulk updates, and force reinstalls (#1806) (thanks @deepujain).
 - Web: polish browse/listing surfaces across skills, plugins, and search, including plugin card view parity, clearer search controls, visible safety filtering, and more consistent card metadata treatment (#2084) (thanks @vyctorbrzezowski).
 - Web: rename the skills and plugins browse alternate view from Cards to Grid while keeping legacy `view=cards` URLs compatible (#2090) (thanks @vyctorbrzezowski).
 - Web: allow skill owners and publisher admins to edit a skill summary from the detail page (#1411) (thanks @SylvanXiao).
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -139,6 +139,17 @@ Stores your API token + cached registry URL.
 ### `list`
 
 - Reads `<workdir>/.clawhub/lock.json` (legacy `.clawdhub`).
+- Shows `pinned` next to skills frozen with `clawhub pin`, including the optional reason.
+
+### `pin <slug>`
+
+- Marks an installed skill as pinned in the lockfile.
+- `--reason <text>` records why the skill is frozen.
+- Pinned skills are skipped by `update --all` and rejected by direct `update <slug>`.
+
+### `unpin <slug>`
+
+- Removes the lockfile pin from an installed skill so future updates can modify it.
 
 ### `update [slug]` / `update --all`
 
@@ -147,6 +158,9 @@ Stores your API token + cached registry URL.
 - If fingerprint does not match:
   - refuses by default
   - overwrites with `--force` (or prompt, if interactive)
+- Pinned skills are never updated by `--force`.
+- `update <slug>` fails fast for pinned slugs and tells you to run `clawhub unpin <slug>` first.
+- `update --all` skips pinned slugs and prints a summary of what stayed frozen.
 
 ### `skill publish <path>`
 
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -126,6 +126,7 @@ Stores your API token + cached registry URL.
 - Resolves latest version via `/api/v1/skills/<slug>`.
 - Downloads zip via `/api/v1/download`.
 - Extracts into `<workdir>/<dir>/<slug>`.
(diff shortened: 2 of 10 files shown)


deepujain:
Rebuilt this branch on current origin/main to remove unrelated moderation changes that had leaked in, then force-pushed 739a7b2. Clean-worktree sweep is green: oxlint, full vitest, repo/package typechecks, vite build, asset copy, and npm run --workspace packages/clawhub verify:build.

deepujain:
Pushed a UX follow-up for the Greptile P2s: reinstalling an already-pinned skill now reports that the pin state is preserved, and repeating `clawhub pin` without changing the reason reports that the skill is already pinned instead of sounding like a fresh mutation. Added focused tests for both paths.\n\nLocal validation run:\n- `../../node_modules/.bin/vitest run -c vitest.config.ts src/cli/commands/skills.test.ts` from `packages/clawhub`\n- `./node_modules/.bin/tsc -p packages/clawhub/tsconfig.json --noEmit`\n- `git diff --check`

PR #1810, opened 2026-04-24
fix(skill-install): stabilize install surface layout and eliminate visual jumps

momothemage:
Rework the SkillInstallSurface layout so it no longer shifts or jumps when
users interact with the Copy Prompt dropdown or toggle between prompt modes.

Layout
- Make the two install panels share the grid row equally (1fr/1fr instead
  of 1.08fr/0.92fr) and raise the stacking breakpoint from 860px to 1024px
  so narrow viewports collapse to a single column sooner.
- Drop the two-column grid on `.skill-hero-top.has-plugin` (and its 960px
  override) so Prompt Flow / Command Line cards flow full-width beneath the
  skill title instead of sitting in a cramped right rail with a large empty
  gutter on the left.

Interaction stability
- Neutralize react-remove-scroll-bar's `padding-right` compensation on
  `body[data-scroll-locked]` — we already reserve the gutter globally via
  `scrollbar-gutter: stable`, and the double-compensation caused the page
  to jump ~15px horizontally every time a Radix dropdown/dialog opened.
  The override uses `html body[...]` to outrank Radix's runtime-injected
  `!important` rule on specificity.
- Pin `.skill-install-copy-feedback` to its own row via `flex-basis: 100%`
  so long status strings ("Install & Setup prompt copied.") no longer wrap
  under the Copy Prompt button and shove the rest of the panel down.
- Give `.skill-install-prompt-preview` a `min-height: 22em` so switching
  between Install Only and Install & Setup keeps the preview box — and
  therefore the whole install grid row — at a stable height.

Before:
image

After:
Clipboard_Screenshot_1777020261

### DIFF
--- a/src/styles.css
+++ b/src/styles.css
@@ -600,6 +600,26 @@ html {
   scrollbar-gutter: stable;
 }
 
+/*
+ * Radix UI (react-remove-scroll) locks body scroll when a dropdown/dialog/popover
+ * opens and, to prevent the scrollbar from disappearing visibly, injects a
+ * dynamic stylesheet that sets `padding-right: <scrollbar-width>px !important`
+ * on `body[data-scroll-locked]`. We already reserve a stable scrollbar gutter
+ * globally via `scrollbar-gutter: stable` on <html>, so Radix's compensation is
+ * double-work and ends up shifting the whole page ~15px to the left for the
+ * duration the overlay is open — producing a visible horizontal "jump".
+ *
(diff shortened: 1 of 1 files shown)


BunsDev (approved):
Verified the CSS-only layout stabilization locally.

Checks run:
- bunx tsc -p packages/schema/tsconfig.json --noEmit
- bunx tsc -p packages/clawhub/tsconfig.json --noEmit
- bun run build

The remaining Vercel status is the fork preview authorization gate, not a code failure.

Issue #1830, opened 2026-04-27
False positive: skill-factory incorrectly flagged as suspicious


zhelunSun:
## Skill Details
- ClawHub page: https://clawhub.ai/zhelunsun/skillfactory
- Version: v2.2.2
- GitHub repo: https://github.com/zhelunSun/skill-factory

## Issue Description
My skill has been flagged as "Skill flagged — suspicious patterns detected" by ClawHub Security.
I believe this is a false positive.

## Why this is a false positive
This skill is **instruction-only** — it contains only:
- `SKILL.md`: A plain-text SOP workflow document
- `references/*.md`: Knowledge base markdown files
- `README.md` / `README_zh.md`: Documentation

There are **no scripts, no network calls, no data exfiltration, no executable code** of any kind.

## Evidence
The skill's entire content is static markdown instructions for building other skills (a "meta-skill").
No VirusTotal security vendors flagged the content as malicious — the "Suspicious" label appears 
to come solely from VirusTotal's AI Code Insight behavioral heuristics, which are known to produce 
false positives on instruction-heavy markdown files.

## Request
Please review and remove the "Suspicious" flag, or apply a "Benign" manual override.
Thank you.

steipete:
Fixed and live.

What landed:
- Suppress uncorroborated VirusTotal Code Insight-only suspicious verdicts when static scan + LLM + AV engine stats are clean.
- Clear legacy `scanner.*.suspicious` moderation state so old rows do not remain `isSuspicious` after the aggregate verdict is clean.
- Normalize VT engine stats before caching so production resync handles VT's extra response keys.

Verified:
- CI `25040605404` green.
- Backend deploy `25041700134` green.
- `https://clawhub.ai/api/v1/skills/skillfactory` now returns `moderation: null`.
- `https://clawhub.ai/zhelunsun/skillfactory` returns 200.

Closing.

PR #1841, opened 2026-04-27
fix: calibrate VT Code Insight moderation

deepujain:
## What problem was happening

VirusTotal Code Insight can return a `suspicious` verdict even when the local ClawHub static scanner, the LLM moderation result, and VirusTotal AV engine stats are all clean. That makes benign repositories, such as the skill-factory false positive in #1830, stay quarantined without corroborating scanner evidence.

## Why this is the right fix

The moderation aggregate should still treat malicious VT verdicts and AV-engine suspicious findings as strong signals. The false-positive seam is narrower: a lone Code Insight `suspicious` verdict should not override otherwise clean local and AV scanner evidence.

## What changed

- Stores VT scanner provenance and engine stats in `vtAnalysis` for skills and packages.
- Suppresses the `suspicious.vt_suspicious` moderation reason only when the VT scanner is `code_insight`, static scan is clean, LLM moderation is clean/benign, and VT AV engine malicious/suspicious counts are zero.
- Keeps VT suspicious moderation active when AV engines also report suspicious results.
- Clears legacy quarantine flags when a newly computed aggregate snapshot is clean, while preserving admin/moderator bypass reasons.
- Adds regression coverage for both the uncorroborated Code Insight false-positive case and the corroborated suspicious case.

## What did not change

- Malicious VT verdicts still quarantine.
- Suspicious VT verdicts from AV-engine evidence still quarantine.
- Existing already-flagged rows may need a rescan/backfill after deploy so stored `vtAnalysis.engineStats` is available for aggregate recalculation.
- Merging to `main` does not deploy production; production deploys remain manual.

## Validation

- `./node_modules/.bin/vitest run convex/skills.rateLimit.test.ts convex/lib/moderationEngine.test.ts convex/vt.test.ts convex/packages.public.test.ts`
- `git diff --check`

### DIFF
--- a/convex/lib/moderationEngine.test.ts
+++ b/convex/lib/moderationEngine.test.ts
@@ -527,4 +527,60 @@ describe("moderationEngine", () => {
     expect(snapshot.reasonCodes).toContain("suspicious.env_credential_access");
     expect(snapshot.reasonCodes).toContain("suspicious.vt_suspicious");
   });
+
+  it("does not let uncorroborated VT Code Insight suspicious override clean local scans", () => {
+    const snapshot = buildModerationSnapshot({
+      staticScan: {
+        status: "clean",
+        reasonCodes: [],
+        findings: [],
+        summary: "",
+        engineVersion: "v2.1.1",
+        checkedAt: Date.now(),
+      },
+      vtAnalysis: {
+        status: "suspicious",
+        scanner: "code_insight",
+        source: "VirusTotal Code Insight",
+        engineStats: {
+          malicious: 0,
+          suspicious: 0,
+          harmless: 12,
+          undetected: 54,
+        },
+      },
+      llmStatus: "clean",
+    });
+
+    expect(snapshot.verdict).toBe("clean");
+    expect(snapshot.reasonCodes).toEqual([]);
+  });
+
+  it("keeps VT Code Insight suspicious when AV engines also report suspicious", () => {
+    const snapshot = buildModerationSnapshot({
+      staticScan: {
+        status: "clean",
+        reasonCodes: [],
+        findings: [],
+        summary: "",
+        engineVersion: "v2.1.1",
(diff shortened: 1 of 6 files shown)


deepujain:
Pushed a small follow-up for the Greptile P2 on static scan gating: documented why older records with no persisted static scan are treated as having no static findings available to corroborate VT Code Insight. I intentionally left the validator consolidation as a separate refactor-sized cleanup because the current PR is already green and the shared-validator extraction would touch schema + mutation surfaces.\n\nLocal validation run:\n- `./node_modules/.bin/vitest run convex/lib/moderationEngine.test.ts`\n- `./node_modules/.bin/tsc --noEmit`\n- `git diff --check`

steipete:
Landed this manually on `main` because I needed to add a follow-up fix in this repo and the PR branch is from a fork.

Landed commits:
- `d855d09` / `932155c` from this PR, rebased onto current `main`.
- `232e429` to clear the legacy `scanner.*.suspicious` path so clean aggregate verdicts also clear `isSuspicious`.
- `b8ba595` changelog entry with thanks.
- `3701733` to normalize VT `last_analysis_stats` before caching; live prod rescan exposed extra VT keys (`failure`, `type-unsupported`, etc.) that Convex rejects.

Verified:
- Focused moderation/VT tests passed.
- Full `bun run test` passed: 135 files / 1072 tests.
- `bun run build` passed.
- `bunx tsc --noEmit` passed.
- Package typechecks passed for `packages/schema` and `packages/clawhub`.
- CI `25040605404` green.
- Security gate `25040605376` green.
- Backend deploy `25041700134` green.
- Live API now returns `moderation: null` for `https://clawhub.ai/api/v1/skills/skillfactory`; page `https://clawhub.ai/zhelunsun/skillfactory` returns 200.

Closing as landed. Thanks @deepujain.

PR #1842, opened 2026-04-27
fix: constrain plugin catalog queries

deepujain:
## What problem was happening

The public plugin catalog endpoints were not constraining default queries to plugin packages. `/api/v1/plugins` and `/api/v1/plugins/search` could fall through to the generic package query shape when no explicit `family` filter was present, which made the plugin catalog route depend on mixed package/skill catalog data instead of the plugin families it serves.

## Why this is the right fix

The `/plugins` API should default to the two plugin package families: `code-plugin` and `bundle-plugin`. Explicit family filters still win, but the default plugin route now asks the backend for plugin families only and the backend enforces that family set during listing/search scans.

## What changed

- Added an optional `families` filter to public package list/search query internals.
- Applies `families: ["code-plugin", "bundle-plugin"]` to `/api/v1/plugins` and `/api/v1/plugins/search`.
- Added handler regressions for the plugin list and search default family filters.
- Added a backend regression that proves skill rows are excluded when a plugin family set is requested.

## What did not change

- `/api/v1/packages` continues to support the combined package/skill catalog behavior.
- Explicit `family=code-plugin` and `family=bundle-plugin` routes still use their single-family paths.
- Merging to `main` does not deploy production; production deploy remains a manual workflow.

## Validation

- `./node_modules/.bin/vitest run convex/packages.public.test.ts convex/httpApiV1.handlers.test.ts`
- `git diff --check`


steipete (approved):
Reviewed and updated. Plugin list/search now use family-indexed code-plugin and bundle-plugin streams instead of broad package scans, with cursor and merge regression coverage. Local and PR CI are green.

PR #1843, opened 2026-04-27
fix: tolerate stale auth in star status

deepujain:
## What problem was happening

Skill detail pages could crash from the read-only `stars:isStarred` query when the browser believed the user was authenticated but the Convex user lookup was stale, deleted, deactivated, or otherwise unavailable. In that state the page did not need to fail; it only needed to know whether to render the star button as active.

## Why this is the right fix

Star status is a read-only affordance. If the active user cannot be resolved, the safest fallback is `false`, while the write paths that mutate stars should continue to require a valid active user.

## What changed

- `stars.isStarred` now uses optional active-user resolution and returns `false` for stale auth.
- `soulStars.isStarred` gets the same defensive behavior for the equivalent soul detail affordance.
- Added regressions for stale skill-star auth, stale soul-star auth, and the still-working active-user starred case.

## What did not change

- Star/unstar mutations still require an active authenticated user.
- This does not make moderation-hidden or soft-deleted skills public; it only prevents the star status query from crashing the page.
- Merging to `main` does not deploy production; production deploy remains a manual workflow.

## Validation

- `./node_modules/.bin/vitest run convex/stars.test.ts src/__tests__/skill-detail-page.test.tsx`
- `git diff --check`

### DIFF
--- a/convex/soulStars.ts
+++ b/convex/soulStars.ts
@@ -1,12 +1,13 @@
 import { v } from "convex/values";
 import { mutation, query } from "./functions";
-import { requireUser } from "./lib/access";
+import { getOptionalActiveAuthUserId, requireUser } from "./lib/access";
 import { toPublicSoul } from "./lib/public";
 
 export const isStarred = query({
   args: { soulId: v.id("souls") },
   handler: async (ctx, args) => {
-    const { userId } = await requireUser(ctx);
+    const userId = await getOptionalActiveAuthUserId(ctx);
+    if (!userId) return false;
     const existing = await ctx.db
       .query("soulStars")
       .withIndex("by_soul_user", (q) => q.eq("soulId", args.soulId).eq("userId", userId))
--- a/convex/stars.test.ts
+++ b/convex/stars.test.ts
@@ -9,20 +9,22 @@ vi.mock("@convex-dev/auth/server", () => ({
   getAuthUserId: vi.fn(),
 }));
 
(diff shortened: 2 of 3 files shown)


deepujain:
Pushed a small test hardening pass for the Greptile P2s: added symmetric soul-star happy-path coverage and guarded the Convex `_handler` unwrap so a future wrapper change fails with an explicit test error. I left the file location unchanged to avoid noisy movement; this repo already has several top-level Convex test files.\n\nLocal validation run:\n- `./node_modules/.bin/vitest run convex/stars.test.ts`\n- `./node_modules/.bin/tsc --noEmit`\n- `git diff --check`

deepujain:
Follow-up pushed after CI lint caught the generic helper shape: rewrote the test `_handler` unwrap to avoid the unnecessary type parameter while keeping the explicit guard and typed handler aliases.\n\nAdditional local validation run:\n- `./node_modules/.bin/oxlint --type-aware --tsconfig ./tsconfig.oxlint.json ./convex ./packages/clawhub/src ./packages/schema/src`\n- `./node_modules/.bin/vitest run convex/stars.test.ts`\n- `./node_modules/.bin/tsc --noEmit`\n- `git diff --check`\n\nCI is rerunning on the corrected head.

steipete (approved):
Reviewed locally. The read-only star status query now tolerates stale auth while star/unstar mutations still require an active user. Focused tests and TypeScript checks pass.

Issue #1820, opened 2026-04-25
ClawHub does not allow the publication of plugins


FJRG2007:
ClawHub does not allow the publication of add-ons either via the CLI or via the web; both options return the same error: GitHub rate limit

```
GitHub API rate limit exceeded — please try again in a few minutes
```

steipete:
Resolved on `main` in bc06c47.

The trusted-publisher GitHub repository identity lookup now uses the configured Convex `GITHUB_TOKEN`, matching the existing GitHub account-age lookup behavior. This avoids falling back to GitHub's anonymous API limit during plugin/package publish setup. Docs now call out that `GITHUB_TOKEN` applies to publish-gate GitHub API calls.

PR #1846, opened 2026-04-27
fix: authenticate GitHub repository lookups

deepujain:
## Summary

GitHub repository identity lookups used by the plugin trusted-publisher flow now send the deployment `GITHUB_TOKEN` when it is configured. That keeps the lookup aligned with the existing GitHub account-age check and avoids falling back to GitHub's much smaller anonymous API rate limit during publish/setup.

Refs #1820.

## What Changed

- Added shared headers for GitHub repository lookups, including `Accept`, `User-Agent`, and optional `Authorization`.
- Reused `GITHUB_TOKEN` as a bearer token when present, matching the repo's existing GitHub metadata lookup pattern.
- Added a regression test that proves repository identity fetches include the configured token and still parse the response.

## What Did Not Change

- This does not bypass GitHub ownership, account-age, or trusted-publisher checks.
- Deployments still need `GITHUB_TOKEN` configured for authenticated GitHub API access.
- Production deploys remain manual after merge.

## Validation

- `./node_modules/.bin/vitest run convex/lib/githubActionsOidc.test.ts convex/httpApiV1.handlers.test.ts`
- `git diff --check`

### DIFF
--- a/convex/lib/githubActionsOidc.test.ts
+++ b/convex/lib/githubActionsOidc.test.ts
@@ -3,6 +3,7 @@
 import { describe, expect, it, vi } from "vitest";
 import {
   extractWorkflowFilenameFromWorkflowRef,
+  fetchGitHubRepositoryIdentity,
   verifyGitHubActionsTrustedPublishJwt,
   type TrustedGitHubActionsPublisher,
 } from "./githubActionsOidc";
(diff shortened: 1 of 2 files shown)


steipete:
Landed on `main` in bc06c47.

I also addressed the review nit by adding coverage for the blank/no-auth-token path, so repository lookups now have tests for both:
- sending `Authorization: Bearer <GITHUB_TOKEN>` when configured
- omitting `Authorization` when the token is blank

Docs were updated to clarify that `GITHUB_TOKEN` covers publish-gate GitHub API calls, including trusted-publisher repository identity lookups, and changelog credit was added.

Validation: local full gate passed (`docs:list`, `git diff --check`, focused Vitest, lint, root/package typechecks, full tests, coverage, build), and main CI + secret scan are green for bc06c47.

Thanks @deepujain.

PR #1855, opened 2026-04-28
feat(cli): show skill moderation in inspect

### LINKED ISSUE #1483: Allow to view security from CLI

deepujain:
## Summary

Fixes #1483 by making `clawhub inspect <slug>` show the skill-level moderation/security status returned by the v1 skill detail API, without requiring a separate version fetch. JSON output now also preserves the `moderation` block so scripts can filter for clean/suspicious/malicious skills.

## What changed

- Print moderation verdict, reason codes, update timestamp, engine version, and summary in plain `clawhub inspect` output.
- Include `moderation` in `--json` inspect output.
- Allow version-level `security` metadata through the CLI schema so existing version security output remains typed.
- Added regression tests for plain inspect output and JSON output.

## What did not change

- This does not add install-time filtering or automatic benign-only install selection.
- This does not change backend moderation or scanner behavior.

## Validation

- `git diff --check`
- `../../node_modules/.bin/vitest run -c vitest.config.ts src/cli/commands/inspect.test.ts`

### DIFF
--- a/packages/clawhub/src/cli/commands/inspect.test.ts
+++ b/packages/clawhub/src/cli/commands/inspect.test.ts
@@ -132,6 +132,80 @@ describe("cmdInspect", () => {
     expect(mockLog).toHaveBeenCalledWith("Model: gpt-5.2");
   });
 
+  it("prints skill moderation status without requiring a version fetch", async () => {
+    httpMocks.apiRequest.mockResolvedValueOnce({
+      skill: {
+        slug: "demo",
+        displayName: "Demo",
+        summary: null,
+        tags: { latest: "2.0.0" },
+        stats: {},
+        createdAt: 1,
+        updatedAt: 2,
+      },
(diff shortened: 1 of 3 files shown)


vincentkoc:
Thanks. The inspect moderation output is now on main via https://github.com/openclaw/clawhub/commit/1461d0fcfcf1db373b1a56fdc802f57d64bfe36f.

The landed version keeps the already-present API schema and adds plain-text + JSON CLI output with tests. Validation: package Vitest, formatter, CLI typecheck, plus the later full lint/typecheck pass. Closing this draft PR as superseded by main.

PR #1869, opened 2026-04-28
[codex] Use GitHub App auth for publish gate lookups

Patrick-Erichsen:
## Summary

- extract the existing GitHub App installation-token signing flow from backup helpers into a shared Node helper
- let package, skill, and soul publish account-age lookups use GitHub App auth when `GITHUB_TOKEN` is absent
- let trusted-publisher repository identity lookups use the same fallback
- keep `GITHUB_TOKEN` as the first-choice override and leave non-publish comment gates on the existing behavior

## Why

Production already has GitHub App credentials, but publish-gate GitHub reads only used `GITHUB_TOKEN`. That meant ClawHub could still hit unauthenticated GitHub API limits unless a PAT was configured. This reuses the backup auth path instead of requiring a long-lived PAT for the publish flow.

## Validation

- `bun run test convex/lib/githubAppAuth.test.ts convex/lib/githubAccount.test.ts convex/lib/githubActionsOidc.test.ts`
- `bun run lint`

### DIFF
--- a/convex/lib/githubAccount.test.ts
+++ b/convex/lib/githubAccount.test.ts
@@ -44,7 +44,11 @@ describe("requireGitHubAccountAge", () => {
     const fetchMock = vi.fn();
     vi.stubGlobal("fetch", fetchMock);
 
-    await requireGitHubAccountAge({ runQuery, runMutation } as never, "users:1" as never);
+    await requireGitHubAccountAge(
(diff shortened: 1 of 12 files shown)


Patrick-Erichsen:
Closing in favor of the quick GITHUB_TOKEN configuration path; the GitHub App auth refactor is not needed right now.

PR #1879, opened 2026-04-29
fix(slug): enforce length, pattern, and reserved-word rules on skill & soul slugs

momothemage:
## Background

Today the only slug check on write paths is a single regex:

```ts
/^[a-z0-9][a-z0-9-]*$/;
```

This accepts a number of values that are unsafe in practice:

| Case                        | Example                                                                                            | Why it's bad                                                                   |
| --------------------------- | -------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------ |
| No max length               | `a-a-a-…-a` (MBs)                                                                                  | Bloats the `by_slug` index, URLs, SEO, log lines                               |
| No min length               | `"a"`                                                                                              | Collides with single-letter route segments                                     |
| No reserved-word blocklist  | `admin`, `api`, `settings`, `souls`, `u`, `login`, `packages`, `plugins`, `openclaw`, `clawhub`, … | Can shadow/confuse our own routes (`/owner/$slug` vs `/admin`, `/souls`, etc.) |
| Consecutive hyphens allowed | `a---b`                                                                                            | Visual/matching oddities in UI                                                 |

The rename flow (`renameOwnedSkillByActor`) and the publish flow (`publishVersionForUser`, `publishSoulVersionForUser`) all inline the same weak regex, so any fix has to happen in several places at once — which is exactly why this PR extracts a single validator.

### Added

- **`convex/lib/skillSlugValidator.ts`** — single source of truth for slug rules:
  - length `3..48` (upper bound aligned with the publisher handle limit, so `/{owner}/{slug}` URLs stay balanced)
  - pattern `^[a-z0-9]+(-[a-z0-9]+)*$` (no leading / trailing / consecutive hyphens, no `_`, `.`, or spaces)
  - reserved-word blocklist covering our route segments and brand terms
  - public API:
    - `assertValidSkillSlug(input)` — throws `ConvexError` with a specific reason (length / pattern / reserved)
    - `isValidSkillSlugShape(input)` — shape-only, used by read/search paths
    - `isReservedSkillSlug(slug)`
    - `normalizeSkillSlug(input)` / `normalizeSkillSlugOrNull(input)` — lowercase + trim, no validation
- **`convex/lib/skillSlugValidator.test.ts`** — unit tests covering accepted / rejected shapes, the reserved-word matrix, and the `allowReserved` escape hatch used for internal callers that need to reference legacy data.

### Changed — write paths (now fully validated)

- **`convex/lib/skillPublish.ts`** — `publishVersionForUser` uses `assertValidSkillSlug` instead of the local regex, so reserved / oversized / malformed slugs can no longer reach the DB via publish.
- **`convex/lib/soulPublish.ts`** — `publishSoulVersionForUser` uses `assertValidSkillSlug` too. Souls share the same rules with skills so `/souls/$slug` routing stays safe.
- **`convex/skills.ts`**:
  - `normalizeSkillSlugForWrite` delegates to `assertValidSkillSlug`.
  - `renameOwnedSkillByActor` now validates `newSlug` with the full validator (this was the most exploitable gap).
  - `normalizeSkillSlugKey` (read-path) now uses `normalizeSkillSlug` and is intentionally lenient — see note below.
- **`convex/souls.ts`** — same split: `normalizeSoulSlugForWrite` validates, `normalizeSoulSlugKey` only normalizes.

### Changed — read paths (shape-only, legacy-safe)

- **`convex/search.ts`**:
  - `isSlugLikeQuery` → `isValidSkillSlugShape`
  - `lexicalFallbackSkills` exact-slug lookup → `normalizeSkillSlug` + `isValidSkillSlugShape`
  - **Reserved-word blocklist is intentionally NOT consulted here** so that any legacy row whose slug happens to match a newly-reserved word remains searchable.

## Design notes

- **Write-strict / read-lenient split.** Existing rows in production may carry slugs that the new validator would reject (single-char, reserved words, very long, …). Blocking reads would be a regression. So writes use `assertValidSkillSlug`, reads use `isValidSkillSlugShape` + `normalizeSkillSlug`. Grandfathered rows keep working until their owner renames them — at which point the new rules apply.
- **No data migration in this PR.** We consciously don't sweep existing data. If we later decide to retroactively clean up, it should be its own PR with an admin tool + communication to affected owners.
- **Error messages are specific.** Rather than a generic "invalid slug", callers get "Slug must be at least 3 characters", "Slug contains consecutive hyphens", "Slug is reserved", etc., which is much more useful in CLI output.
- **Souls reuse the skill validator on purpose.** Their routing constraints are the same, and keeping one validator avoids divergence.

### DIFF
--- a/convex/lib/skillPublish.ts
+++ b/convex/lib/skillPublish.ts
@@ -34,7 +34,7 @@ import {
   parseFrontmatter,
   sanitizePath,
 } from "./skills";
-import { assertValidSkillSlug } from "./skillSlugValidator";
+import { assertValidSkillSlug, normalizeSkillSlug } from "./skillSlugValidator";
 import { generateSkillSummary } from "./skillSummary";
 import { runStaticPublishScan } from "./staticPublishScan";
 import type { WebhookSkillPayload } from "./webhooks";
@@ -91,8 +91,14 @@ export async function publishVersionForUser(
   options: PublishOptions = {},
 ): Promise<PublishResult> {
   const version = args.version.trim();
-  // Full slug validation: length, pattern, reserved-word blocklist, etc.
-  const slug = assertValidSkillSlug(args.slug);
+  // Normalize first so we can look up the existing skill before deciding
+  // how strictly to validate. The reserved-word blocklist and length floor
+  // are only enforced for brand-new skills; owners of grandfathered slugs
+  // (reserved, <3 chars, or >48 chars) must still be able to publish new
+  // versions without being blocked by the write-path validator.
+  const normalizedSlug = normalizeSkillSlug(args.slug);
+  if (!normalizedSlug) throw new ConvexError("Slug is required.");
+
   const displayName = args.displayName.trim();
   if (!displayName) throw new ConvexError("Display name required");
   if (!semver.valid(version)) {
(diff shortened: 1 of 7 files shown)


momothemage:
@codex review

momothemage:
@codex review

momothemage:
Merged via squash.

- Prepared head SHA: [d93026fd4b013876a4009974b4c7052b416f3bbf](https://github.com/openclaw/clawhub/commit/d93026fd4b013876a4009974b4c7052b416f3bbf)
- Merge commit: [571a85f539025cb16ae133f544aea925583a9707](https://github.com/openclaw/clawhub/commit/571a85f539025cb16ae133f544aea925583a9707)

Thanks @momothemage!

PR #2009, opened 2026-05-04
Include package OpenClaw env metadata in ClawScan prompts

joshuaswarren:
## Summary
- parse package.json `openclaw.environment` during package LLM evaluation
- pass declared env vars and config paths into the existing scanner requirements metadata
- add coverage for required/optional env vars and OpenClaw config paths

## Why
ClawScan currently derives `environment:declared` capability tags, but the LLM prompt still reports env vars/config paths as `none`. That causes packages with correctly declared `openclaw.environment` metadata to be reviewed as underdeclared.

## Verification
- `bunx vitest run convex/llmEval.test.ts convex/lib/securityPrompt.test.ts`

### DIFF
--- a/convex/llmEval.test.ts
+++ b/convex/llmEval.test.ts
@@ -1,7 +1,7 @@
 /* @vitest-environment node */
 
 import { afterEach, describe, expect, it, vi } from "vitest";
-import { backfillLlmEval } from "./llmEval";
+import { backfillLlmEval, buildOpenClawMetadataForEval } from "./llmEval";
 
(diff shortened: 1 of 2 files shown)


joshuaswarren:
Closing this PR because it was opened too aggressively as an unsolicited code change. The underlying report is better handled first as an issue for maintainers to triage: ClawScan appears to derive environment:declared capability tags from package openclaw.environment, but the LLM prompt still reports env vars/config paths as none for package review.

PR #2051, opened 2026-05-06
fix: wait for Convex functions before seeding worktrees

vyctorbrzezowski:
## Summary
- retry seeded worktree Convex run commands while local Convex is still making functions queryable
- keep non-readiness Convex failures failing immediately
- stop the managed Convex child before returning from `--seed-only`

## Reproduction
- Before this patch, `bun run dev:worktree -- --seed-only` failed twice while Convex was still preparing functions with `Could not find function for 'devSeed:seedNixSkills'`.
- After this patch, the same command retried until Convex functions were ready, seeded successfully, exited, and left port 3210 free.

## Tests
- `bunx vitest run scripts/dev-worktree.test.ts`
- `bun run format:check`
- `bun run lint`
- `git diff --check`
- `bun run dev:worktree -- --seed-only`

## AI assistance
AI-assisted; I reviewed the patch and verified the focused behavior locally.

### DIFF
--- a/scripts/dev-worktree.test.ts
+++ b/scripts/dev-worktree.test.ts
@@ -1,5 +1,10 @@
 import { describe, expect, it } from "vitest";
-import { buildEnvFileCandidates, parseEnv, parseGitWorktreeList } from "./dev-worktree";
+import {
+  buildEnvFileCandidates,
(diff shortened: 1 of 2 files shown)


Patrick-Erichsen:
Thanks for the fix!

PR #2073, opened 2026-05-07
Fix plugin publish ownership visibility

Patrick-Erichsen:
## Summary
- enforce scoped plugin package names against the selected owner in shared schema/backend validation and surface the mismatch in the publish UI
- add package soft-delete audit logging, block releases onto soft-deleted packages, and ignore soft-deleted runtime id holders for fresh publishes
- move publish pages to `/plugins/publish` and `/skills/publish` with legacy redirects for `/publish-plugin` and `/publish-skill`
- reserve route-colliding public names: unscoped package `publish`, plus owner handles `skills` and `plugins`
- update publish route links, tests, CODEOWNERS, and CodeQL publish-surface metadata

## Validation
- `bunx vitest run packages/schema/src/schemas.test.ts convex/packages.public.test.ts src/__tests__/plugins-publish-route.test.tsx src/__tests__/skills-publish-route.test.tsx src/__tests__/publish-redirect-routes.test.ts src/components/Footer.test.tsx src/routes/-dashboard.test.tsx src/__tests__/import.route.test.tsx`
- `bun run format:check`
- `bun run lint`
- `bunx tsc --noEmit`

## Review
- `/review` confirmed in thread before publication


VintageAyu:
🦞

BunsDev (commented):
Directionally good, but I’d want one explicit behavior proof + one policy confirmation before merging.

What I like:
- Fixes the important ownership mismatch: `@openclaw/foo` can’t be published while selecting `@vintageayu`.
- Enforces it in both UI and backend — good, not just frontend polish.
- Adds audit logging for package deletion.
- Blocks new releases on soft-deleted packages.
- Keeps legacy publish URLs with redirects.
- Checks are green on the PR.

Main caution:

The runtime-id behavior is policy-sensitive: this allows a new package to claim a plugin/runtime ID if the previous holder is soft-deleted.

That may be intended for spam cleanup, but I’d like explicit confirmation on the policy:

> When a package is soft-deleted, should its runtime ID become reusable immediately, or should it stay reserved unless an admin explicitly releases it?

If runtime IDs are used by clients/install flows, automatic reuse could create confusion or a takeover-shaped edge case.

Before merge, I’d ask for:
1. Screenshot/recording/log showing the mismatch blocked in the publish UI.
2. Backend proof showing mismatched scoped name + owner is rejected.
3. Confirmation that runtime ID reuse after soft-delete is desired policy.

So: approveable after proof, but I’d prefer not to merge completely blind on the runtime ID reuse decision.

BunsDev (commented):
Duplicate review — removing body to avoid noise.

Patrick-Erichsen:
Confirmed policy: soft-deleted packages leave the active namespace immediately. Active packages still reserve runtime IDs, but a runtime ID held only by soft-deleted packages can be claimed by a new package. If we add package restore later, restore should fail or require admin repair when the name/runtime ID has already been reclaimed.

Proof logs from the PR branch:

```text
$ bunx vitest run src/__tests__/plugins-publish-route.test.tsx --testNamePattern "blocks scoped package names" --reporter verbose
✓ plugins publish route > blocks scoped package names that do not match the selected owner
```

This renders the publish UI with package `@openclaw/dronzer` and selected owner `@vintageayu`, verifies the mismatch message is shown, verifies Publish is disabled, and verifies `publishRelease` is not called.

```text
$ bunx vitest run convex/packages.public.test.ts --testNamePattern "rejects package scopes|allows plugin ids held only by soft-deleted packages|rejects plugin id collisions with active packages|rejects new releases on a soft-deleted package" --reporter verbose
✓ rejects new releases on a soft-deleted package
✓ rejects package scopes that do not match the selected owner handle
✓ rejects plugin id collisions with active packages
✓ allows plugin ids held only by soft-deleted packages
```

This exercises the backend publish action path for the scoped-owner mismatch and verifies the publish mutation is not called after rejection.

```text
$ bunx vitest run packages/schema/src/schemas.test.ts --testNamePattern "reports scoped package names" --reporter verbose
✓ reports scoped package names that do not match the selected owner
```

PR #2079, opened 2026-05-07
fix(search): disclose suspicious results hidden on unified search

momothemage:
## Summary

Unified search (homepage / `/search`) and the `/skills` browse page used
**different default values** for the moderation filter
`nonSuspiciousOnly`:

| Surface            | Hook / query call                                              | Default            |
| ------------------ | -------------------------------------------------------------- | ------------------ |
| `/search` (before) | `searchSkills({ ..., nonSuspiciousOnly: true })` _(hardcoded)_ | always `true`      |
| `/skills`          | `search.nonSuspicious ?? false`                                | `false` (show all) |

The same query therefore returned **less result on the
homepage but more results on `/skills`**, with nothing in the UI to
explain why. This was confusing for end users (they thought search was
broken) and for publishers (they thought their skill was missing).

Closes the linked moderation-visibility issue.

### 1. `useUnifiedSearch` — make the filter a parameter

`src/lib/useUnifiedSearch.ts`

- `UnifiedSearchOptions` now accepts an optional `nonSuspiciousOnly`.
- Default is `true` to **preserve the current moderation-safe default**
  for every caller that hasn't been updated.
- The flag is forwarded to `searchSkills(...)` and added to the
  effect's dependency array so a flip triggers a fresh recall.

### 2. `/search` — read URL, render toggle chip, persist opt-out

`src/routes/search.tsx`

- `validateSearch` now also parses `nonSuspicious` (mirroring the
  `/skills` URL parameter name). Only the **opt-out** (`false`) is
  persisted in the URL; the default remains absent for a clean URL.
- Page reads `search.nonSuspicious ?? true` and feeds it to the hook.
- Adds `<SuspiciousFilterChip>` styled to match the `FilterChip` used
  by `/skills`. It is rendered only when a query is present and the
  active tab is not `plugins`. Labels:
  - default state: `✓ Hiding suspicious skills · Show all`
  - opt-out state: `Showing all skills · Hide suspicious`
- Clicking the chip flips the URL param (default → `false`, `false` →
  cleared). The `replace: true` navigation avoids polluting browser
  history.
- `setType` and `handleSearch` preserve the user's `nonSuspicious`
  choice across tab/query changes.

#### New: hook-level coverage — `src/lib/useUnifiedSearch.test.tsx`

- `defaults nonSuspiciousOnly to true when option is omitted`
- `forwards nonSuspiciousOnly=false when caller opts out`
- `re-runs the skills query when nonSuspiciousOnly flips`
- `does not call searchSkills when activeType is plugins`

#### Extended: `src/__tests__/search-route.test.tsx`

- The existing `can request more results from global search` assertion
  was updated to include `nonSuspiciousOnly: true` (the new default
  payload from the page).
- Added five new cases:
  1. `defaults nonSuspiciousOnly to true when URL param is absent`
  2. `passes nonSuspiciousOnly=false to the hook when URL opts out`
  3. `shows the 'Hiding suspicious skills' chip by default and opts out
     on click`
  4. `shows the 'Showing all skills' chip when opted out and clears
     the URL param on click`
  5. `hides the suspicious-filter chip on the plugins tab`

## Out of scope (intentionally not changed)

- **`/skills` defaults are unchanged** (`nonSuspicious ?? false`).
  `/skills` is an explorer surface with its own toggle UI and a long
  history of that default; flipping it would be a breaking UX change
  unrelated to this bug.
- **Header / navbar typeahead** uses `useUnifiedSearch` without passing
  `nonSuspiciousOnly` and continues to inherit the moderation-safe
  default. No change needed.
- **`convex/search.ts` `searchSkills` API** is untouched. The same
  parameter that has always existed is now passed by the page; no new
  fields, no new return shape.

## Risk & rollback

- Risk is contained: the only behavior change visible to users is on
  `/search`, and the default state matches the previous hardcoded
  behavior. The opt-out is fully URL-encoded so it can be disabled by
  reverting this commit alone.
- All existing tests around the homepage (`home-route.test.tsx`),
  navbar (`header.test.tsx`), and `/skills` (`skills-index.test.tsx`,
  `skills-toolbar.test.tsx`) continue to pass without modification.

## Verification

```sh
bun run test -- \
  src/__tests__/search-route.test.tsx \
  src/lib/useUnifiedSearch.test.tsx \
  src/__tests__/home-route.test.tsx \
  src/__tests__/header.test.tsx \
  src/__tests__/skills-index.test.tsx \
  src/__tests__/skills-toolbar.test.tsx
# → 48 passed (48)

bun run lint        # → 0 warnings, 0 errors
bun run format:check # → all 696 files OK
bunx tsc --noEmit   # → clean
```

## Manual repro of the original bug

Before this change:

1. `https://clawhub.ai/?q=xxx` → less skill
2. `https://clawhub.ai/skills?q=xxx` → more skills
After this change:

1. `/search?q=xxx` → less skill, but with a visible chip explaining
   `Hiding suspicious skills · Show all`
2. Click `Show all` → URL becomes `/search?q=xxx&nonSuspicious=false`
   → more skills, chip flips to `Showing all skills · Hide suspicious`

The `/skills` page returns the same results as before.

## Screenshots / UX notes

- Chip styling is intentionally identical to the `FilterChip` in
  `src/routes/skills/-SkillsToolbar.tsx` (same border/bg tokens,
  `min-h-[36px]`, light/dark adaptations) so users feel the two
  surfaces are part of the same system.
- The chip is **only** rendered when a query has been entered and the
  active tab is not `plugins`; it would be misleading otherwise since
  the flag has no effect on plugin recall.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Skills: repair publisher-owned skill merges, bound historical slug redirects, block protected slug namespaces, and expire owner-unpublished slug reservations after 30 days (#2115) (thanks @fuller-stack-dev).
 - Skills: allow confirmed owner migration when republishing an existing skill to another publisher, preserving versions, stats, aliases, and audit history (#1998, #2102) (thanks @momothemage).
 - Security: block owner delete/undelete paths from overriding moderator or scanner hides, and return explicit 403 authz responses for owner restore denials (#2078) (thanks @momothemage).
+- Search/Web: disclose when `/search` is hiding suspicious skills and add an explicit opt-out so unified search no longer silently differs from `/skills` for the same query (#2079) (thanks @momothemage).
 
 ## 0.12.3 - 2026-05-06
 
--- a/src/__tests__/search-route.test.tsx
+++ b/src/__tests__/search-route.test.tsx
@@ -5,7 +5,11 @@ import type { ComponentType, ReactNode } from "react";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 
 const navigateMock = vi.fn();
-let searchMock: { q?: string; type?: "all" | "skills" | "plugins" } = {};
+let searchMock: {
+  q?: string;
+  type?: "all" | "skills" | "plugins";
+  nonSuspicious?: boolean;
+} = {};
 const useUnifiedSearchMock = vi.fn();
 
 vi.mock("@tanstack/react-router", () => ({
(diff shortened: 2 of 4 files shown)


momothemage:
For [P2] Gate vector suspicious matches on query tokens — convex/search.ts:448-452

1. The lines in question are in directPrefixSkills, not the vector path. Lines 448–452 sit inside the per-digest filter loop of the directPrefixSkills internal query. Every digest that reaches that loop has already been selected by one of four range scans on normalizedSlug / normalizedDisplayName / normalizedSlugFirstToken / normalizedDisplayNameFirstToken against the query's first token (lines ~380–435). In other words, the index scan itself is the token gate for this path — it is a prefix recall by construction.
2. Adding matchesExactTokens here would degrade prefix recall into exact-token recall. Users typing gua and expecting guardian to surface is exactly what this path is for. The vector path already runs matchesExactTokens on its own candidates (around line 244) and the two pools are merged via mergeUniqueBySkillId(directMatches, exactMatches). The split is intentional: vector contributes token-gated semantic recall, directPrefixSkills contributes high-confidence prefix recall. Collapsing them would lose the latter.
3. The suspicious concern is already covered by the default. nonSuspiciousOnly defaults to true on every public entry point (homepage unified search, /search, /skills). Suspicious skills are only reachable when a viewer explicitly opts in via the "Show all" chip, which is the moderation-visibility surface. An additional token gate on the opt-in path doesn't reduce the attack surface — it just hides legitimate prefix matches from moderators.
4. Net effect of the proposed change
Prod users (default nonSuspiciousOnly=true): no change, suspicious entries are filtered before the gate would run.
Opt-in / moderation viewers: would lose prefix-only matches that are exactly what they need to see.


Re-review progress:
- State: Complete
- Detail: The targeted re-review finished, the durable review comment was updated, and the synced verdict was routed.
- Run: https://github.com/openclaw/clawsweeper/actions/runs/25597426919
- Updated: 2026-05-09T09:19:53.350Z

momothemage:
Merged via squash.

- Prepared head SHA: [1477ca7120eaf0d4cd1de332f45b92302c7b9e36](https://github.com/openclaw/clawhub/commit/1477ca7120eaf0d4cd1de332f45b92302c7b9e36)
- Merge commit: [df7e0f77a6bb47bb75cd73657fdc00ab98cd6763](https://github.com/openclaw/clawhub/commit/df7e0f77a6bb47bb75cd73657fdc00ab98cd6763)

Thanks @momothemage!

PR #2138, opened 2026-05-11
Export skills function added by Mirror Site maintainer

wuchengan734-a11y:
## Summary

We are the ClawHub China mirror site from bytedance. To keep our mirror in sync with the upstream skill registry, we need an efficient way to bulk-fetch skill data. Currently there is no batch export API — the only option is to download skills one by one, which triggers rate limits and takes hours for 62k+ skills.

This PR adds a new `GET /api/v1/skills/export` endpoint that returns a merged ZIP archive of up to 1000 skills per request, with cursor-based pagination for full-catalog traversal.

### What's changed

- **New endpoint**: `GET /api/v1/skills/export?startDate=<ms>&endDate=<ms>&limit=<n>&cursor=<token>`
  - Returns a ZIP containing skill files organized in `<slug>/` subdirectories
  - Includes `_manifest.json` (skill metadata array) and `_meta.json` per skill
  - Supports cursor-based pagination via `X-Next-Cursor` / `X-Has-More` response headers

- **Authentication & Authorization** (`convex/lib/apiTokenAuth.ts`)
  - New `requireExportAuth()` — enforces Bearer token + admin-only role check
  - No token → 401, valid token but non-admin → 403, admin → allowed
  - Role whitelist configurable via `EXPORT_ALLOWED_ROLES` array

- **Security hardening** (`convex/lib/skillZip.ts`)
  - `validateSlug()` — prevents Zip Slip via malicious slug paths (`../../etc`)
  - `validateFilePath()` — blocks absolute paths, `..` traversal, backslashes, empty segments
  - `buildMergedExportZip()` — detects duplicate ZIP paths to prevent silent overwrites

- **Error recording** (`convex/httpApiV1/skillsV1.ts`)
  - Missing versions or blobs are logged to `_errors.json` inside the ZIP (not silently skipped)
  - `X-Export-Errors` response header reports error count

- **Rate limiting** (`convex/lib/httpRateLimit.ts`)
  - New `export` tier: `{ ip: 10, key: 60, adminKey: 600 }` requests/min

- **Query layer** (`convex/skills.ts`)
  - `listByDateRange` internalQuery — range scan on `skillSearchDigest.by_active_created` index
  - Chunked parallel blob reads (50 concurrent) to stay within Convex transaction limits

### Files changed (16)

| File | Change |
|------|--------|
| `packages/schema/src/routes.ts` | Add `skillsExport` route constant |
| `convex/lib/apiTokenAuth.ts` | Add `requireExportAuth` (admin-only, enabled) |
| `convex/lib/httpRateLimit.ts` | Add `export` rate limit tier |
| `convex/lib/skillZip.ts` | Add `validateSlug`, `validateFilePath`, `buildMergedExportZip` |
| `convex/skills.ts` | Add `listByDateRange` internalQuery |
| `convex/httpApiV1/skillsV1.ts` | Add `exportSkillsV1Handler` with security checks |
| `convex/httpApiV1.ts` | Wire up httpAction export |
| `convex/http.ts` | Register GET route |
| `convex/devSeedExport.ts` | Dev seed: 200 test skills for export testing |
| `test_export_api.sh` | Integration test script (31 tests) |
| `convex/lib/access.ts` | Minor type alignment |
| `convex/_generated/api.d.ts` | Auto-generated |
| `packages/schema/dist/*` | Built output |
| `readme_mirror_sync.md` | Mirror sync documentation |

## Testing

> All checks below were **actually executed** against the local Convex dev server and verified.
- ✅ **TypeScript**: `npx tsc --noEmit` — 0 errors
- ✅ **Unit tests**: `bun run test` (vitest) — 180 test files, 1701 tests passed, 0 failed
- ✅ **Integration tests**: `bash test_export_api.sh` — **31/31 passed**, covering:
  - Auth (6 tests): no token→401, invalid→401, user→403, admin→200, case-insensitive bearer, empty bearer
  - Param validation (4 tests): missing startDate/endDate→400, startDate>endDate→400
  - Empty results (3 tests): future date range→empty ZIP, correct headers
  - Export output (5 tests): valid ZIP, correct Content-Type/Disposition, error/date headers
  - ZIP content (5 tests): _manifest.json valid, skill subdirs with _meta.json, no path traversal
  - Pagination (2 tests): cursor + hasMore flow, second page fetch
  - Rate limit (2 tests): X-RateLimit headers present
  - Edge cases (4 tests): limit clamping, single-point date range

### No breaking changes
- All existing endpoints are unaffected
- The new `export` rate limit tier is additive (existing `read`/`write`/`download` unchanged)
- `skillZip.ts` additions are new exports; existing `buildDeterministicZip` unchanged

### Reviewer attention points
1. Memory usage: The handler loads up to 1000 skills × N files into memory before zipping. For the current average file size (~10KB/file, ~5 files/skill) this is ~50MB peak, within Convex Action limits. Consider adding a total-size guard if skill file sizes grow significantly.
2. `adminKey: 600`: The admin rate limit for export is generous (600 req/min). This was intentional to allow fast full-catalog sync, but could be reduced to 60-120 if the team prefers a more conservative setting.
3. `Uint8Array.from(zipSync(...))` in `buildMergedExportZip` creates a copy of the ZIP buffer. Could be optimized to return `zipSync()` directly if memory pressure is a concern.

### Context
We are the **ClawHub China mirror site** operators. China mainland users experience high latency and intermittent connectivity when accessing clawhub.ai directly. This export API enables us to efficiently bulk-sync the full skill catalog to our mirror, keeping it up-to-date with hourly incremental exports using date-range queries + cursor pagination. Without this endpoint, syncing 62k+ skills requires individual downloads that take 10+ hours and frequently hit rate limits.


Patrick-Erichsen:
@wuchengan734-a11y I pushed a follow-up commit to this branch with some updates:

- export now requires a normal API token but is no longer site-admin-only
- authenticated users use the export key tier at 60 requests/min; no separate mirror role
- exportable content is limited to public/installable/non-hidden/non-deleted/non-malicious skills
- incremental sync now uses updatedAt
- ZIP entries are namespaced as <publisher>/<slug>/... so duplicate slugs across publishers do not collide
- added request file/byte caps and regression coverage

Verified with focused handler/query tests, format, lint, TypeScript, and autoreview. Could you please verify the updated branch against your mirror flow?

---

The above was generated by my AI, but @wuchengan734-a11y , I don't think we need to require this to be admin only. I don't think we would grant you guys admin access just to do this export. 

Will the 60 req/min work for your needs? We could definitely bump this but I just used the `key: 60` that you had scaffolded for authenticated users.

Note that I also updated the exported folder structure to use `<publisher>/<slug>` since slugs are no longer globally unique.

Patrick-Erichsen:
@wuchengan734-a11y also note that in https://github.com/openclaw/clawhub/pull/2376 I capped the pagination at 250 skills. Beyond that Convex was crashing consistently.

wuchengan734-a11y:
Thanks for the update. I agree that requiring site-admin access is unnecessary for our mirror flow, so using a normal API token sounds good.

60 req/min should likely be fine for incremental sync. We’ll verify whether it is sufficient for initial backfill/full mirror runs and let you know if we need a higher burst or rate limit.

The <publisher>/<slug> folder structure also makes sense given that slugs are no longer globally unique. We’ll update/verify our mirror-side parsing accordingly.

We’ll test the updated branch against our mirror flow, thanks for your feedback.

PR #2140, opened 2026-05-11
fix(search): recall non-first-token matches via full-text indexes

### LINKED ISSUE #2137: [Search] Skill search underrecall: non-first-token keywords never match, visibility decays over time
## Summary

Skills are systematically invisible in search when the query keyword is **not a string prefix** of the skill's name or slug, and **not the first token**.

Example — `baidu-yijian-vision` (displayName: `Baidu Yijian Vision`):

| Query | Result | Why |
|-------|--------|-----|
| `baidu yijian` | ✅ rank #1 | slug first token `"baidu"` hits the prefix index |
| `baidu` | ✅ found | same reason |
| `yijian` | ❌ zero results | non-first token, all prefix indexes miss |
| `yijian vision` | ❌ zero results | same |
| `vision` | ❌ zero results | same |

Additionally, publishing a new version temporarily restores visibility, but as other skills are updated the skill disappears again — a time-based visibility decay that makes search results unpredictable.

---

## Root Cause

The search system has three recall paths. All three are unreliable for non-first-token queries.

### Bug 1 — `directPrefixSkillMatches`: four prefix-index queries all miss (core issue)

Code review of `convex/search.ts` shows the function runs **4 string-prefix queries** against `skillSearchDigest` in parallel:

```typescript
// Query 1: full slug string prefix
.withIndex("by_active_normalized_slug", (q) =>
  q.eq("softDeletedAt", undefined)
   .gte("normalizedSlug", normalizedQuery)   // slug >= "yijian"
   .lt("normalizedSlug", upperBound)         // slug <  "yijian\uffff"
)

// Query 2: full displayName string prefix (same logic)

// Query 3: slug first-token prefix
.withIndex("by_active_normalized_slug_first_token", (q) =>
  q.eq("softDeletedAt", undefined)
   .gte("normalizedSlugFirstToken", firstToken)   // firstToken("yijian") = "yijian"
   .lt("normalizedSlugFirstToken", firstTokenUpperBound)
)

// Query 4: displayName first-token prefix (same logic)
```

Since `baidu-yijian-vision` starts with `"baidu"` in both slug and displayName, all four queries miss when the search term is `"yijian"` or `"vision"`. Because the prefix matching is inherently strict, this path produces **only underrecall, never overrecall**.

| Query | slug prefix | displayName prefix | slugFirstToken | displayNameFirstToken |
|-------|:-----------:|:------------------:|:--------------:|:---------------------:|
| `baidu yijian` | ❌ | ❌ | ✅ firstToken=`baidu` | ✅ |
| `yijian` | ❌ | ❌ | ❌ `yijian` ≠ `baidu` | ❌ |
| `vision` | ❌ | ❌ | ❌ `vision` ≠ `baidu` | ❌ |

### Bug 2 — `lexicalFallbackSkills`: bounded by `updatedAt` recency window (visibility decay)

```typescript
recentByUpdatedQuery.order("desc").take(scanLimit)  // most-recently-updated N skills only
recentByCreatedQuery.order("desc").take(scanLimit)
```

This path scans only the **most recently updated/created N skills**. After publishing a new version, the skill floats to the top of the window and becomes findable. As other skills are updated it gets pushed out of the window and disappears again.

### Bug 3 — Vector search: weak semantic signal for non-English / transliterated terms

`"yijian"` is Chinese pinyin. The embedding model has weak and inconsistent semantic associations for transliterated terms, so the skill cannot reliably enter the Top-N candidate pool. By contrast, `"vision"` finds `computer-vision-expert` (a purely English skill) because the semantic link is strong — confirming the vector path works for English but is unreliable for mixed-language skills.

---

## Fix

Add Convex full-text search indexes to the `skillSearchDigest` table (the table `directPrefixSkillMatches` actually queries) and run them alongside the existing prefix queries during a zero-downtime transition.

**No backfill code required** — Convex automatically backfills all existing documents in the background after the schema change is deployed. Index readiness can be monitored in the Convex dashboard.

### Schema change — `convex/schema.ts`

```typescript
const skillSearchDigest = defineTable({ ... })
  // ... existing indexes unchanged ...
  .searchIndex("search_by_display_name", {
    searchField: "displayName",
    filterFields: ["softDeletedAt", "isSuspicious"],
  })
  .searchIndex("search_by_slug", {
    searchField: "slug",
    filterFields: ["softDeletedAt", "isSuspicious"],
  });
```

### Query change — `convex/search.ts`

Add two full-text queries to the existing `Promise.all` in `directPrefixSkillMatches`, then spread their results into the dedup array:

```typescript
// Append to the existing Promise.all — keep the 4 prefix queries intact for now
const [...existing, ftDisplayNameDigests, ftSlugDigests] = await Promise.all([
  /* existing 4 prefix queries unchanged */,

  // NEW: full-text search on displayName — matches any token, not just the first
  args.nonSuspiciousOnly
    ? ctx.db.query("skillSearchDigest")
        .withSearchIndex("search_by_display_name", (q) =>
          q.search("displayName", args.query)
           .eq("softDeletedAt", undefined)
           .eq("isSuspicious", false))
        .take(MAX_DIRECT_SKILL_SEARCH_CANDIDATES)
    : ctx.db.query("skillSearchDigest")
        .withSearchIndex("search_by_display_name", (q) =>
          q.search("displayName", args.query).eq("softDeletedAt", undefined))
        .take(MAX_DIRECT_SKILL_SEARCH_CANDIDATES),

  // NEW: full-text search on slug — matches any token, not just the first
  args.nonSuspiciousOnly
    ? ctx.db.query("skillSearchDigest")
        .withSearchIndex("search_by_slug", (q) =>
          q.search("slug", args.query)
           .eq("softDeletedAt", undefined)
           .eq("isSuspicious", false))
        .take(MAX_DIRECT_SKILL_SEARCH_CANDIDATES)
    : ctx.db.query("skillSearchDigest")
        .withSearchIndex("search_by_slug", (q) =>
          q.search("slug", args.query).eq("softDeletedAt", undefined))
        .take(MAX_DIRECT_SKILL_SEARCH_CANDIDATES),
]);

const digests = [
  ...slugDigests, ...displayNameDigests,
  ...slugFirstTokenDigests, ...displayNameFirstTokenDigests,
  ...ftDisplayNameDigests, ...ftSlugDigests,  // NEW
].filter(/* existing dedup */);
```

This resolves all three bugs simultaneously:
- **Bug 1**: full-text indexes match any token at any position — not just the first
- **Bug 2**: full-text indexes are time-invariant; visibility no longer depends on `updatedAt`
- **Bug 3**: exact token matching does not rely on semantic embeddings; transliterated terms match reliably

A patch against commit `d16a62f5` is attached.

### Two-phase zero-downtime deployment

| Phase | Action |
|-------|--------|
| **Deploy 1** | Add `searchIndex` to schema + add new full-text queries while **keeping all existing prefix queries** in parallel; results merged via existing dedup logic |
| **Deploy 2** | Once index `status = ready` (visible in Convex dashboard), remove the four legacy prefix queries and their corresponding schema indexes |

During the transition, the legacy path continues to cover all previously-working queries with no regression, while the new path progressively fills in non-first-token gaps as the backfill completes.

---

## Impact

Any skill whose name does not begin with the user's search keyword is systematically undiscoverable through keyword search — unless it happens to be caught by vector search (unreliable for non-English terms) or the recency window (temporary). This is a structural issue in the search architecture, not an isolated edge case.

momothemage:
## Summary

- What changed:
  - Added two Convex full-text `searchIndex`es to `skillSearchDigest` in `convex/schema.ts`: `search_by_display_name` and `search_by_slug`, both with `filterFields: ["softDeletedAt", "isSuspicious"]`.
  - Appended two `withSearchIndex(...)` queries to the existing `Promise.all` in `directPrefixSkillMatches` (`convex/search.ts`) so non-first-token tokens are now recalled. The four legacy prefix-index queries are kept untouched and merged through the existing `skillId`-based dedup.
  - Extended the test mock in `convex/search.test.ts` to model Convex's token-level inverted-index semantics, and added 4 regression tests (non-first-token slug, non-first-token displayName, `nonSuspiciousOnly` safety, cross-path dedup).
- Why:
  - Skill search systematically returned zero results when the user's query was not a string prefix of the skill's `slug` / `displayName` and was not the first token. For example, searching `yijian` or `vision` against `baidu-yijian-vision` produced an empty list even though the skill was active and non-suspicious.
  - The four prefix indexes used by `directPrefixSkillMatches` are anchored to the start of either the full normalized field or its first alphanumeric token, so non-first-token queries miss every path simultaneously. The lexical fallback only scans the most-recently-updated/created N rows, which explains the "publish a new version, becomes findable, then disappears again" time-decay symptom. Vector search has weak recall for transliterated tokens (e.g. Chinese pinyin), so it cannot reliably backfill the gap.
  - This is Phase 1 of a two-phase zero-downtime rollout: the legacy prefix paths stay live during the index backfill, so worst case the result set is no smaller than today; once the new search indexes show `ready` in the Convex dashboard, a follow-up PR will retire the redundant prefix indexes.

## Linked Issue

- Closes #2137 
- Related #

## Screenshots

For website/UI changes, attach screenshots or recordings from the real app. Include mobile/narrow views when layout changes.

- [x] Screenshots/recordings attached, or `N/A` — N/A (backend-only change to Convex query layer; no UI surface modified)

## Security / Trust Impact

- [x] No security/trust impact
- [ ] Security/trust impact explained

Notes (kept for reviewer context, not a separate impact):

- The new `searchIndex`es declare `softDeletedAt` and `isSuspicious` as `filterFields`, and every new query passes `.eq("softDeletedAt", undefined)` plus (when `nonSuspiciousOnly` is set) `.eq("isSuspicious", false)`, matching the existing prefix-query behavior.
- The post-hydration `isSkillSuspicious(skill)` and `isSkillHighlighted(skill)` guards in `directPrefixSkillMatches` are unchanged, so suspicious or non-highlighted skills cannot leak through the new path.

## Data / Deploy Impact

- [ ] No data/deploy impact
- [x] Data/deploy impact explained

- Schema change: two new `searchIndex` entries on `skillSearchDigest`. Convex will automatically backfill all existing documents in the background once the schema is deployed; no migration script is required.
- Index readiness can be observed in the Convex dashboard. During backfill, queries against the new indexes still execute and progressively return more matches as the backfill advances; the legacy prefix paths run in parallel, so user-visible recall never regresses below the pre-PR baseline.
- No writes to existing fields, no row migrations, no data deletions.
- Rollback is safe: removing the schema entries and the two `withSearchIndex` calls restores the previous behavior with no data fix-up needed.

## Verification

- [x] `bun run format:check` — verified via `bunx oxfmt --check` on `convex/schema.ts`, `convex/search.ts`, `convex/search.test.ts`: "All matched files use the correct format."
- [x] `bun run lint` — verified via `bunx oxlint` on the same three files: 0 warnings, 0 errors across 119 rules.
- [x] `bun run test` — `bunx vitest run convex/`: 96 files / 1176 tests passed; `bunx vitest run convex/search.test.ts`: 38 / 38 passed (4 newly added).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 ### Fixes
 
+- Search: recall skill matches by non-first slug/display-name tokens while keeping multi-token queries on the direct recall path constrained to all query tokens (#2140) (thanks @momothemage).
 - Moderation: stop treating static suspicious-only findings as a verdict; keep file/line evidence for review while VT/LLM decide public suspicious status.
 - Moderation: stop treating VirusTotal Code Insight/Palm verdicts as a hide authority for skills; real AV-engine hits and ClawScan findings still contribute moderation verdicts.
 - ClawScan: lower false positives by treating purpose-aligned notes as benign unless structured LLM findings contain a material concern, and add targeted rescan batches for suspicious skills/plugins.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## Unreleased
+
+### Changes
+
+### Fixes
+
+- Search: recall skill matches by non-first slug/display-name tokens while keeping multi-token queries on the direct recall path constrained to all query tokens (#2140) (thanks @momothemage).
+
 ## 0.13.0 - Unreleased
 
 ### Changes
@@ -14,7 +22,6 @@
 
 ### Fixes
 
-- Search: recall skill matches by non-first slug/display-name tokens while keeping multi-token queries on the direct recall path constrained to all query tokens (#2140) (thanks @momothemage).
 - Moderation: stop treating static suspicious-only findings as a verdict; keep file/line evidence for review while VT/LLM decide public suspicious status.
 - Moderation: stop treating VirusTotal Code Insight/Palm verdicts as a hide authority for skills; real AV-engine hits and ClawScan findings still contribute moderation verdicts.
 - ClawScan: lower false positives by treating purpose-aligned notes as benign unless structured LLM findings contain a material concern, and add targeted rescan batches for suspicious skills/plugins.
--- a/convex/schema.ts
+++ b/convex/schema.ts
@@ -810,7 +810,15 @@ const skillSearchDigest = defineTable({
     "isSuspicious",
     "statsInstallsAllTime",
     "updatedAt",
-  ]);
+  ])
+  .searchIndex("search_by_display_name", {
+    searchField: "displayName",
+    filterFields: ["softDeletedAt", "isSuspicious"],
+  })
+  .searchIndex("search_by_slug", {
+    searchField: "slug",
+    filterFields: ["softDeletedAt", "isSuspicious"],
+  });
 
 const packages = defineTable({
   name: v.string(),
--- a/convex/search.test.ts
+++ b/convex/search.test.ts
@@ -162,6 +162,98 @@ describe("search helpers", () => {
     );
   });
 
+  it("recalls non-first-token slug matches via the full-text search index (Bug 1)", async () => {
+    // Repro of the original bug: searching "yijian" against a skill whose
+    // slug is "baidu-yijian-vision" returned zero results because all four
+    // prefix indexes only match the *first* token. The new search index
+    // should match any token at any position.
+    const skill = makeSkillDoc({
+      id: "skills:baidu-yijian-vision",
+      slug: "baidu-yijian-vision",
+      displayName: "Baidu Yijian Vision",
+    });
+    const ctx = makeDirectPrefixCtx([skill]);
+
+    const result = await directPrefixSkillMatchesHandler(ctx, {
+      query: "yijian",
+      limit: 10,
+    });
+
+    expect(result.map((entry) => entry.skill.slug)).toEqual(["baidu-yijian-vision"]);
+    expect(ctx.usedSearchIndexes).toEqual(
+      expect.arrayContaining(["search_by_display_name", "search_by_slug"]),
+    );
+  });
+
+  it("recalls non-first-token displayName matches via the full-text search index", async () => {
+    // Companion case to the slug repro above: a query that only matches
+    // the displayName (not the slug) at a non-first position must still
+    // surface the skill.
+    const skill = makeSkillDoc({
+      id: "skills:baidu-yijian-vision",
+      slug: "baidu-yijian-vision",
+      displayName: "Baidu Yijian Vision",
+    });
+    const ctx = makeDirectPrefixCtx([skill]);
+
+    const result = await directPrefixSkillMatchesHandler(ctx, {
+      query: "Vision",
+      limit: 10,
+    });
+
+    expect(result.map((entry) => entry.skill.slug)).toEqual(["baidu-yijian-vision"]);
+  });
+
+  it("does not return suspicious skills via full-text search when nonSuspiciousOnly is set", async () => {
+    // Even though the full-text search would token-match the suspicious
+    // skill, the filterField `isSuspicious=false` plus the post-hydration
+    // `isSkillSuspicious` guard must keep it out of the results.
+    const clean = makeSkillDoc({
+      id: "skills:clean",
+      slug: "baidu-yijian-vision",
+      displayName: "Baidu Yijian Vision",
+    });
+    const flagged = makeSkillDoc({
+      id: "skills:flagged",
+      slug: "shady-yijian-trick",
+      displayName: "Shady Yijian Trick",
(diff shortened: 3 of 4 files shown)


momothemage:
Merged via squash.

- Prepared head SHA: [b597454bc242dae60bc004915cd0aaf360a3960e](https://github.com/openclaw/clawhub/commit/b597454bc242dae60bc004915cd0aaf360a3960e)
- Merge commit: [190ce3769f47335f68264cf2bc734bdaf7cfc756](https://github.com/openclaw/clawhub/commit/190ce3769f47335f68264cf2bc734bdaf7cfc756)

Thanks @momothemage!

PR #2172, opened 2026-05-12
fix(schema): add missing text file extensions tsv, conf, properties, dat

### LINKED ISSUE #874: Publish silently drops extensionless files and .tsv files
## Bug

`clawhub publish` silently skips files that don't have a recognized extension. This means:

1. **Extensionless executables** (e.g. `bin/openclaw-kraken`) are dropped — the extension extraction logic (`relPath.split('.').at(-1)`) returns the full filename which doesn't match the allowlist
2. **`.tsv` files** (e.g. `config/endpoints.tsv`) are dropped — `tsv` is not in `TEXT_FILE_EXTENSION_SET`

No warning is emitted. The publish succeeds but the installed skill is broken.

## Root cause

In `dist/skills.js`, the file walker:

```js
const ext = relPath.split('.').at(-1)?.toLowerCase() ?? '';
if (!ext || !TEXT_FILE_EXTENSION_SET.has(ext))
    return;
```

And in `dist/schema/textFiles.js`, `TEXT_FILE_EXTENSION_SET` only contains ~40 extensions. Missing at minimum: `tsv`, and extensionless files need a fallback path.

## Impact

The `kraken-cli` skill (by oscraters) publishes without its `bin/openclaw-kraken` CLI binary and `config/endpoints.tsv` / `config/futures_endpoints.tsv` registry files. The installed skill is non-functional.

## Suggested fix

1. Add `tsv` (and `dat`, `conf`, `properties`, etc.) to the extension allowlist
2. For extensionless files: check if the file is text (e.g. read first bytes, check for null bytes) rather than silently dropping
3. Emit a warning during publish for any skipped files so authors can catch this

## Environment

- clawhub 0.7.0 (also reproduced on 0.8.0)
- macOS Darwin 25.3.0

alexuser:
Fixes #874

## Problem

 silently drops  files and other valid text files because  had an incomplete extension allowlist compared to the CLI package ().

## Changes

- Added missing extensions to : , , , 
- Added canonical content types for  →  and  → 
- Added test assertions in 

## Validation

-  — 5 tests passed
-  — passed
-  — 0 warnings, 0 errors


steipete:
Thanks @alexuser. I landed this manually on `main` in `b1206ed9`.

What shipped:
- schema text extensions now include `.tsv`, `.conf`, `.properties`, and `.dat`
- canonical MIME types use `text/csv` and `text/tab-separated-values`
- generated schema dist files are updated too

Proof:
- `bun run test -- packages/schema/src/textFiles.test.ts`
- dist import smoke confirmed the new extensions and MIME values
- final `main` CI is green at `0e926c6f`

Closing this PR as landed. Thanks for the fix and patience while I folded it into the larger queue cleanup.

PR #2189, opened 2026-05-13
Add package filesystem evidence scanner

jesse-merhi:
## Summary

Adds a narrow pure scanner for package file contents that reports deterministic evidence for:

- raw Node filesystem module imports and calls
- OpenClaw fs-safe module imports and helper calls
- bounded evidence buckets sorted by file and line

This intentionally does not add DB tables, background jobs, HTTP endpoints, admin UI, or CLI commands. Those can layer on top of this scanner in follow-up PRs.

## Validation

- `bun run test convex/lib/packageFilesystemEvidenceScan.test.ts`
- `bun run format:check`
- `bun run lint`
- `git diff --cached --check`

### DIFF
--- a/convex/lib/packageFilesystemEvidenceScan.test.ts
+++ b/convex/lib/packageFilesystemEvidenceScan.test.ts
@@ -1,0 +1,103 @@
+/* @vitest-environment node */
+import { describe, expect, it } from "vitest";
+import {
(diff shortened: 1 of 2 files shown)


Patrick-Erichsen:
Nice, so this is to ensure plugin authors are using things like ` fs-safe` correctly?

PR #2213, opened 2026-05-14
Allow placeholder runtime ID correction on publish

### LINKED ISSUE #2133: Allow safe runtimeId correction for placeholder code-plugin packages
## Summary

Request: ClawHub should provide a safe way to correct or migrate a code-plugin package `runtimeId` when it was accidentally reserved/published with the wrong placeholder id.

Concrete affected package:

- Package: `@openviking/openclaw-plugin`
- Current runtime id: `openviking-openclaw-plugin-placeholder`
- Intended runtime id: `openviking`
- Source repo: `volcengine/OpenViking`
- Plugin source path: `examples/openclaw-plugin`

The package name was reserved with a placeholder runtime id, but the real OpenViking OpenClaw plugin uses `openviking` as its plugin/runtime id. Publishing the real plugin is now blocked because ClawHub treats `runtimeId` as immutable after the first publish.

## Current failure

Publishing the real plugin to `@openviking/openclaw-plugin` fails with:

```text
Package @openviking/openclaw-plugin already exists with plugin id openviking-openclaw-plugin-placeholder; runtime id changes are not allowed
```

## Why this matters

For OpenClaw plugins, the package name and the runtime/plugin id are related but not always created at the same time. During namespace/package-name reservation, a maintainer may publish a temporary placeholder package first, then later publish the real plugin artifact.

If the placeholder runtime id is wrong, the package name becomes permanently bound to an unusable runtime id. A normal soft delete does not appear to fully solve this because the package record still exists by normalized name and future publishing cannot claim the same name with the corrected runtime id.

## Proposal

Please consider supporting a guarded `runtimeId` correction/migration flow for code-plugin packages.

Possible safe constraints:

- Only package owner, publisher admin, moderator, or ClawHub admin can perform it.
- Allow it when the existing package is a placeholder/reservation, or when an admin explicitly approves the migration.
- Reject changes if the target runtime id is already claimed by another active package, unless the operation is an admin-supported package migration.
- Preserve audit logs for old runtime id, new runtime id, actor, package name, and reason.
- Optionally expose this through CLI/API/dashboard, for example:

```text
clawhub package runtime-id set @scope/name <new-runtime-id> --reason ...
```

or an admin-only migration command/API.

## Requested immediate fix for OpenViking

For this specific case, we need `@openviking/openclaw-plugin` to use runtime id `openviking` so the upstream GitHub Actions workflow in `volcengine/OpenViking` can publish future dev/latest releases to ClawHub and npm under the intended package name.

There is also an existing real package:

- `@openclaw/openviking`
- runtime id: `openviking`

So the ideal fix may require an admin-supported package rename/migration or runtime-id transfer, not just a blind edit.

## Related note

The npm package `@openviking/openclaw-plugin` can already be published from GitHub Actions. The current blocker is specifically the ClawHub package runtime id immutability/migration path.

LinQiang391:
﻿## Summary
- allow code-plugin publishes to correct an existing package-name-derived placeholder runtimeId only when publishing a latest release through trusted GitHub Actions or an admin publish
- keep ordinary owner user-publish runtimeId immutability and active runtimeId collision checks intact
- write an audit log when a placeholder runtimeId is corrected through publish

Fixes #2133.

## Security boundary
The exception is intentionally narrow:

- the existing runtimeId must exactly equal the deterministic placeholder derived from the normalized package name
- `@openviking/openclaw-plugin` -> `openviking-openclaw-plugin-placeholder`
- `demo-plugin` -> `demo-plugin-placeholder`
- the release must promote `latest`
- the publish must be from trusted GitHub Actions (`publishActor.kind === "github-actions"`) or an admin actor
- active runtimeId collisions are still rejected before the release is written

A normal owner user-publish and a placeholder-like runtimeId such as `some-other-placeholder` are both still rejected by the runtimeId immutability error.

## Real behavior proof
Targeted publish-path tests after the fix:

```text
RUN  v4.1.6 G:/gitcode/clawhub/clawhub

Test Files  1 passed (1)
Tests  6 passed | 121 skipped (127)
Start at  14:19:16
Duration  1.29s (transform 315ms, setup 30ms, import 937ms, tests 11ms, environment 0ms)
```

Full package public test file after the fix:

```text
RUN  v4.1.6 G:/gitcode/clawhub/clawhub

Test Files  1 passed (1)
Tests  127 passed (127)
Start at  14:20:52
Duration  1.18s (transform 316ms, setup 32ms, import 941ms, tests 81ms, environment 0ms)
```

Static checks:

```text
npx -y bun run format:check
Checking formatting...
All matched files use the correct format.

npx -y bun run lint:oxlint
Found 0 warnings and 0 errors.

npx -y bun x tsc --noEmit
# exited 0
```

## Tests
- npx -y bun x vitest run convex/packages.public.test.ts --testNamePattern "runtime id|plugin id collisions"
- npx -y bun run format:check
- npx -y bun run lint:oxlint
- npx -y bun x tsc --noEmit
- npx -y bun x vitest run convex/packages.public.test.ts

### DIFF
--- a/convex/packages.public.test.ts
+++ b/convex/packages.public.test.ts
@@ -157,6 +157,16 @@ const insertReleaseInternalHandler = (
       verification?: unknown;
       staticScan?: unknown;
       artifactKind?: "legacy-zip" | "npm-pack";
+      publishActor?:
+        | { kind: "user"; userId: string }
+        | {
+            kind: "github-actions";
+            repository: string;
+            workflow: string;
+            runId: string;
+            runAttempt: string;
+            sha: string;
+          };
       clawpackStorageId?: string;
       clawpackSha256?: string;
       clawpackSize?: number;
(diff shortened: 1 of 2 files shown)


LinQiang391:
Addressed the ClawSweeper review feedback:

- narrowed the exception from any placeholder-like runtimeId to only the deterministic package-name-derived placeholder, e.g. `@openviking/openclaw-plugin` -> `openviking-openclaw-plugin-placeholder`
- added a regression test proving `some-other-placeholder` is still rejected by runtimeId immutability
- updated the PR body with actual local test/static-check output

@clawsweeper re-review

Patrick-Erichsen:
Closing this out as I don't feel it's the right approach, it's too implicit.

PR #2279, opened 2026-05-16
Recommend homepage search topics

fuller-stack-dev:
## Summary
- Replace the hard-coded homepage search chips with materialized recommendation topics.
- Record submitted search terms through a filtered, deduped, rate-limited UI telemetry endpoint.
- Add package daily stats and blend search, skill, and plugin activity into homepage recommendations.
- Add the visible homepage chip label: “What people are looking for”.

## Behavior
- Homepage chips now read from `recommendationTopics:listHomepageTopics` and fall back to the existing static suggestions when no materialized topics are available.
- Recommendation rebuilds score search demand, skill daily stats, and plugin/package daily stats on an hourly cron.
- Search telemetry now posts to `/api/v1/search/telemetry`, not the public `/api/v1/search` read endpoint.
- The telemetry endpoint rejects non-ClawHub browser origins before rate-limit consumption.
- Raw submitted search text can only boost curated recommendation topic buckets; it cannot mint arbitrary public homepage chips.
- Search telemetry rejects emails, URLs/domains, secrets, empty/noisy terms, and overlong inputs before writing aggregate daily counts.
- Telemetry retention now prunes both dedupe rows and aggregate rows, and schedules continuation when a prune batch fills.
- Clicking a chip still navigates to `/search?q=<topic>` and records that chip search submission best-effort.

## Tests
- `bunx vitest run convex/searchTelemetry.test.ts convex/recommendationTopics.test.ts convex/httpApiV1/searchTelemetryV1.test.ts`
- `bunx vitest run convex/packages.stats.test.ts convex/recommendationTopics.test.ts convex/searchTelemetry.test.ts convex/httpApiV1/searchTelemetryV1.test.ts src/__tests__/home-route.test.tsx src/__tests__/search-route.test.tsx src/__tests__/header.test.tsx`
- `bun run ci:static`
- `bun run ci:unit`
- `bun run ci:types-build`
- `git diff --check`

## Backend Proof
- Full proof transcript: https://raw.githubusercontent.com/fuller-stack-dev/clawhub/qa-artifacts/clawhub-ui-proof/pr-2279/backend-recommendation-proof-20260516-005916/report.md
- Raw proof files: https://github.com/fuller-stack-dev/clawhub/tree/qa-artifacts/clawhub-ui-proof/pr-2279/backend-recommendation-proof-20260516-005916
- `bunx convex dev --once --typecheck=disable` completed with `Convex functions ready`.
- `POST /api/v1/search/telemetry` with `Origin: http://localhost:3000` returned `202` and body `{"ok":true}`.
- `POST /api/v1/search/telemetry` with `Origin: https://spam.example` returned `403` and `Search telemetry source not allowed`.
- Old `POST /api/v1/search` returned `404`, proving the telemetry sink is no longer mounted on the public read search endpoint.
- The allowed telemetry POST was visible in today's aggregate stats via `internal.recommendationTopics.getSearchStatsForDay`.
- Seeded 12 distinct curated buckets and 12 distinct uncurated buckets through `recordSearchInternal`, the same aggregate mutation called by the HTTP handler.
- `internal.recommendationTopics.rebuildHomepageRecommendationTopicsAction({ limit: 8 })` returned `{ "ok": true, "count": 2 }`.
- `recommendationTopics:listHomepageTopics({ limit: 8 })` returned only curated topics: `github integration` and `security scanner`; the high-volume raw `cheap seo backlinks ...` query did not surface.
- Retention proof seeded two expired rows, ran prune with `batchSize: 1`, observed `{ "dedupeDeleted": 1, "statsDeleted": 1, "deleted": 2 }`, then observed `{ "deleted": 0 }` after scheduled continuation.

## UI Proof
- Local proof was run against `http://localhost:3000/` with seeded recommendation topics.
- Proof artifacts are published in the PR comment: https://github.com/openclaw/clawhub/pull/2279#issuecomment-4465456422
- Verified the homepage label, the seeded chip texts, and click-through from `mcp tools` to `/search?q=mcp+tools`.


fuller-stack-dev:
## ClawHub UI Proof

Status: `pass`
Mode: `feature`
Scenario: `local Playwright behavior proof`
Provider: `local-playwright-chrome`
Baseline: not run for feature proof.
Candidate: `codex/dynamic-search-recommendations`

### Candidate Screenshots

**Homepage shows labeled recommendation chips**

Homepage shows labeled recommendation chips

**Clicking a recommendation opens matching search**

Clicking a recommendation opens matching search

Verified: homepage label `What people are looking for`, seeded topics `github integration`, `security scanner`, `mcp tools`, `dashboard builder`, and click-through from `mcp tools` to `/search?q=mcp+tools`.

Raw proof files: https://github.com/fuller-stack-dev/clawhub/tree/qa-artifacts/clawhub-ui-proof/pr-2279/local-recommendation-chips-20260515-214931

vyctorbrzezowski (commented):
Reviewed alongside @clawsweeper. Confirming one of its findings and 
adding others it didn't cover.

### Confirming Codex P2 (dedupe prune loop)

Agreed with Codex's finding verbatim:

> The new retention job only deletes the first batchSize expired dedupe 
> documents and then returns. Since the cron runs once per day with 
> batchSize: 500, any day with more than 500 expired rows leaves the 
> rest past the intended 14-day retention window, and the backlog can 
> grow indefinitely.

Source inspection matches: `convex/searchTelemetry.ts:179` deletes one 
batch and returns; `convex/crons.ts:50` invokes it once per 24h with 
batchSize 500. Fix should either loop until the batch comes back 
non-full, or self-schedule a continuation (same pattern already used 
in `convex/packages.ts:2434` for package stat events).

---

### New finding [P1?]: Public endpoint allows farming homepage chips

The Codex review flagged telemetry retention but didn't catch the 
content-integrity vector:

- `convex/http.ts:69` registers a **public** `POST /api/v1/search`
- `convex/lib/httpHeaders.ts:16` sets CORS `*`
- `convex/recommendationTopics.ts:98` promotes raw search terms to 
  homepage chips at a fixed threshold of 10 with a static weight

An external site with modest traffic can farm buckets/IPs and 
materialize arbitrary chips that pass the filter, turning the 
homepage into an SEO-spam / malicious-content surface.

Suggested mitigations (pick at least one):
- Origin/Referer allowlist on the POST endpoint
- Higher threshold + diversity requirement (distinct clients/days)
- Explicit topic allowlist or maintainer approval before chips 
  surface on the homepage

I'd argue this is closer to P1 than P2 — homepage integrity is a 
trust boundary.

---

### New finding [P3]: No retention on `searchQueryDailyStats`

Extends Codex's privacy concern. The new prune only covers 
`searchQueryDailyDedupe`. `convex/searchTelemetry.ts:115` writes 
`searchQueryDailyStats` with raw user input, and only a 7-day window 
is read — but rows are never deleted. Either add explicit retention 
for stats, or document why aggregates persist indefinitely.


------------------

### Nit: `POST /api/v1/search` path overloads a public read endpoint

The PR description frames the POST as best-effort UI telemetry, which 
is fine — but reusing the `/api/v1/search` path (which currently serves 
GET search results, documented in `openapi.json:2991`) couples a 
public read API with an internal telemetry sink under the same URL. 

Consider `/api/v1/search/telemetry`, `/api/internal/search-events`, 
or similar — makes it harder to accidentally treat the POST as part 
of the public v1 contract, and avoids future OpenAPI drift if the 
GET endpoint evolves.

Non-blocking.

fuller-stack-dev:
Addressed the review comments in `bfcf8bcf`:

- Retention now prunes both `searchQueryDailyDedupe` and `searchQueryDailyStats`, and schedules a continuation when either table fills the prune batch.
- Search telemetry moved from `POST /api/v1/search` to `POST /api/v1/search/telemetry` so it is no longer coupled to the public read search endpoint.
- The telemetry handler rejects non-ClawHub browser origins before rate-limit consumption.
- Submitted search text can only boost curated recommendation topic buckets now; high-volume unmatched raw terms cannot become homepage chips.
- Added backend proof to the PR body: allowed origin returns `202`, foreign origin returns `403`, and a rebuild surfaced only `github integration` after seeding both curated and uncurated high-volume telemetry.

Local verification passed: focused Vitest, `ci:static`, `ci:unit`, `ci:types-build`, and `git diff --check`.

fuller-stack-dev:
@clawsweeper re-review

Added fresh backend behavior proof for the exact areas requested:

- Full proof transcript: https://raw.githubusercontent.com/fuller-stack-dev/clawhub/qa-artifacts/clawhub-ui-proof/pr-2279/backend-recommendation-proof-20260516-005916/report.md
- Raw proof files: https://github.com/fuller-stack-dev/clawhub/tree/qa-artifacts/clawhub-ui-proof/pr-2279/backend-recommendation-proof-20260516-005916

What it proves:
- `POST /api/v1/search/telemetry` reaches Convex and returns `202` for `Origin: http://localhost:3000`.
- The same endpoint returns `403` for `Origin: https://spam.example`.
- Old `POST /api/v1/search` now returns `404`, so telemetry is no longer mounted on the public read endpoint.
- The allowed telemetry POST created a visible daily aggregate stats row.
- Rebuild ran live and materialized only curated homepage topics (`github integration`, `security scanner`), while a high-volume raw `cheap seo backlinks ...` query did not surface.
- Retention prune deleted both dedupe and aggregate rows, then scheduled continuation drained the remaining expired dedupe row before the second prune.

Patrick-Erichsen:
Closing this out. I think we should instead be focusing homepage discover on 1) download counts, 2) popular creators

PR #2284, opened 2026-05-16
fix: abort stale unban skill restore batches

vyctorbrzezowski:
## Summary

When a user is unbanned, ClawHub restores their hidden skills in background steps. This PR makes each step check whether the user was banned again before restoring anything. The latest ban always wins.

## What changed

- Unban skill-restore pages abort when the user has already been banned/deactivated again.
- Later ban markers are monotonic; older delayed ban pages cannot roll `softDeletedAt` back from a newer ban timestamp.
- Removed rows are not retimestamped or restored.

## Public behavior

Unban restore is fail-closed. If a restore page is stale because the user was banned or deactivated again, the page aborts instead of restoring old state.

Ban-hidden rows keep the newer ban marker. Removed rows remain removed.

## Behavior proof

Live Convex runtime proof from a stale unban restore page after a later ban:

```text
$ bunx convex run --push --typecheck=disable --codegen disable proof2284:run
- Preparing Convex functions...

✔ Convex functions ready!
{
  "restoreResult": {
    "aborted": true,
    "ok": true,
    "restoredCount": 0,
    "scheduled": false
  },
  "after": {
    "moderationReason": "user.banned",
    "moderationStatus": "hidden",
    "softDeletedAt": "<later-ban-timestamp>"
  },
  "fixture": {
    "bannedAt": "<later-ban-timestamp>",
    "ownerUserId": "<redacted>",
    "skillId": "<redacted>"
  }
}
```

This proof creates a stale restore page, applies a later ban before the page runs, then confirms the restore aborts and the skill remains hidden with the later ban marker.

Focused regression suite:

```text
$ bun run test convex/skills.banBatches.test.ts
Test Files  1 passed (1)
Tests       6 passed (6)
```

## Validation

```text
$ bun run ci:unit
Test Files  205 passed (205)
Tests       1991 passed (1991)
Statements  85.98%
Branches    75.17%
Functions   86.37%
Lines        89.48%
```

```text
$ bun run ci:types-build
tsc, package typechecks, clawhub-mod typecheck, Vite build, and Nitro build passed.
```

Current GitHub CI for this head also has `packages`, `types-build`, `e2e-http`, `playwright-smoke`, and `playwright-local-auth` passing. The `static` job currently stops at `bun audit` on the existing transitive `ws` advisory `GHSA-58qx-3vcg-4xpx`.


vyctorbrzezowski:
@clawsweeper re-review

vyctorbrzezowski:
@clawsweeper re-review

PR #2287, opened 2026-05-16
fix(api): guard moderated skill files and tags

vyctorbrzezowski:
## Summary

Makes the public skill APIs consistently refuse files, versions, and tags that should not be public. Hidden, deleted, cross-skill, or stale data now returns a blocked/not found response instead of leaking old files.

## What changed

- Public skill file readers now share the same moderation block semantics as ZIP downloads.
- Skill tags must point at a non-deleted version that belongs to the same skill.
- Digest-backed public lists preserve legacy markerless latest-version rows, but fail closed when an explicit marker points to another skill.
- `getBySlug` suppresses `latestVersion` when the latest row is soft-deleted or belongs to another skill.
- `resolveVersionByHash` suppresses stale, cross-skill, or soft-deleted `latestVersion` rows instead of serializing them.
- The digest/backfill validation model stays intact; expensive validation is still kept out of hot public list paths.

## Public behavior

Invalid latest-version state becomes `null`/absent in public detail and resolver responses. It is never returned as a serialized stale version object.

Moderated file and package compatibility readers fail closed. Cross-skill tags and soft-deleted versions are rejected instead of being treated as public artifacts.

Legacy digest rows without the new owner marker keep their latest version for compatibility; rows with an explicit cross-skill owner marker are suppressed.

## Behavior proof

Live Convex runtime proof for the endpoint paths ClawSweeper requested on rebased head `3ff605dd`:

```text
$ bunx convex run --push --typecheck=disable --codegen disable proof2287:run
✔ Convex functions ready!
{
  "fixture": {
    "activeSlug": "proof2287-vjc9cp6b-active",
    "moderatedSlug": "proof2287-vjc9cp6b-moderated"
  },
  "results": [
    {
      "body": "Blocked: this skill has been flagged as malicious by ClawScan and cannot be downloaded.",
      "label": "raw-file moderated skill",
      "status": 403
    },
    {
      "body": "Blocked: this skill has been flagged as malicious by ClawScan and cannot be downloaded.",
      "label": "package compatibility file for moderated skill",
      "status": 403
    },
    {
      "body": "Version not found",
      "label": "raw-file cross-skill tag",
      "status": 404
    },
    {
      "body": "Version not available",
      "label": "raw-file soft-deleted tag",
      "status": 410
    },
    {
      "body": "Version not found",
      "label": "package compatibility cross-skill tag",
      "status": 404
    },
    {
      "body": "Version not found",
      "label": "package compatibility soft-deleted tag",
      "status": 404
    }
  ]
}
```

After that proof, I removed the temporary `convex/proof2287.ts` helper and pushed the deployment again without it:

```text
$ bunx convex run --push --typecheck=disable --codegen disable publishers:listMine
✔ Convex functions ready!
[]
```

Live Convex runtime proof for the compatibility-sensitive digest path from the earlier head:

```text
$ bunx convex run --push --typecheck=disable --codegen disable proof2287:run
- Preparing Convex functions...

✔ Convex functions ready!
{
  "fixture": {
    "legacySlug": "proof2287-...-legacy",
    "staleSlug": "proof2287-...-stale"
  },
  "relevant": [
    {
      "latestVersion": "1.0.0",
      "slug": "proof2287-...-legacy"
    },
    {
      "latestVersion": null,
      "slug": "proof2287-...-stale"
    }
  ]
}
```

This creates one legacy markerless digest row and one explicit cross-skill/stale digest row. The legacy row still serializes `latestVersion`; the explicit stale row returns `null`, which addresses the compatibility concern without reopening the leak.

Focused regression suite on rebased head `3ff605dd`:

```text
$ bun run test convex/skills.public.test.ts convex/httpApiV1.handlers.test.ts convex/downloads.test.ts convex/skills.deleteTags.test.ts convex/skills.publicListCursor.test.ts convex/lib/skillSearchDigest.test.ts
Test Files  6 passed (6)
Tests       269 passed (269)
```

## Validation

```text
$ bun run ci:unit
Test Files  204 passed (204)
Tests       2003 passed (2003)
Statements  85.98%
Branches    75.17%
Functions   86.37%
Lines        89.48%
```

```text
$ bun run ci:types-build
tsc, package typechecks, clawhub-mod typecheck, Vite build, and Nitro build passed.
```


vyctorbrzezowski:
@clawsweeper re-review

vyctorbrzezowski:
@clawsweeper re-review

Issue #46, opened 2026-01-27
Missing undici dependency in package.json causes ERR_MODULE_NOT_FOUND on fresh install


webdevtodayjason:
## Description

After installing `clawdhub` globally via npm, the CLI fails immediately with a module resolution error for the `undici` package.

## Environment
- **Node.js version**: v20.20.0 (LTS "Iron")
- **npm version**: 10.8.2
- **ClawdHub version**: 0.3.0
- **OS**: macOS (arm64)

## Steps to Reproduce

1. Install ClawdHub globally:
   ```bash
   npm install -g clawdhub
   ```

2. Try to run any ClawdHub command:
   ```bash
   clawdhub --version
   ```

## Expected Behavior

ClawdHub should execute successfully and display its version or help information.

## Actual Behavior

The command fails with the following error:

```
node:internal/modules/esm/resolve:873
  throw new ERR_MODULE_NOT_FOUND(packageName, fileURLToPath(base), null);
        ^

Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'undici' imported from /Users/sem/.nvm/versions/node/v20.20.0/lib/node_modules/clawdhub/dist/http.js
    at packageResolve (node:internal/modules/esm/resolve:873:9)
    at moduleResolve (node:internal/modules/esm/resolve:946:18)
    at defaultResolve (node:internal/modules/esm/resolve:1188:11)
    at ModuleLoader.defaultResolve (node:internal/modules/esm/loader:708:12)
    at #cachedDefaultResolve (node:internal/modules/esm/loader:657:25)
    at ModuleLoader.resolve (node:internal/modules/esm/loader:640:38)
    at ModuleLoader.getModuleJobForImport (node:internal/modules/esm/loader:264:38)
    at ModuleJob._link (node:internal/modules/esm/module_job:168:49) {
  code: 'ERR_MODULE_NOT_FOUND'
}
```

## Root Cause

The `undici` package is used in `dist/http.js` but is not declared as a dependency in `package.json`. This means it's not installed when users run `npm install -g clawdhub`.

## Workaround

Manually install `undici` in the ClawdHub module directory:

```bash
cd ~/.nvm/versions/node/v20.20.0/lib/node_modules/clawdhub
npm install undici
```

After this, ClawdHub works as expected.

## Suggested Fix

Add `undici` to the `dependencies` section of `package.json`:

```json
{
  "dependencies": {
    "undici": "^6.0.0"
  }
}
```

## Additional Context

[...]

PR #2293, opened 2026-05-18
feat(search): expose per-component score breakdown opt-in

SFKislev:
## Why
Skill authors who want to understand why their skill ranks where it does currently have no observability. They can see the final `score` but not which component (vector similarity, lexical match, popularity, rankTier) is dominating it. This makes "my skill should rank higher for X" debugging a black box.

## What
Adds an opt-in `includeScoreBreakdown: boolean` arg to the `searchSkills` action and a `?includeScoreBreakdown=true` query param to `/api/v1/search`. When set, each result gains:

```ts
scoreBreakdown: {
  vectorScore: number;      // raw cosine, [-1, 1]
  lexicalBoost: number;     // EXACT_SLUG_BOOST | SLUG_TOKEN_BOOST | ... + name boost
  popularityBoost: number;  // log1p(downloads) * POPULARITY_WEIGHT
  rankTier: number;         // 0..3 from classifySkillMatch
}
```

`vectorScore + lexicalBoost + popularityBoost === score` exactly.

## Design notes
- **Opt-in, default-off.** Existing API consumers see no change; snapshot shape is preserved. The internal `searchSouls` handler is also updated for the shared `scoreSkillResult` signature change but unaffected externally.
- **No schema change.** This is response data, not stored state.
- **Tests:** new direct-unit test asserts the four components sum to `score`; new end-to-end test asserts the field is present iff opted in and absent by default.

## Why this is useful (concrete)
I hit this trying to understand why a skill that should rank for `news` sits at #46. With breakdown, an author could see immediately whether their `vectorScore` is the bottleneck (rewrite description), their `lexicalBoost` didn't fire (slug/name mismatch), or `rankTier` is 2/3 (matched only on summary). Right now this requires reading the convex source and back-solving from the aggregated score.

## Test plan
- [x] `bunx vitest run convex/search.test.ts` — 45/45 pass (44 existing + 2 new)
- [x] `bunx tsc --noEmit` on both `convex/tsconfig.json` and root `tsconfig.json` — clean
- [x] Full suite: only pre-existing failures on `main` (ui-proof + crabbox scripts), nothing in `convex/`
- [x] Verified default-off behavior preserves existing response shape

### DIFF
--- a/convex/httpApiV1/skillsV1.ts
+++ b/convex/httpApiV1/skillsV1.ts
@@ -37,6 +37,12 @@ import {
 
 type SearchSkillEntry = {
   score: number;
+  scoreBreakdown?: {
+    vectorScore: number;
+    lexicalBoost: number;
+    popularityBoost: number;
+    rankTier: number;
+  };
   skill: {
     slug?: string;
     displayName?: string;
@@ -425,6 +431,9 @@ export async function searchSkillsV1Handler(ctx: ActionCtx, request: Request) {
     url.searchParams.get("nonSuspiciousOnly"),
     url.searchParams.get("nonSuspicious"),
   );
+  const includeScoreBreakdown = parseBooleanQueryParam(
+    url.searchParams.get("includeScoreBreakdown"),
+  );
 
   if (!query) return json({ results: [] }, 200, rate.headers);
 
@@ -433,6 +442,7 @@ export async function searchSkillsV1Handler(ctx: ActionCtx, request: Request) {
     limit,
     highlightedOnly: highlightedOnly || undefined,
     nonSuspiciousOnly: nonSuspiciousOnly || undefined,
+    includeScoreBreakdown: includeScoreBreakdown || undefined,
   })) as SearchSkillEntry[];
 
   return json(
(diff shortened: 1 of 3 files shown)


SFKislev:
@Patrick-Erichsen hi, nudging this politely.

Patrick-Erichsen:
Hey @SFKislev , thanks for the bump.

I don't think we want to expose our ranking algo directly, mainly to prevent people gamifying the search.

If you have feedback on how we can improve the search rankings though please share however.

SFKislev:
hi @Patrick-Erichsen , I think it would just give SEO observability. You're already exposing how it works with serious detail, but it's difficult to understand if this is really working. I can recreate the behavior locally, but seeing a breakdown of the score woud tell developers if they should work on / optimize the slug, description, skill text etc. Currently, even though the system is described in detail, there's no observability as to whether it works.

Patrick-Erichsen:
> You're already exposing how it works with serious detail, but it's difficult to understand if this is really working.

This is in the docs? Mind linking?

But yeah I definitely hear you. We want folks to be able to understand the ranking algo, and it's also OSS so not that hard to just have an agent explain how it works. 

But I still worry about exposing the actual exact results. I can't think of any marketplace products that do that because it just leads to gamification.

PR #2310, opened 2026-05-19
Fix legacy personal skill rows on publisher profiles

fuller-stack-dev:
## Summary

- Include legacy `ownerUserId`-owned rows when loading a user-kind publisher profile catalog.
- Keep `ownerPublisherId` as the canonical public owner ID, but union/dedupe legacy personal rows where `ownerPublisherId` is absent.
- Filter out rows that already belong to a different publisher so org-owned content does not leak back onto a personal profile.

## Root Cause

`/user/:handle` loads `api.publishers.getProfileByHandle` and `api.publishers.listPublishedPage`. Both depended on `getPublisherPublishedRows()`, which only queried content by `ownerPublisherId`.

That misses legacy skills published before the publisher migration where the row has `ownerUserId` but no `ownerPublisherId`. Skill detail pages still worked because `api.skills.getBySlug` falls back from `ownerUserId` to the user's personal publisher, so individual skills resolved while the profile catalog did not.

## Behavioral Proof

Live production baseline before this fix, using `https://wry-manatee-359.convex.cloud`:

```json
{
  "currentProdProfileStats": {
    "downloads": 1253,
    "installs": 0,
    "packages": 13,
    "skills": 1,
    "stars": 0
  },
  "currentProdPublishedPage": {
    "count": 14,
    "isDone": true,
    "names": [
      "Email",
      "Clawic CLI",
      "Database",
      "Excel",
      "Tasks",
      "Deep Research",
      "WhatsApp",
      "Google",
      "Notes",
      "Drive",
      "Flights",
      "Real Estate",
      "Cloudflare",
      "Sentry"
    ]
  },
  "currentProdLegacyUserSkillCount": 955,
  "legacyUserPages": 10
}
```

Regression proof on this branch:

- `bunx vitest run convex/publishers.test.ts --testNamePattern "legacy user-owned"`
  - 1 passed, 21 skipped
  - Covers `listPublishedPage` returning legacy personal skill/plugin rows, deduping scoped rows, and rejecting rows already assigned to another publisher.
  - Covers `getProfileByHandle` stats and preview items including legacy personal rows.
- `bunx vitest run convex/publishers.test.ts`
  - 22 passed
- `bun run ci:unit`
  - 205 test files passed
  - 2022 tests passed
  - global coverage: statements 85.99%, branches 75.11%, functions 86.37%, lines 89.49%
- `bun run ci:static`
  - peer checks passed
  - `bun audit --ignore GHSA-rmmr-r34h-pfm5` passed
  - format check passed
  - oxlint passed
  - knip dead-code checks passed
- `bunx tsc --noEmit`
  - passed

No screenshot proof was generated because this is a Convex profile read fix; the failing and fixed behavior is at the query boundary used by the page.

### DIFF
--- a/convex/publishers.test.ts
+++ b/convex/publishers.test.ts
@@ -3,6 +3,7 @@ import { describe, expect, it, vi } from "vitest";
 import { assertCanManageOwnedResource } from "./lib/publishers";
 import {
   addMember,
+  getProfileByHandle,
   listPublicPage,
   listPublic,
   listMine,
@@ -103,6 +104,22 @@ const listPublishedPageHandler = (
       continueCursor: string;
       isDone: boolean;
     }
+  >
+)._handler;
+
+const getProfileByHandleHandler = (
+  getProfileByHandle as unknown as WrappedHandler<
+    { handle: string },
+    {
+      stats: {
+        skills: number;
+        packages: number;
+        downloads: number;
+        installs: number;
+        stars: number;
+      };
+      publishedItems: Array<{ displayName: string; downloads: number }>;
+    } | null
   >
 )._handler;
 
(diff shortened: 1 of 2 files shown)


Patrick-Erichsen:
I think the right solution here is to backfill publishers rather than fallback to owner id. Working on that rn.

PR #2323, opened 2026-05-19
Surface OAuth callback returns without auth code

h-mascot:
## Summary
- Track browser OAuth redirect attempts before GitHub sign-in starts.
- Detect the return-to-app state where Convex redirects back without a client auth code, URL error, or session.
- Show an actionable auth diagnostic instead of leaving users on the generic sign-in screen.

## Evidence
A real ClawHub CLI auth attempt returned from GitHub through `/api/auth/callback/github` with HTTP 302, but the response `Location` was only `https://clawhub.ai/cli/auth?...` with no `code=` or `error=`. That leaves the client unable to complete token creation or explain the failure.

## Tests
- `bunx vitest run src/components/AppProviders.test.tsx src/components/SignInButton.test.tsx src/__tests__/header.test.tsx`
- `bun run lint`
- `bunx tsc --noEmit`
- `VITE_CONVEX_URL=https://example.invalid bun run build`

### DIFF
--- a/src/components/AppProviders.test.tsx
+++ b/src/components/AppProviders.test.tsx
@@ -103,6 +103,35 @@ describe("AuthCodeHandler", () => {
       expect(getAuthErrorSnapshot()).toBe(DELETED_SIGN_IN_MESSAGE);
     });
   });
+
+  it("does not consume initial CLI device codes as OAuth callback codes", async () => {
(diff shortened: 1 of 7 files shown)


Patrick-Erichsen:
Hey @h-mascot , this logic seems quite sensitive. Mind explaining the issue in more detail to me?

PR #2391, opened 2026-05-24
fix: allow package publishes when matching skill slug is owned

### LINKED ISSUE #2390: Package publish blocked when an owner also has a skill with the same slug
## Summary

Publishing a new package version can be blocked by a skill slug owned by the same account, even when the package already exists under that package name.

## What I tried

I have both a skill and a package with the slug/name `openclaw-delx-plugin` under my account. The package is already present in ClawHub as a `code-plugin` and `clawhub package inspect openclaw-delx-plugin --json` reports latest package version `0.2.0`.

A dry run for publishing package version `0.2.1` succeeds:

```bash
clawhub package publish . \
  --family code-plugin \
  --name openclaw-delx-plugin \
  --display-name "Delx Witness Protocol for OpenClaw" \
  --version 0.2.1 \
  --source-repo davidmosiah/openclaw-delx-plugin \
  --source-ref main \
  --source-commit 05ccf08796f317b135ae54ec2bfcaec306422f02 \
  --dry-run --json
```

The real publish fails:

```text
Error: Package name collides with existing skill slug "openclaw-delx-plugin"
    at publishPackageImpl (../convex/packages.ts:4682:6)
```

## Expected behavior

If a package already exists and the authenticated owner is publishing a new version for that same package, the same-owner skill slug should not block the package version publish. Alternatively, the CLI/server should provide an actionable migration path for moving the skill slug away from the package name.

## Actual behavior

The package version publish is rejected because a skill slug with the same string exists, even though the package already exists too.

## Rename attempt

I tried to rename the skill away from the package name:

```bash
clawhub skill rename openclaw-delx-plugin delx-witness-openclaw-skill --yes
```

That failed with:

```text
Error: newSlug required
```

This is surprising because the CLI command shape appears to accept `<slug> <new-slug>`.

## Impact

This blocks package authors from publishing upgrades when an older explanatory skill used the same slug as the newer package/plugin. It also makes it harder to keep the package as the canonical install surface.

## Environment

- ClawHub CLI: `0.12.3`
- Package family: `code-plugin`
- Existing package: `openclaw-delx-plugin`
- Existing skill slug: `openclaw-delx-plugin`

davidmosiah:
## Summary

- Allow package publishes to proceed when an existing skill with the same slug is owned by the same user/publisher.
- Keep the existing collision block when the matching skill slug belongs to a different owner.
- Add regression coverage for both same-owner and different-owner package publish paths.

Fixes #2390.

## Tests

- `bunx vitest run convex/packages.public.test.ts --testNamePattern "allows package publishes when the matching skill slug belongs to the same owner" --reporter verbose` (red before implementation, green after)
- `bunx vitest run convex/packages.public.test.ts --testNamePattern "matching skill slug" --reporter verbose`
- `bunx vitest run convex/packages.public.test.ts --reporter dot`
- `bunx oxfmt --check convex/packages.ts convex/packages.public.test.ts`
- `git diff --check`
- `bunx oxlint --type-aware --tsconfig ./tsconfig.oxlint.json convex/packages.ts convex/packages.public.test.ts`

## Notes

Full repo `bun run format:check` currently reports an unrelated formatting issue in `convex/lib/securityPrompt.test.ts`, and full `bun run lint` reports unrelated `no-unnecessary-type-conversion` errors in `convex/securityScan.test.ts`. I left those files untouched.

### DIFF
--- a/convex/packages.public.test.ts
+++ b/convex/packages.public.test.ts
@@ -3486,6 +3486,94 @@ describe("packages public queries", () => {
     expect(ctx.runMutation).not.toHaveBeenCalled();
   });
 
+  it("allows package publishes when the matching skill slug belongs to the same owner", async () => {
+    const ctx = {
+      runQuery: vi
+        .fn()
+        .mockResolvedValueOnce(null)
+        .mockResolvedValueOnce({
+          _id: "users:owner",
+          githubCreatedAt: Date.now() - 20 * 24 * 60 * 60 * 1000,
+        })
+        .mockResolvedValueOnce({
+          _id: "users:owner",
+          role: "user",
+          githubCreatedAt: Date.now() - 20 * 24 * 60 * 60 * 1000,
+        })
+        .mockResolvedValueOnce({
+          _id: "skills:demo",
+          slug: "demo-plugin",
+          ownerUserId: "users:owner",
+          ownerPublisherId: "publishers:owner",
+        }),
+      runMutation: vi.fn(async () => ({
+        publisherId: "publishers:owner",
+        linkedUserId: "users:owner",
+      })),
+      scheduler: { runAfter: vi.fn() },
+      storage: { get: vi.fn() },
+    };
+
+    await expect(
+      publishPackageForUserInternalHandler(ctx as never, {
+        actorUserId: "users:owner",
+        payload: {
+          name: "demo-plugin",
+          displayName: "Demo Plugin",
+          family: "bundle-plugin",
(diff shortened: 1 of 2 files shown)


Patrick-Erichsen:
Hey @davidmosiah thanks for bringing this issue up. I'm going to close this PR since I don't think it's the right shaped fix for the issue, but let's continue the conversation in https://github.com/openclaw/clawhub/issues/2390#issuecomment-4557296870

PR #2414, opened 2026-05-27
Require multipart package publishes

jesse-merhi:
## Summary

- Require `multipart/form-data` for `POST /api/v1/packages`; JSON package publish bodies now fail before publish actions run.
- Keep public package publish payloads metadata-only: caller-supplied `payload.files`, `payload.artifact`, multipart `artifact`, `files[]`, `tarball`, and other unsupported fields are rejected.
- Preserve the released multipart API shapes the CLI uses: `payload` + repeated `files` uploads, or `payload` + one `clawpack` `.tgz`; `files` and `clawpack` cannot be mixed.
- Derive stored file metadata and npm-pack artifact metadata from uploaded bytes server-side, then pass only that server-built request to internal publish actions.
- Preserve large ClawPack publishes by staging tarballs over the direct multipart budget through upload-url + upload-ticket flow.
- Bind staged ClawPack storage ids to the auth context that requested the upload URL, with expiry, storage creation-time checks, and retry-safe reuse for the same storage id.
- Document the package publish boundary in docs/specs.

## Tests

- `bunx vitest run convex/uploads.test.ts --reporter=dot`
- `bunx vitest run convex/httpApi.handlers.test.ts convex/httpApiV1.handlers.test.ts --testNamePattern "package publish|multipart package publish|multipart ClawPack|staged ClawPack|browser session auth|cliUploadUrl|upload tickets" --reporter=dot`
- `bun run --cwd packages/clawhub test:src -- src/cli/commands/packages.test.ts --testNamePattern "ClawPack tarballs over|stages ClawPack|publishes a ClawPack|cleans generated ClawPack"`
- `bun run format:check`
- `bun run lint`
- `bun run ci:unit`
- `git diff --check origin/main`

## Not Run

- `bunx convex codegen` locally; this checkout has no `CONVEX_DEPLOYMENT` configured.


Patrick-Erichsen:
Maintainer note after tracing the publish paths: I think the GHSA-shaped fix can land without breaking existing publisher CI.

The important security invariant is that public clients must not be able to provide trusted `artifact` metadata. The server should derive artifact metadata from the uploaded bytes.

As-is, this PR also changes the multipart field contract: current released CLI clients send `clawpack` for tarballs and `files` for extracted uploads, while this branch only accepts `tarball` and `files[]` and rejects the old names. A backend deploy before all publishers upgrade would likely break existing `clawhub package publish` workflows.

Recommended narrower path:
- Keep accepting the existing `clawpack` / `files` fields.
- Move the web UI publish flow to multipart if desired.
- Reject caller-supplied `artifact` metadata on all public publish paths.
- Preserve current CLI compatibility while the server remains the source of truth for artifact hashes/integrity.


---

Clanker comment above, but in my own words, I think this will break existing CLI clients because of the form field name change, but I'm not sure that's actually required? I believe if we just left the field names as-is we could still address the GHSA without breaking CLI clients.

jesse-merhi:
@clawsweeper re-review

PR #2440, opened 2026-05-29
fix: surface organization creation errors

RomneyDa:
## Summary
- Surface create-organization mutation failures in the settings dialog instead of silently dropping them.
- Disable the create button while the mutation is pending and clear stale errors when editing/reopening the dialog.
- Add a settings regression test for duplicate-handle Convex errors.
- Apply a formatter-only import reorder in `convex/packages.public.test.ts` so the full static gate passes.

## Tests
- `VITE_CONVEX_URL=https://example.invalid bunx vitest run src/routes/-settings.test.tsx`
- `bun run ci:static`
- `bun run ci:unit`
- `AUTOREVIEW_AUTO_TESTS=0 .agents/skills/autoreview/scripts/autoreview --mode local`

### DIFF
--- a/convex/packages.public.test.ts
+++ b/convex/packages.public.test.ts
@@ -2,6 +2,7 @@
 
 import { getAuthUserId } from "@convex-dev/auth/server";
 import { afterEach, describe, expect, it, vi } from "vitest";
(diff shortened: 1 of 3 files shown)


Patrick-Erichsen:
Nice cleanup, much appreciated!

PR #2492, opened 2026-06-04
fix: deprecate legacy artifact LLM eval

vyctorbrzezowski:
## Summary

This PR removes the old hosted LLM artifact evaluator from paths that can be mistaken for authoritative ClawScan results.

## Scenario this protects

The old evaluator builds a bounded prompt from artifact text. If a long skill or package has important behavior outside that prompt excerpt, an LLM answer from that path should not be able to write a clean public verdict.

This PR keeps that path advisory only and routes authoritative artifact scanning through the Codex-backed worker.

## What changed

- Old manual and backfill helpers now enqueue the Codex-backed `securityScanJobs` worker.
- Direct legacy evaluator calls return advisory diagnostics only.
- Legacy evaluator calls no longer patch `llmAnalysis`, moderation state, package trust, or public scan state.
- The local dry-run script labels hosted LLM output as advisory and points users to `clawhub scan` for authoritative coverage.
- The moderation spec records the invariant so future changes do not accidentally restore authoritative writes from this path.

## What maintainers should expect

Authoritative skill and package verdicts stay on the worker path that materializes the artifact workspace and writes through `securityScan.completeCodexScanJob`.

The hosted evaluator can still be useful for diagnostics, but it no longer changes trust state.

## What this intentionally does not change

- API-key-required evaluation stays unchanged.
- Comment scam evaluation stays unchanged.
- Static scan aggregation and worker schemas stay unchanged.
- The public `clawhub scan` command remains on the canonical scan API.

## Validation

- `bunx vitest run convex/llmEval.test.ts scripts/local-clawscan-dry-run.test.ts`
- `bun test packages/clawhub/src/cli/commands/scan.test.ts`
- `bun run ci:static`
- `VITE_CONVEX_URL=https://example.invalid bun run coverage`
- `bun run ci:types-build`

## Runtime proof (June 15, 2026)

This was exercised against an isolated local Convex deployment using the PR branch and repository-provided synthetic moderation fixtures. No production data or production credentials were used. The worker execution used the real Codex CLI with `gpt-5.5`; identifiers and local paths are redacted below.

### Backfill routes through `securityScanJobs`

The PR schema and functions deployed successfully to a fresh local Convex runtime. A one-item preserve-mode backfill then scheduled the fixture through the worker queue:

```text
[llmEval:backfill] Processing batch of 1 skills
(cursor=0, accumulated=0, moderationMode=preserve, dryRun=false)
[llmEval:backfill] Scheduled eval for local-agentic-risk-demo
```

```json
[...]

PR #2489, opened 2026-06-04
Require Codex scan artifact coverage proof

vyctorbrzezowski:
## Summary

This PR makes the Codex scan worker prove artifact coverage before a benign result can be stored as a clean ClawScan result.

## Scenario this protects

If a submitted artifact has several files, or one large file, the worker result needs to say what it actually examined. Without that proof, a benign answer can be ambiguous: it may mean the artifact looked fine, or it may mean the worker only inspected the easiest part of it.

The important case is simple: if meaningful content is omitted, or only the head of a large file is considered, ClawHub should not store that result as clean.

## What changed

- Requires `artifact_coverage` in the Codex worker output schema.
- Updates the worker prompt so benign verdicts must include submitted-file coverage.
- Requires full coverage for small files.
- Requires both head and tail coverage for large files.
- Rejects benign results before storage when coverage proof is missing, incomplete, or missing large-file tail coverage.
- Documents the worker-time coverage invariant in `specs/security-moderation.md`.

## What maintainers should expect

This stays at the worker/schema boundary. It does not add a public API field or change the product surface. The stored trust decision gets stricter, while raw worker diagnostics still keep the full result for operator debugging.

Suspicious or malicious worker results can still be useful with incomplete coverage. Only benign results need complete enough coverage to become clean.

## Validation

- `bun run test scripts/security/run-codex-scan-worker.test.ts`
- `bun run test scripts/security/run-codex-scan-worker.test.ts convex/securityScan.test.ts convex/lib/securityPrompt.test.ts`
- `bunx tsc --noEmit --pretty false`
- `bun run ci:static`

## Runtime proof (June 15, 2026)

This was exercised against an isolated local Convex deployment using only a synthetic skill artifact. No production data or production credentials were used. The positive case used the real Codex CLI with `gpt-5.5`; identifiers and local paths are redacted below.

### Real Codex worker: complete benign coverage is accepted

The synthetic artifact contained one 278-byte `SKILL.md`. The real worker claimed the queued scan and completed it cleanly:

```text
diagnostics directory: [redacted]
claimed 1 job(s)
completed [redacted-job-id]: clean
worker summary: claimed=1 completed=1 failed=0 elapsedMs=37342
```

The redacted model result contained exact full-file coverage:

```json
{
  "verdict": "benign",
  "confidence": "high",
  "artifact_coverage": {
    "status": "complete",
    "inspected_files": [
      {
        "path": "SKILL.md",
        "coverage": "full",
        "ranges": [
          {
[...]

PR #2493, opened 2026-06-04
fix: floor incomplete ClawScan coverage

vyctorbrzezowski:
## What this PR does

This PR makes the prompt-backed ClawScan evaluator fail closed when ClawHub did not review the full material artifact text.

## Scenario this protects

The legacy/local evaluator builds a prompt from `SKILL.md` and uploaded file contents. That prompt path has size limits.

If `SKILL.md` is truncated, file contents are truncated, or later file blocks are omitted after the aggregate prompt limit, a benign model answer should not be stored as clean. In that situation, there is still unreviewed artifact text that may matter.

## What changed

- Adds an explicit artifact coverage signal for the prompt-backed evaluator path.
- Records incomplete coverage when prompt assembly truncates `SKILL.md`.
- Records incomplete coverage when prompt assembly truncates file content.
- Records incomplete coverage when file blocks are omitted after the aggregate prompt limit.
- Records incomplete coverage when hidden comments are stripped before review.
- Converts benign answers from incomplete coverage into suspicious/review results before storage.
- Keeps incomplete coverage visible in moderation with `review.artifact_incomplete`.

## What maintainers should expect

A clean-looking result from incomplete prompt coverage can no longer silently clear review state. The caller gets a review-floor result instead, with the reason preserved.

The modern GitHub Actions Codex worker path is different: it materializes artifact files into a workspace for Codex instead of relying on the sliced prompt text. This PR still preserves parsed `artifactCoverage` if that parser path receives it.

SkillSpector and static scan findings remain advisory/context inputs. They do not replace the ClawScan verdict, so the review floor belongs at the LLM result/moderation boundary.

## Validation

- `bun test convex/lib/securityPrompt.test.ts scripts/local-clawscan-dry-run.test.ts convex/lib/moderationEngine.test.ts`
- `bunx vitest run convex/securityScan.test.ts convex/llmEval.test.ts scripts/security/run-codex-scan-worker.test.ts`
- `bunx convex codegen`
- `bun run ci:static`
- `bun run ci:unit`
- `bun run ci:types-build`
- `bun run ci:packages`

`bun test convex/llmEval.test.ts convex/securityScan.test.ts scripts/security/run-codex-scan-worker.test.ts` was also tried, but `convex/securityScan.test.ts` requires Vitest APIs such as `vi.stubEnv` and `vi.mocked`; the same files passed under Vitest as listed above.


vyctorbrzezowski:
Updated in bf890c42.

After-fix validation:
- `bun test convex/lib/moderationEngine.test.ts`
- `bun run format:check -- convex/lib/moderationEngine.test.ts`

@clawsweeper re-review

vyctorbrzezowski:
@clawsweeper re-review

vyctorbrzezowski:
Closing this in favor of consolidating the scanner policy around #2492 and #2489. If #2492 is accepted, the legacy prompt-backed evaluator becomes advisory-only, so this branch’s authoritative coverage floor is no longer the right trust boundary.

PR #2495, opened 2026-06-04
feat: add exact skill version revocation

vyctorbrzezowski:
## Summary

This PR proposes the ClawHub side of a coordinated exact-version revocation flow with OpenClaw.

The proposed contract is:

- ClawHub moderators can mark one exact skill version as `revoked` or restore it to `active`.
- ClawHub keeps the version record and audit evidence, instead of hiding or deleting the whole skill.
- ClawHub blocks future public artifact delivery for that exact version.
- ClawHub exposes the revocation state in the existing machine-readable bulk security-verdict endpoint.
- OpenClaw consumes that explicit verdict in the companion PR and quarantines only installed skills whose exact version, origin, lock entry, and path all still match.

Companion PR: https://github.com/openclaw/openclaw/pull/90201

## Why This Is Split Across Two Repos

This change needs both halves to be useful and safe:

- ClawHub is the source of truth for exact skill-version revocation and artifact blocking.
- OpenClaw is the installed-client side that can move an already-installed revoked version out of active skill loading.

The deployment order should be ClawHub first, then OpenClaw. The OpenClaw PR is intentionally defensive: if the new `revocation` field is missing, malformed, unavailable, or inconsistent, it leaves installed skills untouched.

## Proposed ClawHub Behavior

- Add optional `skillVersions.manualRevocation` metadata.
- Add a moderator API path:
  `POST /api/v1/skills/{slug}/versions/{version}/revocation`
- Add a `clawhub-mod skills moderate-version` command for setting `active` or `revoked` with an audit reason.
- Block revoked exact versions from ZIP download, raw-file reads, Skill Card delivery, export, and legacy package compatibility raw-file delivery.
- Keep public API compatibility by extending `POST /api/v1/skills/-/security-verdicts` additively with:
  `revocation: { revoked: boolean, revokedAt: number | null }`
- Return a failed exact-version verdict with `reason: "version.revoked"` only when that exact version is revoked.
- Document the moderator endpoint, the security-verdict revocation field, and the `clawhub-mod skills moderate-version` command.

## Safety Properties

- Revocation is exact-version scoped; other versions of the same skill remain available.
- The version record is preserved for investigation and audit.
- Lookup failures are not revocation signals.
- Public API response changes are additive.
- Existing skill-version rows remain compatible because the new field is optional.

## Validation

- `bun run ci:pr`
- `bunx vitest run convex/downloads.test.ts convex/httpApiV1.handlers.test.ts convex/skills.versionRevocation.test.ts packages/clawhub-mod/src/commands/moderation.test.ts`
- `bun run --cwd packages/schema build`
- `bunx tsc --noEmit`
- `bun run --cwd packages/clawhub-mod typecheck`
- `bunx vitest run convex/httpApiV1.handlers.test.ts packages/clawhub-mod/src/commands/moderation.test.ts`
- `bun run format:check -- docs/http-api.md packages/clawhub-mod/README.md CHANGELOG.md`
- `git diff --check`

Goal fechado: 1,157,217 tokens, ~87m54s.


vyctorbrzezowski:
@clawsweeper re-review

Patrick-Erichsen:
Closing after our convo - if we deem a skill to be malicious than we want to just completely ban that user.

PR #2542, opened 2026-06-07
Retire dependency registry scans

jesse-merhi:
## Summary

- Stop scheduling dependency registry existence scans after skill publish.
- Retire the dependency registry scanner implementation while keeping a no-op internal action so already queued Convex jobs drain safely.
- Keep the v1 `security.signals.dependencyRegistry` response key for compatibility, but always return `null` because the scanner is retired.
- Update HTTP API docs, CLI fixtures, and tests to reflect the retired scanner.
- Add an admin-only guarded cleanup action for stored dependency registry scan data.
- Keep the legacy dependency registry schema fields, cache table, and scan-status index so existing production data remains valid until an approved cleanup runs.
- Remove the old dependency-registry static finding reason code from active scanner output.

No production cleanup command has been run. This PR removes scanner code and public scanner behavior, but it does not delete stored dependency registry data by itself.

## Maintainer decisions recorded

Accepted for this PR:

- Retire npm/PyPI/Cargo registry existence checks as a ClawHub security signal.
- Rely on ClawScan and other current security signals instead of dependency registry existence checks.
- Preserve the v1 `dependencyRegistry` key as a nullable compatibility placeholder.
- Ship the cleanup command as guarded operator tooling only.

Required before any production cleanup apply:

- Run the cleanup in dry-run mode.
- Review the dry-run output for the exact moderation fields, version fields, and cache rows that would change.
- Get explicit operator approval for that production cleanup run.

## Cleanup command

Dry-run first:

```bash
bunx convex run maintenance:cleanupDependencyRegistryScanData '{"dryRun":true,"batchSize":100,"maxBatches":20}' --prod
```

Apply only after reviewing dry-run output and approving the delete:

```bash
bunx convex run maintenance:cleanupDependencyRegistryScanData '{"dryRun":false,"confirm":"delete-dependency-registry-scan-data","batchSize":100,"maxBatches":1}' --prod
```

The apply path requires the exact confirmation token and only patches/deletes IDs returned by the cleanup page query:

- unsets `skillVersions.depRegistryAnalysis`
- unsets `skillVersions.depRegistryScanStatus`
- removes `suspicious.dep_not_found_on_registry` from legacy static scan findings/reason codes
- removes that same retired code from skill moderation reason/evidence fields
- deletes `depRegistryCache` rows by row ID

## Validation

Latest focused checks passed on head `410181bbbbfd36d56f0edff9296ca5b504c59704`:

- `bun run test -- convex/depRegistryScan.test.ts --reporter=dot` — 1 passed
- `bun run test -- convex/depRegistryScan.test.ts convex/httpApiV1.handlers.test.ts convex/maintenance.test.ts --reporter=dot` — 318 passed
- `git diff --check`

Earlier local checks passed on the previous head before the spec-only PR cleanup:

- `bun run format:check`
- `bun run lint`
- `bun run --cwd packages/clawhub test:src -- src/cli/commands/inspect.test.ts --reporter=dot` — 12 passed
- `bun run test:e2e -- e2e/clawhub.e2e.test.ts --testNamePattern "skill verify accepts the legacy json flag"`
- `bunx convex codegen`
- `git diff --check 8e1bad4ac1f0d68309e168623c1c0179622e684d`
- `bun run test -- convex/lib/skillPublish.test.ts --reporter=dot --environment=node` — 7 passed

Remote GitHub checks passed for previous head `6e0d57b374a0bd3fc7bc34bbc67472382f67bb2c`, including `static`, `unit`, `types-build`, `packages`, `e2e-http`, `playwright-local-auth`, `playwright-smoke`, CodeQL, and secret scanning. New checks are now running for `410181bbbbfd36d56f0edff9296ca5b504c59704`.

Local broad checks that did not complete on the previous head:

- `bun run ci:unit` failed locally from Bun/jsdom worker startup errors: `ERR_REQUIRE_ESM` from `html-encoding-sniffer` requiring ESM `@exodus/bytes/encoding-lite.js`. It reached 1,457 tests before coverage failed because workers did not start.
- `bun run ci:static` failed locally at `deadcode:knip` from `ERR_REQUIRE_ESM` in `oxc-parser`.


jesse-merhi:
Maintainer acceptance / validation note for this PR:

- Accepted: retire the npm/PyPI/Cargo dependency-registry existence scanner and rely on ClawScan / other current security signals instead.
- Accepted: keep the v1 `security.signals.dependencyRegistry` key for compatibility, but always return `null` because the scanner is retired.
- Accepted: include guarded cleanup code for legacy dependency-registry data. Cleanup is dry-run by default, admin-gated, and requires the explicit `delete-dependency-registry-scan-data` confirmation token before applying. No cleanup/data deletion was run as part of this PR.
- Required before any production cleanup apply: review the dry-run output and get explicit operator approval for that run.

Validation evidence for head `6e0d57b374a0bd3fc7bc34bbc67472382f67bb2c`:

- GitHub checks passed: `static`, `unit`, `types-build`, `packages`, `e2e-http`, `playwright-local-auth`, `playwright-smoke`, CodeQL, and secret scanning.
- Local focused checks passed: `format:check`, `lint`, focused Convex/API cleanup tests, CLI inspect test, legacy verify e2e test, root/package TypeScript checks, `convex codegen`, and `git diff --check`.
- Local full `ci:unit` and local `ci:static` were attempted but hit local Bun/Node ESM loader issues in jsdom/knip; the corresponding GitHub `unit` and `static` checks passed.

Issue #2374, opened 2026-05-22
Skill embedding doesn't seem to refresh on publish


SFKislev:
I've been republishing my skill (`the-news`) with substantially different SKILL.md content across several versions and the search score barely moves — only by the amount I'd expect from download growth, not from content changes.

I checked locally by reproducing the embedding pipeline (same model, `text-embedding-3-small`, same `buildEmbeddingText` logic including the byte cap from #2337) and the predicted cosine deltas from my edits are real and meaningful. But they don't show up in the live ranking.

There's also a tell: I added some distinctive words to the body that don't appear in any other published skill, and queries for those words don't return my skill at all. If the embedding had been regenerated, they should.

Static scan is clean, no moderation flags, every publish goes through fine — it just doesn't seem to take. My guess is silent failure somewhere in the embedding generation path, but I can't confirm without seeing the stored vector.

Would be helpful either way to know whether `the-news`'s embedding is actually current. Related: #2293 would have made this diagnosable from the outside.

momothemage:
Thank you for providing the information. However, before we actually try to resolve your question, we would appreciate it if you could describe it more specifically. For example, what changes did you make, and what search terms did you use? 

I added some distinctive words to the body that don't appear in any other published skill, and queries for those words don't return my skill at all. -- What specific words were used and how did you search for them? Words in the slug and description are usually more likely to match the search terms. At the same time, we will also check the embedding generation path.

SFKislev:
hi!
I did more debugging and I think the issue may not be stale embeddings after all.
The skill in question is https://clawhub.ai/sfkislev/the-news . I want it to rank for a "news" search. It doesn't, and I see in second place a random-looking cctv-news-fetcher, which is clearly not what humans and bots would want when they search for "news". 

For query `news`, live `/api/v1/search?q=news&limit=200` currently gives:

| `the-news` | `2.5900` score | `2.5` lexical boost | `0.09` popularity bonus | `~0.0000` implied vector score |
| `cctv-news-fetcher` | `2.9468`  score | `2.5`  lexical boost | `0.09` popularity bonus | `0.3568` implied vector score |

Both skills get the same lexical boost for `news`:
- slug token match: `+1.4`
- display-name token match: `+1.1`
- total lexical: `+2.5`

Both also hit the popularity cap of `0.09`.

Using the production embedding input construction (`SKILL.md` plus non-`.md` extra files only, 7,500-byte cap), local `text-embedding-3-small` cosine for query `news` is:

[...]

PR #2591, opened 2026-06-10
Preserve vector scores for direct lexical skill matches

SFKislev:
## Summary
- preserve a skill's vector score by skill id when the same skill is also recalled through direct lexical search
- keep existing merge/recall ordering intact while allowing lexical-retained entries to receive their vector contribution
- add a regression test for a direct lexical match shadowing a hydrated vector candidate

## Context
This addresses the ranking failure discussed in #2374. For searches like `news`, a skill can be recalled first through `directPrefixSkillMatches` without an `embeddingId`, then also appear as a vector-hydrated candidate with an `embeddingId`. The current merge keeps the direct lexical entry and drops the hydrated duplicate, so scoring falls back to a zero vector score.

The fix stores vector scores by skill id as hydrated candidates are seen, then uses that map when the retained entry does not carry `embeddingId`.

## Tests
- `bunx vitest run convex/search.test.ts`
- `bunx oxfmt --check convex/search.ts convex/search.test.ts`
- `bunx oxlint --type-aware --tsconfig ./tsconfig.oxlint.json convex/search.ts convex/search.test.ts`

### DIFF
--- a/convex/search.test.ts
+++ b/convex/search.test.ts
@@ -1521,6 +1521,46 @@ describe("search helpers", () => {
     expect(resultA).toBeDefined();
     expect(resultA!.score).toBeGreaterThan(1.0);
   });
+
+  it("preserves vector scores when a direct lexical match shadows the hydrated candidate", async () => {
+    generateEmbeddingMock.mockResolvedValueOnce([0, 1, 2]);
+
+    const lexicalEntry = {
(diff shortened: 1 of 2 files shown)


Patrick-Erichsen:
Thanks for the fix @SFKislev !

PR #2643, opened 2026-06-15
fix: normalize publisher recovery handle comparisons

Patrick-Erichsen:
## Summary
- compare stored recovery user handles using canonical publisher-handle normalization
- retire a legacy mixed-case user handle before assigning the recovered publisher handle
- cover the production-shaped mixed-case recovery regression

## Proof
- before fix, the focused regression failed with `Next GitHub provider id does not belong to @gingiris-1031`
- after fix, `bunx vitest run convex/publishers.test.ts` passes all 76 tests

## Tests
- `bun run ci:static`
- `bun run ci:unit`
- `bun run ci:types-build`
- `bun run ci:packages`
- `bun run ci:e2e-http`

### DIFF
--- a/convex/publishers.test.ts
+++ b/convex/publishers.test.ts
@@ -4166,6 +4166,7 @@ describe("legacy publisher migration", () => {
     options: {
       destinationHasResources?: boolean;
       legacyResources?: boolean;
+      mixedCaseUserHandles?: boolean;
       tooManyLegacySkills?: boolean;
       unexpectedResourceOwner?: boolean;
       unexpectedReservationOwner?: boolean;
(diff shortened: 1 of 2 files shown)


Patrick-Erichsen:
Maintainer review: confirmed that this change only canonicalizes the mutable handle-casing guard. The previous and next immutable GitHub `providerAccountId` values are still resolved first, must each map unambiguously to distinct active users, and remain authoritative for the recovery.

Accepting focused unit coverage for this pure comparison/retirement-branch change. The regression passed locally, all code/package/build/HTTP CI gates are green, and the post-deploy production recovery command will be run in dry-run mode only as runtime proof.

PR #2718, opened 2026-06-18
fix: redirect sign-in to dashboard

vyctorbrzezowski:
## Summary

- route the global GitHub sign-in flow to `/dashboard`
- keep contextual sign-in flows unchanged
- update header authentication tests for the proposed destination

## Why

After authentication, the primary header sign-in should open the user's dashboard instead of returning to the home page.

## Validation

- `bunx vitest run src/__tests__/header.test.tsx`
- `bun run ci:static`
- `VITE_CONVEX_URL=https://example.invalid bun run ci:unit`
- `git diff --check`

No screenshot is included because this changes post-authentication navigation without changing the rendered UI.

### DIFF
--- a/src/__tests__/header.test.tsx
+++ b/src/__tests__/header.test.tsx
@@ -342,7 +342,7 @@ describe("Header", () => {
 
     fireEvent.click(screen.getByRole("button", { name: "Sign in with GitHub" }));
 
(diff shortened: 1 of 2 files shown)


vyctorbrzezowski:
@clawsweeper review

PR #2732, opened 2026-06-18
feat: move topic discovery into search and category browse

fuller-stack-dev:
## Summary

- Remove the open-ended global Topics sections from the Skills and Plugins browse sidebars while preserving direct normalized topic browse links.
- Show up to five normalized top-topic chips beneath a selected category, ranked from a bounded sample of that category's highest-ranked public items.
- Let topic chips update the URL-backed topic filter; selecting the active chip clears it.
- Treat skill categories, plugin categories, and author topics as explicit search relevance evidence.
- Recall exact author-topic matches before bounded topic-prefix expansion and exclude the generic `other` fallback category from search evidence.
- Document the navigation/search/category-topic split and add regression coverage across browse, global skill, plugin package, and skill package-catalog search.

## Linked Issue

- Closes N/A
- Related N/A

## Screenshots

- [x] Screenshots/recordings attached, or `N/A`
  - [Real Chromium UI proof: selected-category topic chips on Skills and Plugins, including desktop/mobile and active-topic state](https://github.com/openclaw/clawhub/pull/2732#issuecomment-4745486740)
  - [Published proof artifact tree](https://github.com/openclaw/clawhub/tree/qa-artifacts/clawhub-ui-proof/pr-2732/local-category-topics)

## Behavioural Proof

- [x] Behavioural proof included, or `N/A`
  - Added category-topic query coverage for normalized top-five ranking, category exclusion, public-item filtering, kind/category validation, and a category sample beginning after the first 240 globally ranked skills.
  - Added route regressions proving selected categories render five chips, chip selection filters by normalized topic, and selecting the active chip clears it.
  - Added search regressions for category-only matches, partial author-topic matches, exact-topic priority under a bounded recall budget, and exclusion of `other`.
  - `bunx vitest run convex/catalogTopics.test.ts src/__tests__/skills-index.test.tsx src/__tests__/packages-route.test.tsx` passed: 105 tests.
  - Real Chromium verification confirmed selected-category chips on Skills and Plugins, active-chip URL state, no global Topics heading, and mobile wrapping without overflow.

## Security / Trust Impact

- [x] No security/trust impact
- [ ] Security/trust impact explained

## Data / Deploy Impact

- [ ] No data/deploy impact
- [x] Data/deploy impact explained
  - No schema migration or backfill is required.
  - The backend search and category-topic query functions require the normal manual production deployment after merge.
[...]

PR #2733, opened 2026-06-18
fix(deps): resolve static audit advisories

fuller-stack-dev:
## Summary
- bump the DOMPurify override to the patched `3.4.11` release
- force transitive undici consumers, including jsdom, onto patched `7.28.0`
- remove the stale nested undici `7.27.1` lockfile resolution without adding audit ignores

This resolves one high-severity and two moderate dependency advisories that caused the repository-wide static gate to fail.

## Verification
- `bun audit`
- `bun run ci:static`
- `bun run ci:unit`
- `bun run ci:packages`
- `bun run ci:types-build`
- `bun run ci:e2e-http`
- `git diff --check`
- `$autoreview` (clean, no actionable findings)

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -149,10 +149,11 @@
   },
   "overrides": {
     "ast-v8-to-istanbul": "1.0.4",
-    "dompurify": "3.4.10",
+    "dompurify": "3.4.11",
     "esbuild": "0.28.1",
     "next": "16.2.6",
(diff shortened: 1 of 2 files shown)


fuller-stack-dev:
Superseded by the same dependency remediation now included in #2732 so its blocking static audit can pass.

PR #2758, opened 2026-06-19
fix: align official plugin browse with official publishers

vyctorbrzezowski:
## Summary

- add a family-and-official installs index for package listings
- use that index for official plugin pages sorted by installs
- add an admin repair path to normalize public plugins from official publishers to `channel: "official"` / `isOfficial: true`
- keep private plugins private and out of official browse

## Root cause

`/api/v1/plugins?isOfficial=true&sort=installs` queried code and bundle plugin families separately. Each family used the broad family installs index and filtered `isOfficial` afterward, so asking for more official items than one family currently had could scan community rows until Convex hit the read limit and returned HTTP 500.

There was also a data-contract mismatch for legacy official publisher packages: publisher profiles treat items from an Official publisher as official, but public plugin browse relies on denormalized package/digest fields. This PR keeps browse fast and index-backed by repairing public `code-plugin` and `bundle-plugin` packages owned by official publishers so their package and digest fields match the product rule.

## Product rule

Any public plugin owned by an Official publisher is Official. Private plugins stay private.

## Operational note

After deploy, run `packages:normalizeOfficialPublisherPackages` for both plugin families to repair existing production rows:

```text
family=code-plugin, dryRun=true
family=code-plugin, dryRun=false
family=bundle-plugin, dryRun=true
family=bundle-plugin, dryRun=false
```

Use the returned cursor to continue if a family needs multiple batches.

## Reproduction before this PR

```text
/api/v1/plugins?isOfficial=true&sort=installs&limit=4 -> HTTP 200
/api/v1/plugins?isOfficial=true&sort=installs&limit=5 -> HTTP 500
```

## Validation

- `bun run test -- convex/packages.public.test.ts convex/httpApiV1.handlers.test.ts` (608 passed)
- `bun run ci:unit` (3,576 passed, 1 skipped)
- `bun run ci:types-build`
- `bun run ci:static`
- `git diff --check`

### DIFF
--- a/convex/httpApiV1.handlers.test.ts
+++ b/convex/httpApiV1.handlers.test.ts
@@ -8799,7 +8799,7 @@ describe("httpApiV1 handlers", () => {
     }
   });
 
-  it("plugins list install sort forwards to both plugin families and merges by installs", async () => {
+  it("official plugins install sort forwards the filter to both families and merges by installs", async () => {
     const codePlugin = makeCatalogItem("code-installed", {
       family: "code-plugin",
       updatedAt: 100,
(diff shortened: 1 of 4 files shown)


vyctorbrzezowski:
@clawsweeper re-review

PR #2760, opened 2026-06-20
fix(web): filter New listings, trim CLI steps, and link creator cards

vyctorbrzezowski:
## Summary

This PR proposes exactly three homepage follow-ups:

1. **Filter the New listings**
   - Hide Skills and Plugins with `pending` or `suspicious` audit states from the **New** tabs.
   - Apply the same rule to catalog searches performed while **New** is active.
   - Apply plugin exclusions before pagination and skill exclusions before the final search result limit, allowing eligible matches to refill the page.

2. **Remove the retired CLI step**
   - Remove `clawhub sync --all` from **Bring your skills to ClawHub**.

3. **Link the Popular creators cards**
   - Make each creator card a normal navigable link.
   - Preserve drag-to-scroll behavior by starting pointer capture only after the pointer crosses the drag threshold.

## Behavioural Proof

- Opened the homepage locally at `/` and verified all three proposed changes in the real app: the Skills/Plugins New tabs, the shortened CLI command list, and clickable creator cards.
- Verified New plugin browse and search requests forward `pending,suspicious` exclusions to both plugin families.
- Verified New skill search requests server-side non-suspicious and non-pending results, with filtering performed before the final result limit.
- Verified `not-run` plugin scans remain eligible; only `pending` and `suspicious` are excluded by this change.

- [x] Behavioural proof included

## Screenshots

The changes reuse the existing homepage layout and styling; no layout or visual redesign is proposed.

- [x] N/A

## Security / Trust Impact

The proposed catalog behavior reduces exposure to items whose audit is still pending or currently suspicious. It does not change moderation verdicts, publishing, installability, or public response schemas.

- [x] Security/trust impact explained

## Data / Deploy Impact

The plugin catalog endpoint gains an optional `excludeScanStatus` query filter, and skill search gains an internal pending-scan exclusion option. No schema change, migration, or backfill is required. The Convex backend must be deployed with the merged revision before the frontend uses the new filters in production.

- [x] Data/deploy impact explained

## Verification

- [x] `bun run ci:static`
- [x] Focused tests:
  - `bun run test -- convex/search.test.ts src/__tests__/home-listing-section.test.tsx src/__tests__/home-bring-skills-section.test.tsx src/__tests__/home-popular-publishers-section.test.tsx convex/packages.public.test.ts convex/httpApiV1.handlers.test.ts`
  - Result: 6 files passed, 695 tests passed.
- [x] `bun run ci:unit`
  - Result: 283 files passed, 1 skipped; 3,605 tests passed, 1 skipped; coverage thresholds passed.
- [x] `bun run ci:types-build`
  - Includes root TypeScript, schema package TypeScript, CLI package TypeScript, admin typecheck, and production build.
- [x] `git diff --check`
- [x] Two direct `codex review` passes. The first found missing suspicious/search handling; the second found post-limit filtering in New skill search. Both findings were fixed and covered by regression tests.

### DIFF
--- a/convex/httpApiV1.handlers.test.ts
+++ b/convex/httpApiV1.handlers.test.ts
@@ -8841,6 +8841,35 @@ describe("httpApiV1 handlers", () => {
     ]);
   });
 
+  it("plugins list forwards excluded scan statuses to both plugin families", async () => {
+    const runQuery = vi.fn((_, args: Record<string, unknown>) => {
+      if (Object.keys(args).length === 0) {
+        throw new Error("filtered listings must not query the global count");
+      }
+      return { page: [], isDone: true, continueCursor: "" };
+    });
+    const runMutation = vi.fn().mockResolvedValue(okRate());
+
+    const response = await __handlers.listPluginsV1Handler(
+      makeCtx({ runQuery, runMutation }),
+      new Request(
+        "https://example.com/api/v1/plugins?sort=updated&excludeScanStatus=pending,suspicious",
+      ),
+    );
+
+    expect(response.status).toBe(200);
+    const familyCalls = runQuery.mock.calls.filter(([, args]) => "family" in args);
+    expect(familyCalls).toHaveLength(2);
+    for (const [, args] of familyCalls) {
+      expect(args).toEqual(
+        expect.objectContaining({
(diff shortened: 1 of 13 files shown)


vyctorbrzezowski:
@clawsweeper review

PR #2764, opened 2026-06-20
perf: cache homepage listing tabs

vyctorbrzezowski:
Visual proof: [homepage listing cache walkthrough](https://tawny-nectar-xbms.here.now/droppie-2026-06-20T18-19-53Z.mp4)

## Summary
- Adds a home route loader for the default Skills / Top listing and passes that payload into the homepage listing section.
- Seeds the client-side listing cache from that initial payload, so the first visible tab does not fetch again after hydration.
- Keeps other browse tabs lazy-loaded and cached by kind, tab, categories, and limit after first visit.
- Leaves search fetch-on-query and does not add prefetching or load all tabs upfront.
- Falls back to normal client loading if the initial listing loader fails, so a catalog hiccup does not take down the full homepage.

## Tests
- `bun run test -- src/__tests__/home-listing-section.test.tsx src/__tests__/home-route.test.tsx`
- `bun run ci:static`
- `bun run ci:unit`
- `bun run ci:types-build`
- `git diff --check upstream/main`

## Browser proof
- Local homepage checked at `http://127.0.0.1:13438/` using production Convex env.
- Verified Skills / Top arrives from initial data with no browser `skills:listPublicPageV4` refetch.
- Verified Top → Trending → Top reuses cached Top data.
- Verified Plugins Official → Top → Official reuses cached Official plugin data after first load.

### DIFF
--- a/src/__tests__/home-listing-section.test.tsx
+++ b/src/__tests__/home-listing-section.test.tsx
@@ -100,6 +100,44 @@ describe("HomeListingSection", () => {
     });
   });
 
+  it("renders the initial Skills Top listing without refetching on mount", async () => {
+    render(
+      <HomeListingSection
+        initialListing={{
+          kind: "skills",
+          tab: "popular",
+          categorySlugs: [],
+          fetchLimit: 20,
+          items: [
+            {
+              skill: {
+                _id: "skills:initial" as never,
+                slug: "initial-skill",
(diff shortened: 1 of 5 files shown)


vyctorbrzezowski:
@clawsweeper review

vyctorbrzezowski:
@clawsweeper re-review

PR #2774, opened 2026-06-22
fix: harden CSP and staged secret output

jesse-merhi:
## Summary

This hardens two defense-in-depth surfaces from `GHSA-p2qm-w5vv-54j3`. The staged-secret scanner still blocks commits when it finds likely secrets, but it no longer stores or prints matched token text. The app now serves a dynamic CSP with per-request script nonces, so ClawHub can remove global script `'unsafe-inline'` while keeping TanStack Start SSR scripts and the self-hosted theme bootstrap working.

## What Changed

- Staged-secret scanner output: findings now print path, reason, and `[REDACTED]`; token-shaped material is also redacted from blocked staged paths.
- Theme bootstrap: the existing pre-hydration theme/reset logic moved from an inline script in `src/routes/__root.tsx` to `public/theme-bootstrap.js`, loaded as `/theme-bootstrap.js`.
- CSP runtime: `src/server.ts` generates a per-request nonce, passes it into TanStack SSR, and sets a matching `Content-Security-Policy` header.
- CSP policy: production script CSP removes global `'unsafe-inline'` and uses `'self' 'nonce-...' 'unsafe-eval'`. The eval allowance is intentional for the current client bundle because shared `clawhub-schema` helpers pull in arktype, which probes eval support with `Function()`; inline scripts remain nonce-gated.
- Local development CSP: loopback responses keep HTTP/WebSocket and docs-auth form-post allowances needed for development.
- Vercel headers: the static global CSP was removed from `vercel.json` so it does not conflict with dynamic nonces; other security headers remain.
- Regression coverage: added scanner redaction tests, CSP helper tests, local-development CSP tests, external-bootstrap behavior tests, and an explicit test documenting that `unsafe-eval` must not reopen `'unsafe-inline'`.

## Proof

Behavioral proof:

- A staged fake GitHub token is blocked, and stderr contains `[REDACTED]` without the token prefix or token value.
- A token-shaped staged path such as `leaks-<token>/.env.local` is blocked and printed as `leaks-[REDACTED]/.env.local`.
- Built SSR responses include a fresh nonce in `script-src`; `/theme-bootstrap.js` is loaded as a same-origin external script, and all inline TanStack/runtime scripts carry the matching nonce.
- Browser CSP probe against a built local preview loaded `/`, `/skills`, and `/plugins` with no `securitypolicyviolation` events, no CSP console errors, and no page errors.

Reviewer-checkable CSP sample from the rebuilt local preview:

```text
status 200
script-src 'self' 'nonce-MMoWNwba1xT0oXHKu43OKvOY' 'unsafe-eval'
nonceLength 24
hasUnsafeInlineScript false
hasUnsafeEval true
scriptTags 4
noncedScriptCount 3
externalTheme true
unnoncedInlineScripts 0
browserCspViolations []
```

Screenshots: omitted because the PR intentionally preserves visible UI state. The changed behavior is header/script execution and scanner stderr output, which the commands and CSP sample prove more directly than an unchanged page screenshot.

## Verification

Automated checks passed locally:

- `bun run check:secrets`
- `bun run test -- scripts/check-staged-secrets.test.mjs scripts/vercel-security-headers.test.mjs`
- `bun run format:check -- knip.config.ts public/theme-bootstrap.js scripts/check-staged-secrets.mjs scripts/check-staged-secrets.test.mjs scripts/vercel-security-headers.test.mjs src/lib/securityHeaders.ts src/routes/__root.tsx src/server.ts vercel.json`
- `bun run lint:oxlint`
- `bun run deadcode:files`
- `VITE_CONVEX_URL=https://example.invalid bun run build`
- `bun run ci:static`
- `bun run ci:unit`

Additional CSP breakage check after the eval finding:

- `bun run test -- scripts/vercel-security-headers.test.mjs scripts/check-staged-secrets.test.mjs`
- `bun run format:check -- src/lib/securityHeaders.ts scripts/vercel-security-headers.test.mjs`
- `VITE_CONVEX_URL=https://wry-manatee-359.convex.cloud VITE_CONVEX_SITE_URL=https://wry-manatee-359.convex.site bun run build`
- Built preview on `http://127.0.0.1:4177/` plus Playwright CSP probe for `/`, `/skills`, and `/plugins`: zero CSP violations.
- `bun run ci:static`

Known residual:

- `bunx tsc --noEmit --pretty false` fails in this stale local checkout at `e2e/prod-http-smoke.e2e.test.ts:94` with `Cannot find name 'lastError'`. That file is outside this PR's net diff; remote `types-build` passed before the latest amended push, and the scoped CSP/scanner checks above pass.
- Future hardening can remove `'unsafe-eval'` by splitting browser-safe catalog/schema helpers away from arktype-backed runtime validators.

### DIFF
--- a/knip.config.ts
+++ b/knip.config.ts
@@ -34,6 +34,7 @@ const config = {
     ".": {
       entry: [
         "src/router.tsx!",
+        "src/server.ts!",
         "src/routes/**/*.{ts,tsx}!",
         "src/styles.css!",
         "server/**/*.{ts,tsx}!",
--- a/public/theme-bootstrap.js
+++ b/public/theme-bootstrap.js
@@ -1,0 +1,117 @@
+(function () {
+  try {
+    var d = document.documentElement;
+    var selectionKey = "clawhub-theme-selection";
+    var modeKey = "clawhub-theme";
+    var nameKey = "clawhub-theme-name";
+    var legacyModeKey = "clawdhub-theme";
+    var customThemeKey = "clawhub-custom-theme";
+    var preferencesKey = "clawhub-preferences";
+    var defaults = { theme: "claw", mode: "system" };
+    var storageKeys = [
+      customThemeKey,
+      preferencesKey,
+      selectionKey,
+      modeKey,
+      nameKey,
+      legacyModeKey,
+    ];
+    var cookieKeys = storageKeys;
+
+    function hasCookie(name) {
+      if (!document.cookie) return false;
+      return document.cookie.split(";").some(function (part) {
+        return part.trim().indexOf(name + "=") === 0;
+      });
+    }
+
+    function clearCookie(name) {
+      document.cookie = name + "=; Max-Age=0; path=/";
+      document.cookie = name + "=; expires=Thu, 01 Jan 1970 00:00:00 GMT; path=/";
+    }
+
+    function cleanupDom() {
+      var style = document.getElementById("clawhub-custom-theme-style");
+      if (style) style.remove();
+      var fonts = document.getElementById("clawhub-custom-theme-fonts");
+      if (fonts) fonts.remove();
+      d.classList.remove("theme-custom", "high-contrast", "reduce-motion");
+      delete d.dataset.density;
+      delete d.dataset.animation;
+      d.style.removeProperty("--code-font-size");
+    }
+
+    var reset = false;
+    try {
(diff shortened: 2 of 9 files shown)


jesse-merhi:
Superseded by the split PRs: #2803 for staged-secret scanner redaction and #2804 for CSP hardening.

PR #2808, opened 2026-06-23
fix: prefer publisher handles over plugin aliases

### LINKED ISSUE #2807: Publisher handle can be shadowed by OpenClaw plugin alias on top-level profile routes
## Summary

Creating an org/publisher whose handle matches an OpenClaw official plugin alias can make the publisher profile unreachable from the top-level profile route.

Example: an org named `tencent` should open the `tencent` org profile, but clicking it redirects to `https://clawhub.ai/openclaw/plugins/tencent-provider`.

## Reproduction

1. Create an org/publisher with handle `tencent`.
2. Open a page that links to that publisher, such as the publishers catalog or any publisher badge/list item.
3. Click the `tencent` publisher.

## Expected behavior

The click should open the `tencent` publisher/org profile.

## Actual behavior

The click redirects to the OpenClaw Tencent provider plugin page:

`/openclaw/plugins/tencent-provider`

## Why this appears to happen

On current `upstream/main` (`f4f6f3454250bcfcf485b89991cf1134220283e0`), publisher profile links now use the top-level slug route:

- `src/components/PublisherListItem.tsx` links publishers to `/$slug`.
- `src/lib/ownerRoute.ts` builds publisher profile hrefs as `/${handle}`.
- `src/routes/$slug.tsx` resolves `/$slug` through `resolveTopLevelSlugRoute`.

However, `resolveTopLevelSlugRoute` currently checks OpenClaw plugin aliases before publisher handles:

```ts
const plugin = await resolveOpenClawPluginSlug(slug);
if (plugin) return plugin;

const publisher = await resolvePublisherHandle(slug);
if (publisher) return { kind: "publisher", ... };
```

And `src/lib/openClawExtensionSlugs.ts` maps:

```ts
tencent: "@openclaw/tencent-provider"
```

So `/tencent` resolves as the official plugin alias before the `tencent` publisher handle is checked.

## Suggested fix direction

Consider resolving explicit publisher handles before OpenClaw plugin aliases on the top-level `/$slug` route, or avoid using top-level `/${handle}` as the canonical publisher profile URL when it collides with legacy/official plugin aliases.

Publisher handles are user-created profile identifiers, so they probably should not be shadowed by plugin alias shortcuts.

## Notes

This may affect any publisher handle that matches an entry in `OPENCLAW_EXTENSION_SLUG_TO_PACKAGE`, not only `tencent`.

momothemage:
## Summary

- Resolve top-level publisher handles before OpenClaw plugin aliases so an org like `tencent` opens its publisher profile instead of `/openclaw/plugins/tencent-provider`.
- Keep official OpenClaw plugin aliases available when no publisher owns the handle.
- Update the slug-routing spec to document the publisher-first precedence.

Closes #2807.

## Root Cause

`resolveTopLevelSlugRoute()` checked OpenClaw official plugin aliases before checking publisher handles. After #2793 made `/<handle>` a publisher profile route, handles such as `tencent` could be shadowed by entries in `OPENCLAW_EXTENSION_SLUG_TO_PACKAGE`.

Provenance: made visible by #2793 (`55fe4a856333b9941120528507b56351e4aeca4c`, 2026-06-23), which added owner-qualified catalog routes and placed publisher handle resolution behind existing plugin alias resolution.

## Tests

- `bunx vitest run src/lib/slugRoute.test.ts src/__tests__/slug-route-loader.test.ts src/__tests__/skill-route-loader.test.ts`
- `bun run format:check -- src/lib/slugRoute.ts src/lib/slugRoute.test.ts specs/slug-routing.md`
- `bun run ci:static`
- `VITE_CONVEX_URL=https://example.invalid bun run coverage` from `/private/tmp/clawhub-issue-2807-test`: 294 passed, 1 skipped; 3766 tests passed, 1 skipped

Note: the first full coverage attempt from `/Users/mauriceniu/Documents/New project/clawhub-issue-2807` hit an existing path-space issue in `scripts/skill-cards/run-skill-card-worker.test.ts` (`New%20project` in a template path). The same diff passed from a no-space temp worktree.

### DIFF
--- a/specs/slug-routing.md
+++ b/specs/slug-routing.md
@@ -49,8 +49,9 @@ Plugins:
 - Encoded security compatibility page: `/plugins/%40scope%2Fname/security-audit`
 - Legacy scanner pages redirect to the corresponding plugin security audit page.
 
-Bare `/<handle>` routes are profile routes after static routes and official
-OpenClaw plugin aliases have won precedence.
+Bare `/<handle>` routes are profile routes after static routes and existing
+publisher handles have won precedence. Official OpenClaw plugin aliases remain
+available at the root only when no active publisher owns that handle.
 
 Encoded compatibility routes are npm-style package-name routes. They redirect
 with `308` to the readable scoped route so the address bar shows
@@ -91,9 +92,10 @@ The effective precedence is:
 
 1. Static app routes win first, such as `/search`, `/settings`, `/plugins`, and
    `/api/...`.
-2. A top-level path matching an official OpenClaw extension alias redirects to
-   that plugin package.
-3. Any other top-level path may resolve as a publisher profile.
+2. A top-level path matching an active publisher handle resolves as that
+   publisher profile.
+3. Any other top-level path matching an official OpenClaw extension alias
+   redirects to that plugin package.
 4. Unknown top-level paths may still resolve through the historical skill slug
    fallback and redirect to `/<owner>/skills/<slug>`.
 5. `/openclaw/<alias>` and `/@openclaw/<alias>` only resolve official OpenClaw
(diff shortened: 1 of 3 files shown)


momothemage:
Closing this as superseded by `main` commit 9b352d7f99005f9d07f49f285a55cf43715bb130 (`fix(routing): protect publisher aliases from plugin collisions`).

That landed fix takes the product direction of preserving official OpenClaw top-level aliases while keeping colliding publisher profiles reachable:

- `buildPublisherProfileHref()` now sends handles that match official OpenClaw aliases to `/user/<handle>`.
- `/user/$handle` no longer redirects away for those legacy alias handles, so existing publishers such as `tencent` remain reachable.
- publisher/user creation and handle-claim paths now reserve the OpenClaw alias set, preventing new collisions.
- regression coverage was added on `main` for the alias collision behavior.

So this PR is now conflicting with `main` and represents an alternate policy (`/<handle>` publisher precedence) rather than the landed fix. CI was green except the expected fork Vercel authorization failure, but there is nothing useful left to merge from this branch.

PR #2881, opened 2026-06-25
fix(search): keep skills all tab on approved recommended ranking

vyctorbrzezowski:
## Summary

- Keeps the `/skills` **All** tab on recommended discovery ranking instead of silently falling back to `updatedAt` ordering when recommendation scores are missing or stale (uses the recommended rank index as a score-based proxy).
- Leaves **Recently updated** as the explicit `updatedAt desc` sort only.
- Excludes pending-review, suspicious, hidden, and other non-public items from public browse/search.
- When a skill already has an approved public release and a newer version is pending review, browse/search keeps showing the approved public state (not the pending version).

## Test plan

- [x] `bunx vitest run` on touched Convex tests (`publicBrowse`, `listPublicPageV4`, `publicListCursor`, `packageCatalog`, `search`, `backportLatest`)
- [x] `git diff --check`
- [x] `bun run ci:static`
- [x] `bun run ci:unit`

### DIFF
--- a/convex/lib/publicBrowse.test.ts
+++ b/convex/lib/publicBrowse.test.ts
@@ -4,6 +4,7 @@ import {
   hasPriorApprovedPublicSkillVersion,
   isPubliclyListableSkillVersion,
   isSkillPendingPublicReview,
+  resolvePublicBrowseVersionForSkill,
   shouldExcludeSkillFromPublicBrowse,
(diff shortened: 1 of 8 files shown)


vyctorbrzezowski:
@clawsweeper review

vyctorbrzezowski:
@clawsweeper re-review

PR #2882, opened 2026-06-25
fix: require org member acceptance

jesse-merhi:
## Summary

- Stop self-serve org owners/admins from creating a first membership row for arbitrary user handles.
- Keep existing-member role updates working, but require new organization membership to happen through a future recipient-accepted invite/repair path.
- Remove the settings-page direct “Add member” form until an accepted invite flow exists.
- Document the org membership invariant in `specs/orgs.md`.

## Why

Org membership creates a public association between a user and a publisher. Creating that first association should require recipient acceptance or an admin-gated repair path, not only the current org owner/admin selecting a handle.

This change preserves role updates for users who are already members while blocking unilateral first-time membership creation from the public mutation.

## Verification

- `bunx vitest run convex/publishers.test.ts src/routes/-settings.test.tsx`
- `bun run ci:static`
- `bun run ci:unit`
- `npm_config_registry=https://registry.npmjs.org bunx convex dev --once --typecheck=disable` against dev deployment `admired-dodo-615`
- `bun run format:check -- convex/_generated/api.d.ts convex/publishers.ts convex/publishers.test.ts src/routes/settings.tsx specs/orgs.md`
- `git diff --check`

Autoreview was attempted with `.agents/skills/autoreview/scripts/autoreview --mode local`, but the local Codex review engine failed with `401 Unauthorized` before producing findings.


vyctorbrzezowski:
image
this is sending me 

nice one

PR #2947, opened 2026-07-01
Add package changelog previews for plugin publishing

### LINKED ISSUE #1288: bug: publish-plugin page cannot upload plugin folders; auto-changelog overwrites edits; package changelog uses wrong history
### 1. Cannot upload plugin folders — \"Open\" button stays grayed out

On the publish-plugin page, when a user clicks **Choose folder** and selects a plugin directory, the native file picker's **Open** button remains grayed out and the folder cannot be selected. The page never transitions to showing uploaded files. This makes it impossible to publish a plugin via folder upload.

Root cause: the `webkitdirectory`/`directory` attributes on the directory `<input>` were not being applied reliably, causing browsers to reject folder selection.

### 2. Auto-generated changelog overwrites manual user edits

When files include a README, the frontend fires an async request to auto-generate a changelog preview. If the user starts typing in the changelog textarea before the server response arrives, the async result silently replaces their input. Similarly, removing the README while a request is in flight does not cancel it — the stale result still gets applied.

### 3. Backend changelog generation uses skill history instead of package history

`packages.generateChangelogPreview` calls the skill-specific helper from `lib/changelog.ts`, which queries the `skills` table for previous versions. For packages/plugins there is no matching skill, so the changelog always treats every upload as an "initial release" with no diff context — and fallback text reads "Updated skill." instead of something package-appropriate.

## Expected behavior

- Selecting a plugin folder in the file picker should work; files should appear in the upload list.
- Manual changelog edits must never be overwritten by async auto-generation.
- Package changelog preview should diff against the previous *package* release, not skill versions.

## Fixed in

PR #1287

TurboTheTurtle:
Closes #1288.

Adds README-based package changelog previews for plugin release publishes and guards the async preview path so stale generated text cannot overwrite user edits. Package changelog preview now verifies the caller can publish/manage the existing package before reading previous release storage; direct owners, publisher org members, moderators, and admins are allowed, while unrelated signed-in users are denied before storage is read.

Behavior proof:
```text
Head: 0c6c02c51e49

Real local-auth browser run:
- Disposable local Convex seeded a package with prior README-backed releases.
- Chromium opened the real /plugins/publish release flow and uploaded a new package folder.
- The changelog textarea auto-filled from package history:
  "- added 3 file(s).\n- Updated README and package contents."
- After editing the textarea to "Manual release notes after preview.", the manual text remained after the field settled.
Result: 1 passed (chromium).

Focused regression proof:
- Denied existing-package preview throws before release storage is read; storage.get was not called.
- Direct owner, publisher org owner/admin/publisher, platform admin, and platform moderator access paths pass.
- Delayed route preview cannot overwrite a manually edited changelog.
Result: 11 targeted tests passed; changed test-file set also passed with 3 files / 316 tests.
```

CI state: GitHub Actions checks on this head are green for packages, e2e-http, static, types-build, unit, Playwright local-auth, CodeQL, CodeQL Light, secret scan, and pr-gates. The remaining Vercel status is the OpenClaw Foundation team authorization gate, not a source/test failure.


TurboTheTurtle:
@clawsweeper review

TurboTheTurtle:
@openclaw-mantis visual task: verify plugin release publish auto-fills a README-based changelog and does not overwrite manual edits after the preview returns.

TurboTheTurtle:
@clawsweeper re-review

TurboTheTurtle:
@clawsweeper re-review

Added real behavior proof to the PR body for head `0c6c02c51e49`.

Key proof excerpts:
```text
Real local-auth browser run:
- Disposable local Convex seeded a package with prior README-backed releases.
- Chromium opened the real /plugins/publish release flow and uploaded a new package folder.
- The changelog textarea auto-filled from package history:
  "- added 3 file(s).\n- Updated README and package contents."
- After editing the textarea to "Manual release notes after preview.", the manual text remained after the field settled.
Result: 1 passed (chromium).

Focused regression proof:
- Denied existing-package preview throws before release storage is read; storage.get was not called.
- Direct owner, publisher org owner/admin/publisher, platform admin, and platform moderator access paths pass.
- Delayed route preview cannot overwrite a manually edited changelog.
Result: 11 targeted tests passed; changed test-file set also passed with 3 files / 316 tests.
```

TurboTheTurtle:
@clawsweeper re-review

TurboTheTurtle:
Merged main to clear the conflict and fixed the package changelog prompt labels so package previews refer to README content instead of SKILL.md.

Focused tests passed: `bunx vitest run convex/packages.public.test.ts src/__tests__/plugins-publish-route.test.tsx`.

Patrick-Erichsen:
@TurboTheTurtle mind attaching some visual proof here?

Thanks for contributing though! Appreciate all the work you've been doing.

convex/packages.ts
```diff
@@ -8878,6 +8902,32 @@ export const finalizePackagePublishAttemptInternal = internalAction({
...
+    family: v.union(v.literal("code-plugin"), v.literal("bundle-plugin")),
+    version: v.string(),
+    readmeText: v.string(),
+    filePaths: v.optional(v.array(v.string())),
+  },
+  handler: async (ctx, args) => {
+    const { userId } = await requireUserFromAction(ctx);
+    const name = normalizePackageName(args.name);
+    const version = assertPackageVersion(args.family, args.version);
+    await runQueryRef(ctx, internalRefs.packages.assertCanGenerateChangelogPreviewInternal, {
```
Patrick-Erichsen: [P1] Bind this authorization to the exact release that is later read. This query returns only `{ ok: true }`; `generatePackageChangelogPreview` then re-resolves the package by name and its `latestReleaseId` in later `runQuery` transactions before `storage.get`. If the name is absent here and created before the next query, or ownership/latest-release changes between queries, the action can read retained README storage that this check never authorized. Have the gate resolve and return the authorized prior `releaseId | null`, then read only that immutable ID (and derive the actor from auth inside the authorization function per the Convex guidance). Add a regression for the name/owner/latest-release transition.

src/routes/plugins/publish.tsx
```diff
@@ -497,6 +516,60 @@ export function PublishPluginRoute() {
...
   }, [existing]);
 
+  useEffect(() => {
+    if (!showChangelogField) return;
+    if (isMetadataLocked) return;
+    if (changelogTouchedRef.current) return;
+    if (changelog.trim()) return;
+    const packageName = name.trim();
+    const packageVersion = version.trim();
+    if (!packageName || !semver.valid(packageVersion)) return;
```
Patrick-Erichsen, line 532: [P2] Cancel the outstanding request before these validation returns. Today the request token increments only after name/version/README validation succeeds. If a preview for package A is in flight and the user temporarily clears or invalidates the name/version, this effect returns here without invalidating A; A can then populate the textarea, and the non-empty changelog prevents a new preview when valid package B input is entered. Use an effect cleanup or invalidate on every preview-input change, and cover a delayed preview resolved while the name/version is invalid.

Patrick-Erichsen (changes requested):
This adds README-derived changelog previews to plugin release publishing, with package-history lookup and manual-edit protection. That is the right surface, but the current implementation can still read a release not covered by the authorization result and can apply a preview after its inputs become invalid.

LOC: +590/-7 (6 files)

Findings are inline; both block merge.

Reviewed exact SHA `1b7e15165db4d14b88e27e854aeb1064b2925397` against prior reviewed SHA `bef0ba8e36f8d1d782eb0a3a55f8eb3e58433781` and current main `34b6774848390f47591906569c1a535bdb3c0c23`. The new head is only a main merge. Its one conflict resolution retained both the preview-access tests and main's pagination test. A current-main merge simulation is clean, and main's two later commits do not overlap these six files.

Evidence: all live required CI checks are successful on the exact head; `deploy-test` is skipped and the PR remains `UNSTABLE` because the external preview authorization status is unresolved. Local focused validation passed: `bunx vitest run convex/lib/changelog.test.ts convex/packages.public.test.ts src/__tests__/plugins-publish-route.test.tsx` (3 files, 360 tests). Structured autoreview reported a separate concern about preserving manual notes across a replacement upload; I did not accept its proposed unconditional reset because that would violate the issue's manual-edit preservation contract for a corrected re-selection of the same package.

The requested visual proof is still absent: the PR body contains a text account for old head `0c6c02c51e49`, but no screenshot, video, or linked artifact was attached after the maintainer request. That text and the tests are useful, but they do not establish the visible behavior on this exact head.

Policy decision still required before merge: either (1) explicitly authorize and durably document direct-owner, publisher-org owner/admin/publisher, platform-admin, and platform-moderator access to retained prior-release content, or (2) narrow platform staff access and update the focused tests. I am not inferring that authorization.

Best-fix verdict: plausible layer, incomplete contract. Bind the authorization transaction to the exact prior release later read, and invalidate every outstanding preview when any preview input changes even if the new input fails validation.

Alternatives considered: generating notes only during publish would remove the pre-publish preview; reading prior releases client-side would move the storage boundary to the wrong layer. Both are worse than returning an authorized release ID from the server-side gate.

Code read: `convex/packages.ts`, `convex/lib/changelog.ts`, `convex/lib/publishers.ts`, the three focused test files, `src/routes/plugins/publish.tsx`, sibling skill preview/file-access paths, issue #1288, current main, and the prior durable review.

Remaining uncertainty: no exact-head real-browser artifact and no real Convex race proof were supplied; the transaction gap follows directly from the action's separate `runQuery` calls.

TurboTheTurtle:
Pushed a fix for the preview race issues from review: the server now uses the authorized latest release for changelog history, and the publish form invalidates pending previews when inputs change.

Focused checks pass: `bunx vitest run convex/lib/changelog.test.ts convex/packages.public.test.ts src/__tests__/plugins-publish-route.test.tsx` and `bun run ci:static`.

Patrick-Erichsen:
Thanks @TurboTheTurtle , much appreciated as always!

PR #2948, opened 2026-07-01
feat: add publisher feed model API

giodl73-repo:
## Summary

- expose publisher-only detail and feed routes; ClawHub publishers represent both people and organizations, so there is no parallel accounts API
- reuse the canonical public-publisher visibility resolver and preserve legacy ownerUserId ownership during migration
- publish coherent bounded publisher snapshots with monotonic stored revisions
- page immutable snapshots through opaque publisher/sequence/offset cursors and reject stale or malformed cursors
- fail explicitly with 503 no-store when the bounded snapshot cannot prove completeness instead of silently truncating
- keep publisher entries discovery-only with no install, trust, approval, official, or scan authority

## API

- GET /api/v1/publishers/{publisherId}
- GET /api/v1/publishers/{publisherId}/feed?limit=<1..100>&cursor=<opaque>

All pages for one cursor chain carry the same feedId, sequence, and generatedAt. Publisher feed responses are private, no-store so shared caches cannot bypass stale-cursor checks.

## Review fixes

Addresses Patrick's exact-head review:

- canonical linked/legacy-user visibility is shared with existing public publisher surfaces
- legacy ownerUserId-only skills and packages are merged and deduplicated
- request-time sequence 0 and silent bounded-scan completion are replaced by persisted coherent revisions and explicit capacity failure
- equal-timestamp ordering is resolved before publication, not across a lossy source-page boundary
- origin-relative URLs reject protocol-relative forms, backslashes, and control characters
- summaries are Unicode-safe bounded and snapshot content keys store only SHA-256 digests
- hard publisher deletion removes the stored feed publication

## Validation

- focused schema, projection, HTTP, and OpenAPI tests: 22 passed
- targeted type-aware oxlint
- schema package build
- git diff --check
- full TypeScript check reaches only the existing missing yauzl typings in scripts/staging-seed/snapshotIo.ts
- final Codex review: no discrete correctness issue identified

## Contract dependency

The matching publisher-only and pagination contract is being updated in openclaw/rfcs#39.
## Real behavior proof

Behavior or issue addressed:
The public publisher detail endpoint, bounded publisher-feed pagination, revision refresh, and stale-cursor rejection work against a real Convex runtime. The proof run also exposed and fixed a Convex-only failure caused by multiple paginated queries in one mutation.

Real environment tested:
WSL2 Ubuntu 24.04, exact PR head `2172b6badd2b433e8c305de870a2af47b823be28`, disposable anonymous local Convex deployment, and repository dev fixtures. HTTP actions were exercised over loopback; no shared, staging, or production deployment was targeted.

Exact steps or command run after this patch:

1. `bunx convex dev --once --typecheck=disable --codegen disable`
2. `bunx convex run --no-push devSeed:seedLocalFixtures '{"reset":false}'`
3. Request publisher detail, then request the first feed page with `limit=1` and its opaque continuation cursor.
4. Publish an additional local fixture, refresh the first page, and replay the old cursor.

Evidence after fix:

```text
DETAIL HTTP/1.1 200 OK
publisher={kind:user, handle:local, displayName:"Local Dev"}
feedUrl=/api/v1/publishers/[redacted-local-id]/feed

FIRST_PAGE HTTP/1.1 200 OK
schemaVersion=1 sequence=1 entryCount=1 entryKinds=[plugin] nextCursorPresent=true

CONTINUATION HTTP/1.1 200 OK
sameFeedId=true sameSequence=true sameGeneratedAt=true entryCount=1 nextCursorPresent=true

CONTENT_CHANGE seedPadelSkill=ok
REFRESHED HTTP/1.1 200 OK
oldSequence=1 newSequence=2 sequenceAdvanced=true entryCount=1

STALE_CURSOR HTTP/1.1 409 Conflict
cache-control: no-store
Publisher feed cursor is stale; restart from the first page
```

Observed result after fix:
Publisher detail and both feed pages returned 200. The continuation preserved feed identity, sequence, and generation time. A real content change advanced the stored revision from 1 to 2, and replaying the previous cursor returned 409 with `Cache-Control: no-store`.

What was not tested:
No public preview, staging, or production deployment and no production data. This proof covers the anonymous public HTTP contract on the exact branch code in a disposable real Convex runtime.


giodl73-repo:
@clawsweeper re-review

giodl73-repo:
@clawsweeper re-review

giodl73-repo:
@clawsweeper re-review

convex/accountFeeds.ts
```diff
@@ -0,0 +1,277 @@
...
+      feedUrl: `/api/v1/accounts/${encodeURIComponent(String(user._id))}/feed`,
+    };
+  },
+});
+
+export const getPublisherDetail = internalQuery({
+  args: { publisherId: v.string() },
+  handler: async (ctx, args) => {
+    const publisher = await safeGetPublisher(ctx, args.publisherId);
+    if (!isActivePublisher(publisher)) return null;
```
Patrick-Erichsen: [P1] Apply the canonical public-publisher visibility check here and in `getPublisherFeed`. An active publisher row alone is not public for a personal publisher: `getPublicPublisherVisibility` hides it when the linked or legacy owner user is missing, deleted, or deactivated. This endpoint still returns the publisher, and the feed at line 264 still publishes its entries; the new test explicitly blesses that state by nulling only `account`. A ban/deactivation desync therefore bypasses the visibility invariant used by profile and published-item APIs. Move/reuse that shared resolver and return no detail/feed for an invisible personal publisher.

convex/accountFeeds.ts
```diff
@@ -0,0 +1,277 @@
...
+});
+
+export const getAccountFeed = internalQuery({
+  args: {
+    accountId: v.string(),
+    limit: v.optional(v.number()),
+  },
+  handler: async (ctx, args) => {
+    const user = await safeGetUser(ctx, args.accountId);
+    if (!isActiveUser(user) || !user.personalPublisherId) return null;
```
Patrick-Erichsen: [P2] Preserve the repository's legacy publisher ownership paths. `users.personalPublisherId` is still optional, current reads fall back through `publishers.by_linked_user`, and the production ownership repair is explicitly a targeted per-user canary. This returns 404 when the copied pointer is absent/stale; the collectors likewise query only `ownerPublisherId`, omitting otherwise-public `ownerUserId` rows. Resolve the personal publisher through the existing fallback and merge/dedupe legacy owner-user rows, with regression tests for missing/stale pointers and ownerUserId-only content.

convex/accountFeeds.ts
```diff
@@ -0,0 +1,277 @@
...
+    schemaVersion: ACCOUNT_FEED_SCHEMA_VERSION,
+    feedId: accountFeedId(params.scope, params.stableId),
+    scope: params.scope,
+    accountId: params.user ? String(params.user._id) : null,
+    publisherId: String(params.publisher._id),
+    handle: params.publisher.handle ?? params.user?.handle ?? null,
+    displayName:
+      params.publisher.displayName || params.user?.displayName || params.user?.name || "",
+    generatedAt: new Date().toISOString(),
+    sequence: 0,
```
Patrick-Erichsen: [P1] Do not publish placeholder revision state as the stable v1 contract. `generatedAt` changes on every request while every distinct body reuses sequence `0`; the feed is also sliced/scan-capped with `nextCursor: null`, and the HTTP layer rejects all cursors. RFC 0009 PR #39 requires a monotonic logical revision plus opaque deterministic continuation. As written, clients cannot distinguish updates/rollback and popular publishers are silently reported as complete while older public entries are unreachable. Persist a coherent revision/cursor model, or keep these endpoints/schema explicitly experimental until that exists.

convex/accountFeeds.ts
```diff
@@ -0,0 +1,277 @@
...
+  let pagesRead = 0;
+
+  while (!isDone && pagesRead < ACCOUNT_FEED_MAX_SOURCE_PAGES && entries.length < limit) {
+    const page = await ctx.db
+      .query("skills")
+      .withIndex("by_owner_publisher_active_updated", (q) =>
+        q.eq("ownerPublisherId", publisher._id).eq("softDeletedAt", undefined),
+      )
+      .order("desc")
+      .paginate({ cursor, numItems: limit });
```
Patrick-Erichsen: [P2] The promised stable ID tie-break is not honored at the source-page boundary. Each collector requests only `limit` rows ordered by the Convex index, then applies the kind/ID tie-break later. If two rows share `updatedAt` and the lexicographically earlier ID falls on the next source page, it is never considered, so the first page can change with insertion/index order despite the documented ordering contract. Add a `limit: 1` two-page equal-timestamp regression and select enough boundary ties, or encode the full tie-break into the pagination key.

packages/schema/src/accountFeed.ts
```diff
@@ -0,0 +1,84 @@
...
+    throw new Error("Account feed generatedAt must be a valid ISO date");
+  }
+  for (const entry of feed.entries) {
+    if (!entry.id || !entry.name || !entry.displayName) {
+      throw new Error("Account feed entry identity fields must be non-empty");
+    }
+    if (!Number.isFinite(entry.updatedAt) || entry.updatedAt < 0) {
+      throw new Error("Account feed entry updatedAt must be a non-negative finite number");
+    }
+    if (entry.url.startsWith("/")) {
```
Patrick-Erichsen, line 69: [P2] Reject backslashes before accepting an origin-relative URL. For example, `/\\evil.example/x` passes this branch because it starts with one slash, but WHATWG resolution against `https://clawhub.ai/...` produces `https://evil.example/x`. Since this parser is the consumer boundary for unsigned discovery references, validate by resolving against a fixed origin and requiring the same origin/path form, or reject backslashes and control characters explicitly.

Patrick-Erichsen (changes requested):
This adds public account/publisher identity endpoints plus bounded discovery feeds for public skills and plugins. The discovery-only authority boundary is sound: entries contain no install candidate, integrity, official, approval, or trust fields, and downstream installation still has to resolve through the catalog path.

LOC: +1882/-77 (22 files)

The inline findings block this exact head. The most important issue is that the new endpoints bypass ClawHub's established public-publisher visibility rule, so a personal publisher linked to an inactive account can become discoverable here even though existing profile/catalog surfaces hide it.

Best-fix verdict: too narrow. The API should reuse one shared public identity/ownership resolver, preserve legacy ownership during the migration window, and either implement durable revision/pagination semantics or keep this projection experimental until the v1 contract is real.

Alternatives considered: relying on account-deletion cascades or the targeted ownership repair was rejected because current code deliberately keeps defensive visibility checks and optional legacy fields. Shipping sequence `0` / `nextCursor: null` as stable v1 was rejected because RFC 0009 PR #39 already requires monotonic logical revisions and opaque deterministic continuation.

Code read: `convex/accountFeeds.ts`, `convex/httpApiV1/accountFeedsV1.ts`, `convex/lib/public.ts`, `convex/lib/publishers.ts`, `convex/publishers.ts`, `convex/maintenance.ts`, `convex/catalogFeed.ts`, package/skill public-read helpers, schema/OpenAPI/tests, `specs/account-feeds.md`, `specs/orgs.md`, and RFC 0009 account-feed addendum at PR #39 head.

Validation: the focused 23 tests pass locally. Exact-head static, unit, types-build, packages, e2e-http, CodeQL, and Playwright smoke checks pass. The head is not fully green: secret scanning failed because the runner could not pull the TruffleHog image, moderation-star timed out amid Convex query timeouts, and Vercel requires contributor authorization. No live Convex/HTTP behavior proof was attached.

Contract dependency: RFC PR #39 is still open, so the public v1 release should be coordinated with that contract rather than claiming conformance to an unmerged rewrite.

Remaining uncertainty: I did not deploy this branch to a live Convex instance; runtime evidence is limited to exact-head CI plus focused local tests and source-level repros.

giodl73-repo:
Patrick: the live Convex proof found and fixed a runtime-only multiple-pagination failure. Exact head 2172b6ba now has redacted detail, first-page, continuation, revision-advance, and stale-cursor proof in the PR body. Could you re-review this head and confirm whether ClawHub should own the unsigned publisher discovery API as the supported public contract?

@clawsweeper re-review

PR #2950, opened 2026-07-01
feat: add private publisher follow graph

giodl73-repo:
## Summary

- adds private, idempotent publisher follow/unfollow storage for the authenticated user
- lists the current user's followed publishers with opaque cursor pagination
- stores a reversible `all` / `none` delivery preference for Inbox projections
- uses canonical publisher visibility rules and removes follow edges during account or publisher deletion in bounded, resumable batches
- deliberately exposes no public followers/following graph

## API

- `GET|POST|DELETE /api/v1/publisher-follows`
- authenticated and account-scoped; no public social-graph endpoints

## Product boundary

Following is discovery state only. It does not grant official status, registry review, source trust, scan success, local approval, or install authority. The preference controls delivery projections; the follow remains available to a future chronological Following feed.

## Stack

Depends on #2948. Followed publisher activity is added by #2958. The branch is commit-stacked because cross-fork PRs cannot select another fork branch as their base.

## Validation

- focused follow, HTTP, publisher deletion, and account deletion suites
- bounded cleanup coverage for both follower and publisher edges
- schema package build and generated route contract
- `git diff --check`

PR #2957, opened 2026-07-03
feat: add publisher follow controls and discovery

giodl73-repo:
## Summary

- Adds Follow/Following controls to publisher profiles with GitHub sign-in gating and stable publisher-ownership self-follow suppression.
- Adds a Creators > Following discovery tab backed by the publisher-only follow graph from #2950.
- Uses the normalized `nextCursor` contract, hides the irrelevant grid/list toggle in Following, and avoids flashing signed-out content while auth is loading.
- Removes notification language; following affects discovery, while activity delivery is handled by the timeline slice.

## Dependency

Depends on #2950. This branch is restacked on the corrected public-social-graph implementation.

## Boundary

Following does not mark a publisher, artifact, feed, or install path official, reviewed, trusted, or installable.

## Validation

- `bun x vitest run src/__tests__/creators-route.test.tsx src/__tests__/user-profile-route.test.tsx` (36 passed)
- Type-aware oxlint passes for the changed routes and tests.
- `oxfmt` and `git diff --check` pass.


convex/lib/retentionPolicy.ts
```diff
@@ -209,6 +209,7 @@ export const RETENTION_POLICIES = {
   catalogFeedPublications: permanent("Current published hosted catalog feed snapshot."),
   stars: permanent("User star records."),
   promotions: permanent("Curated promotional offers; ended records stay for launch-page history."),
+  publisherFollows: permanent("User publisher follow preference records."),
```
Patrick-Erichsen, line 213: [P1] Clean follow rows when either identity is hard-deleted. These records are permanent, but `hardDeleteSelfDeletedAccountState` never deletes rows through `by_follower`, and `hardDeletePublisherRows` never deletes rows through `by_publisher`. Account deletion therefore retains the user's private follow graph indefinitely, while publisher deletion leaves dangling follow rows. Please wire both indexed cleanup paths into the existing hard-delete flows and cover them with deletion tests; durable audit logs can remain separate.

src/routes/user/$handle.tsx
```diff
@@ -580,6 +632,34 @@ export function PublisherProfilePage({
     publisher?.kind === "user" &&
     (viewerCanAddToPublisher || (me?.handle != null && me.handle === handle));
   const viewerCanSeeOrgRoles = isAdmin(me) || (me?.handle != null && me.handle === handle);
+  const viewerIsPublisher = me?.handle != null && me.handle === publisher?.handle;
+  const followPublisherId = publisher?._id as Id<"publishers"> | undefined;
+  const followPublisherHandle = publisher?.handle ?? handle;
+  const isFollowingPublisher = useQuery(
+    api.publisherFollows.isFollowingPublisher,
+    isAuthenticated && followPublisherId && !viewerIsPublisher
```
Patrick-Erichsen, line 654: [P2] Use stable publisher ownership for this gate. The repo explicitly supports a personal publisher handle differing from its linked user's handle, but `viewerIsPublisher` only compares handles. In that supported state the owner sees Edit/Add plus Follow on their own publisher and can create a self-follow. Gate both this query and the control with `viewerOwnsPersonalPublisher` or the linked user/personal publisher id, and add the differing-handle case to the route test.

convex/httpApiV1/publisherFollowsV1.ts
```diff
@@ -0,0 +1,161 @@
...
+}
+
+function publisherIdFromPayload(payload: Record<string, unknown>) {
+  const value = typeof payload.publisherId === "string" ? payload.publisherId.trim() : "";
+  return value ? (value as Id<"publishers">) : undefined;
+}
+
+function notificationsFromPayload(payload: Record<string, unknown>) {
+  const value =
+    typeof payload.notifications === "string" ? payload.notifications.trim() : undefined;
```
Patrick-Erichsen: [P2] Reject malformed values when the field is present. A body such as `{"publisherId":"...","notifications":false}` currently treats `false` as omission and creates the follow with the default `all` notification preference. Distinguish an absent property from a present non-string/empty value and return 400 unless the value is exactly `all` or `none`; add cases for null, boolean, numeric, and empty-string inputs.

src/routes/creators/index.tsx
```diff
@@ -341,7 +490,18 @@ function PublishersIndex() {
 
       <div className="browse-layout">
         <div className="browse-results">
-          {highlightedPublishers.length > 0 ? (
+          {followingOnly ? (
```
Patrick-Erichsen, line 493: [P2] The view toggle becomes a no-op on the Following tab. It still updates `view=grid` and shows Grid as selected, but this branch always renders `FollowedPublisherDiscovery` as list markup, so the visible layout never changes. Hide/disable the toggle while Following is active, or pass the selected variant through and implement the grid rendering.

src/routes/creators/index.tsx
```diff
@@ -341,7 +490,18 @@ function PublishersIndex() {
...
+            <FollowedPublisherDiscovery
+              authenticated={isAuthenticated}
```
Patrick-Erichsen, line 495: [P3] Preserve the auth-loading state here. `useAuthStatus()` reports `isAuthenticated=false` while hydration is still loading, so a signed-in visitor to `?following=true` briefly receives the signed-out 'Sign in to see publishers you follow' state. Pass `isLoading` into this component and render its loading placeholder before deciding the viewer is anonymous; cover the transition in the route test.

Patrick-Erichsen (commented):
Deep review of the exact head. The follow-state auth boundary is correctly private, and I found no wording or behavior that treats following as trust, approval, official status, review status, or install authority. The concrete issues are inline. Also verified the already-reported contract blockers remain: `/api/v1/publisher-follows` is absent from `docs/http-api.md` and `public/api/v1/openapi.json`, and `packages/schema/dist/routes.js.map` is stale. Exact-head `pr-gates` is currently red because `src/styles.css` fails `oxfmt --check`.

giodl73-repo:
Superseded by #2958, which now combines publisher follow controls/discovery with the pull-based followed-publisher activity timeline.

PR #2958, opened 2026-07-03
feat: add grouped publisher following activity

giodl73-repo:
## Summary

- records one deduplicated activity event per eligible skill or plugin release
- coalesces repository/commit publication batches so a 50-artifact release is one activity group
- serves presentation-neutral `following` and `inbox` projections over the same grouped substrate
- paginates compact groups per followed publisher and expands a group only on demand
- re-checks current follow, publisher visibility, artifact visibility, exact-release moderation, ownership, and deletion during hydration
- removes granular events and groups in bounded, resumable deletion phases

## Product shape

This PR is backend-only. `following` includes every followed publisher; `inbox` excludes publishers whose delivery preference is muted. The next PR adds private Inbox read state and UI. A future Facebook/X-style Following page can render the same groups chronologically without a second event store or migration.

This lane is independent of signed publisher-feed consumption in OpenClaw and has no trust, install, approval, email, or webhook implication.

## Stack

Depends on #2950 (and transitively #2948). The branch is commit-stacked because cross-fork PRs cannot select another fork branch as their base.

## Scale proof

- 50 skills from one repository commit plus one unrelated plugin produce 2 groups, not 51 inbox rows
- group previews hydrate at most 3 visible artifacts
- group expansion and cleanup remain bounded

## Validation

- focused activity, follows, publication, publisher deletion, account deletion, and package visibility suites: 543 tests passed during restack
- targeted grouped projection suite passes after the final split
- schema package build and `git diff --check`


Patrick-Erichsen (commented):
Reviewed exact head `1a780b6929e9bc16c5b6f113df192befcc5f6192`. LOC: +2039/-1 (19 files). This adds durable private publisher-follow notifications and bounded continuation fanout. I found three actionable blockers; no approval submitted.

Best-fix verdict: **too narrow / not merge-ready**. Keep notifications discovery-only, but derive package eligibility from the inserted release, carry immutable publication time/audience eligibility through fanout, and repair the exact-head test regression. A synchronous unbounded fanout was considered and rejected because it would violate Convex transaction bounds; a durable event row or immutable `eventAt` passed through the existing cursor jobs is the appropriate layer.

Code read: publish callers in `convex/skills.ts` and `convex/packages.ts`; fanout/list/read paths in `convex/publisherNotifications.ts`; follow mutations/indexes in `convex/publisherFollows.ts` and `convex/schema.ts`; retention, HTTP/auth boundaries, tests, current `origin/main`, the local follow spec, and RFC 0009.

Remaining uncertainty: the PR provides no real Convex/API/browser proof, and no frontend currently consumes `publisherNotifications.listMine`.

convex/packages.ts
```diff
@@ -9591,6 +9591,49 @@ export const insertReleaseInternal = internalMutation({
       updatedAt: now,
     });
 
+    const notificationPublisherId = args.ownerPublisherId ?? pkg.ownerPublisherId;
+    const notificationPublisher = notificationPublisherId
+      ? (ownerPublisher ?? (await ctx.db.get(notificationPublisherId)))
+      : null;
+    const notificationScanStatus = shouldPromoteLatest
```
Patrick-Erichsen: [P1] Gate backport notifications on the inserted release

For a non-latest release, this falls back to `pkg.scanStatus`, which describes the existing latest package rather than `releaseId`. A clean latest plus a newly blocked backport therefore satisfies `shouldNotifyPackageFollowers` and advertises content that release visibility/download paths should reject. Derive notification eligibility from the inserted release resolved scan verdict using the same source/helper as public release visibility, and add a blocked non-latest regression test.

convex/publisherNotifications.ts
```diff
@@ -0,0 +1,259 @@
...
+    return { ok: true as const, created: 0, skipped: "publisher_unavailable" as const };
+  }
+
+  const text = buildNotificationText({
+    eventType: args.eventType,
+    publisherDisplayName: publisher.displayName,
+    resourceTitle: args.resourceTitle,
+    version: args.version,
+  });
+  const now = Date.now();
```
Patrick-Erichsen: [P1] Snapshot publication time and audience eligibility

`now` is fanout-job time, while every page queries the current follow graph. A user who follows or unmutes after publication but before a delayed job runs receives the older event, and continuation pages can get different timestamps, so delayed/interleaved publications sort by worker timing rather than publication order. Pass an immutable `eventAt` from the publish mutation through every continuation, constrain eligible follows to the state that existed at that event (or persist an event/audience snapshot), and store that event time for ordering. Add follow/unmute-after-publish and interleaved-page tests.

convex/skills.ts
```diff
@@ -12654,6 +12654,33 @@ export const insertVersion = internalMutation({
       createdAt: now,
     });
 
+    if (isPublicSkillDoc(nextSkill)) {
+      try {
+        await ctx.scheduler.runAfter(
```
Patrick-Erichsen: [P1] Repair the exact-head scheduler regression

This added `runAfter` call makes `convex/skills.rateLimit.test.ts` fail at the existing `expect(runAfter).not.toHaveBeenCalled()` assertion for static-malicious publishes. Exact-head CI run 29430178303 and a local rerun both reproduce the failure. Update that test to identify scheduled functions with `getFunctionName`, assert that `users:autobanMalwareAuthorInternal` is absent, and separately assert the expected notification schedule so the safety invariant remains covered without rejecting all scheduling.

PR #2959, opened 2026-07-03
feat: expose publisher account feed URLs

giodl73-repo:
## Summary

- Add a Feed link to publisher profiles.
- Point it directly at the stable, machine-readable `/api/v1/publishers/{publisherId}/feed` endpoint from #2948.
- Encode the opaque publisher id and cover the URL/helper and rendered profile contract.

## Dependency

Depends on #2948 (account and publisher feed API). This is intentionally a small discoverability child and does not introduce duplicate handle-based feed pages.

## Boundary

The link exposes discovery metadata only. It does not imply official, reviewed, trusted, approved, or installable status.

## Validation

- `bun x vitest run src/lib/ownerRoute.test.ts src/__tests__/user-profile-route.test.tsx` (24 passed)
- `bun x oxfmt --check src/__tests__/user-profile-route.test.tsx src/lib/ownerRoute.test.ts src/lib/ownerRoute.ts 'src/routes/user/$handle.tsx'`
- `bun x oxlint --type-aware src/__tests__/user-profile-route.test.tsx src/lib/ownerRoute.test.ts src/lib/ownerRoute.ts 'src/routes/user/$handle.tsx'`
- `git diff --check`


convex/accountFeeds.ts
```diff
@@ -0,0 +1,277 @@
...
+    schemaVersion: ACCOUNT_FEED_SCHEMA_VERSION,
+    feedId: accountFeedId(params.scope, params.stableId),
+    scope: params.scope,
+    accountId: params.user ? String(params.user._id) : null,
+    publisherId: String(params.publisher._id),
+    handle: params.publisher.handle ?? params.user?.handle ?? null,
+    displayName:
+      params.publisher.displayName || params.user?.displayName || params.user?.name || "",
+    generatedAt: new Date().toISOString(),
+    sequence: 0,
```
Patrick-Erichsen, line 118: [P1] Give each changed feed body a monotonic revision. `generatedAt` and `entries` are recomputed on every request, but this hardcodes `sequence` to `0` forever. Both `specs/account-feeds.md` and RFC #39 define this field as the monotonic revision for cache/replay/rollback checks, and the canonical catalog feed persists and increments its sequence. A client deduplicating by sequence can therefore treat changed feed contents as the same revision. Persist and increment a coherent feed revision, or do not expose this as the stable v1 contract until that state exists.

convex/accountFeeds.ts
```diff
@@ -0,0 +1,277 @@
...
+
+  const sortedCandidates = [...skillEntries, ...packageEntries].sort(
+    (left, right) =>
+      right.updatedAt - left.updatedAt ||
+      left.kind.localeCompare(right.kind) ||
+      left.id.localeCompare(right.id),
+  );
+  const entries = sortedCandidates.slice(0, limit);
+
+  return buildFeed({ scope, stableId, user, publisher, entries, nextCursor: null });
```
Patrick-Erichsen, line 201: [P1] Do not report bounded-scan exhaustion as a terminal feed. Both source collectors stop after three pages, but this call always emits `nextCursor: null`; `convex/accountFeeds.test.ts` already demonstrates that an older public package can exist while the response is an empty, apparently complete feed. Return an opaque continuation or an explicit bounded-scan error/diagnostic (or scan to the true end) instead of a successful terminal response that silently drops discoverable entries.

Patrick-Erichsen (changes requested):
This branch adds a canonical, encoded Feed link to publisher profiles, but because it still targets `main` it also publishes the full account-feed API from #2948. The link helper itself is sound; the included v1 feed semantics have two blocking contract gaps.

LOC: +1910/-78 (26 files)

**Best-fix verdict:** too early in the stack. Fix the revision and bounded-pagination semantics in #2948, then keep #2959 as the small profile-discoverability child.

**Alternatives considered:** Hiding only the profile link does not fix the public routes already included from #2948. Documenting this as a latest-items projection does not fix changing bodies under one revision or successful terminal responses that can omit all older public entries.

**Evidence checked:** current `origin/main` (`3134a204`), clean merge tree, #2948 ancestry, RFC #39 head `92846d1f`, schema/spec/OpenAPI paths, account-feed collectors and tests, HTTP/cache routing, catalog-feed publication behavior, both profile routes, and the 47-test focused suite. Exact-head static/unit/types/packages/e2e/smoke/security checks pass; the unrelated `inspector-version` local-auth shard fails during fixture/bootstrap setup.

**Remaining uncertainty:** no real-browser/live-Convex proof of the new link and API was supplied, but that does not affect either deterministic blocker below.

giodl73-repo:
Closing this raw publisher-feed link surface rather than carrying another review unit. The preferred consumption path is the signed publisher-follow workflow tracked by openclaw/openclaw#109397 and its gateway/UI follow-ups; we can redesign explicit feed-link discovery later if real demand remains.

PR #2969, opened 2026-07-06
fix: add exact and paginated prefix skill discovery

### LINKED ISSUE #1383: Feature request: add prefix/list-style skill discovery to ClawHub CLI search
## Summary

ClawHub search is good for relevance-based lookup, but it is currently weak for registry-style discovery.

Right now, `openclaw skills search <query>` behaves like a ranked search, not like a complete listing mechanism. That makes it difficult to answer simple questions such as:

- which skills exist under a shared prefix?
- how many registry skills start with `aigroup-`?
- what has been published under a given namespace or family?

## Problem

For example, running:

```bash
openclaw skills search aigroup --json
```

may return only the most relevant result, not a reliable enumeration of all matching skills in the registry.

This is fine for “find me the best match”, but not for “show me all skills in this family”.

In practice, this makes it awkward to discover related skills, verify published registries, or count namespace/grouped packages from the CLI.

## Expected behavior

There should be a way to reliably enumerate matching registry skills, especially by prefix or exact slug.

Examples of expected workflows:

```bash
openclaw skills search --prefix aigroup-
openclaw skills search --exact aigroup-market-mcp
openclaw skills list --source hub
openclaw skills list --source hub | grep '^aigroup-'
```

## Actual behavior

Current behavior appears to be relevance search only.

That means:
- searching a prefix does not reliably return all matching skills
- CLI users cannot confidently use `search` for namespace discovery
- counting skills in a family requires manual candidate lists plus one-by-one validation

## Why this matters

This would improve several common workflows:

- publisher / namespace discovery
- registry browsing from terminal-only environments
- verifying whether a family of skills has been published
- comparing local workspace skills vs registry skills
- reducing false conclusions from partial search results

## Suggested improvements

Any of the following would help a lot:

- add `--prefix`
- add `--exact`
- add `skills list --source hub`
- add pagination / full listing mode for registry search
- support filters such as owner, publisher, tag, or namespace

## Example use case

A user wants to know how many skills exist in the registry with the prefix `aigroup-`.

Today, `openclaw skills search aigroup` is not reliable for this, because it returns ranked results rather than a complete list. The workaround is to keep a local candidate list and validate each slug one by one, which is inconvenient and error-prone.

## Request

Please add a registry-oriented discovery path in the CLI, so users can do exact or prefix-based skill enumeration instead of relying only on ranked search.

deepujain:
## Summary
- Add deterministic `clawhub search --exact` matching across canonical skill sources.
- Keep `clawhub search --prefix` as the CLI affordance, but route it through the cursor-paginated `/api/v1/skills?prefix=...` listing contract.
- Return prefix pages in ascending slug order with a query-bound `nextCursor`; `--cursor` continues without a silent 200-result completeness cap.
- Validate exact mode, prefix input, sort compatibility, and CLI flag combinations.
- Document the API and CLI contracts.

Closes #1383.

## Runtime proof
- Pushed the branch functions to the local Convex deployment at `http://127.0.0.1:3210`.
- `skills:listPublicApiPageV1` with `{"prefix":"a","numItems":2}` returned two ascending `a*` slugs and a `skillprefix:` continuation cursor.
- Focused tests prove the next page resumes from the full index key and that a cursor cannot be reused under a different prefix.

## Validation
- `bunx vitest run convex/search.test.ts convex/skills.publicListCursor.test.ts convex/httpApiV1.handlers.test.ts` (538 passed)
- `../../node_modules/.bin/vitest run -c vitest.config.ts src/cli/commands/skills.test.ts` (91 passed)
- `bun run ci:static`
- `bun run ci:unit` (5,631 passed, 2 skipped)
- `bun run ci:packages`
- `bun run ci:types-build`
- `bun run ci:e2e-http`
- `bunx convex dev --once`
- `.agents/skills/autoreview/scripts/autoreview --mode branch --base origin/main` (clean)


deepujain:
@clawsweeper re-review

Addressed the review blocker by adding docs for the new CLI/API modes and updating the PR body with a redacted behavior-proof transcript for `clawhub search --prefix`, `clawhub search --exact`, and the emitted `/api/v1/search?...&mode=` requests.

Patrick-Erichsen:
Thanks @deepujain for the original implementation and for your patience while we updated the search contract and resolved the conflicts. This is merged!

Issue #2984, opened 2026-07-07
[bug] 10 skills stuck in pending_checks after staged prepublication pipeline deployment


zrr000212-netizen:
## What happened

After #2979 and #2981 were deployed, 10 skills from @huaweiclouddev are stuck in `pending_checks` and never get created on ClawHub.

## Repro

Both HTTP API and CLI publish return the same result:
```
POST /api/v1/skills → 200 {ok:true, status:"pending", attemptId:"zx7a0a8..."}
GET /api/v1/skills/{slug} → 404 (skill never created)
```

## Affected skills (all @huaweiclouddev, from huaweicloud/huaweicloud-skills repo)

| # | Skill | Files | Status |
|---|-------|-------|--------|
| 1 | huawei-cloud-ecs-passwordless-login | 5 | pending→404 |
| 2 | huawei-cloud-ecs-sqlbot-deploy | 12 | pending→404 |
| 3 | huawei-cloud-iam-query | 112 | pending→404 |
| 4 | huawei-cloud-monitoring-query | 39 | pending→404 |
| 5 | huawei-cloud-msmodelslim-model-adapt | 25 | pending→404 |
| 6 | huawei-cloud-msmodelslim-model-analysis | 5 | pending→404 |
| 7 | huawei-cloud-msot-msopprof-operator-profiler | 7 | pending→404 |
| 8 | huawei-cloud-network-query | 228 | pending→404 |
| 9 | huawei-cloud-storage-query | 96 | pending→404 |
| 10 | huawei-cloud-vod-collector | 14 | pending→404 |

Known attemptId: `zx7a0a8n29ae55g3kj0d5qj33h8a2a3k` (ecs-passwordless-login)

## Investigation so far

1. **Content validation**: All 10 skills pass `clawhub sync --dry-run` locally. SKILL.md has valid YAML frontmatter, UTF-8 encoding, no CRLF, no binary files, name matches slug. Content quality matches 46 successfully-published skills under the same owner.

2. **Timing**: #2979 merged Jul 6 20:54 UTC, #2981 merged Jul 7 00:47 UTC. One skill (huawei-cloud-dws-sql-check) was successfully published at Jul 7 03:40 UTC by the monitoring script — likely before `CLAWHUB_STAGED_PREPUBLICATION_PUBLISHES=1` was turned on in production.

[...]

Issue #2983, opened 2026-07-07
[bug] Cannot add a version to a skill: API rejects with "skillId: invalid value; versionId: invalid value" on every publish path


Jason-Vaughan:
## What happened

We published a new skill (`decomtangle`, owner `jason-vaughan`) and cannot publish any subsequent version. Every attempt fails with:

```
API response: skillId: invalid value; versionId: invalid value
```

## Repro / what we tried

1. Initial publish (CLI 0.18.0): `clawhub publish <folder> --version 0.1.0 --slug decomtangle` — succeeded. Note we had to pass `--version` explicitly because the CLI reported "--version must be valid semver" despite valid `version: 0.1.0` frontmatter (our `description:` uses a folded block scalar `>`, which the frontmatter parser may not handle — possibly related).
2. Version bump attempts, all failing with the identical error:
   - `clawhub publish <folder> --version 0.1.1 --slug decomtangle` (legacy alias)
   - `clawhub skill publish <folder> --version 0.1.1`
   - `clawhub sync` (workdir layout)
   - Upgraded CLI 0.18.0 → 0.23.1: same error on all paths.
   - `clawhub delete decomtangle --yes` (soft-delete, slug reserved), then fresh `skill publish --version 0.1.1`: same error — the reserved/soft-deleted record appears to still be resolved and rejected.
   - `undelete` + publish as 0.1.2: same error.
3. The listing itself is healthy throughout: `clawhub inspect decomtangle` shows Latest 0.1.0, Moderation CLEAN.

## Expected

Publishing a new version of an owned skill (or at minimum a fresh publish after soft-delete) should succeed or return an actionable validation message.

## Guess at root cause

The server-side version-add path seems to fail resolving/validating the existing skill/version records — possibly a malformed version record created by the initial `--version`-flag publish, given the frontmatter-parsing quirk in (1). Happy to provide the exact folder contents or timestamps if useful.

Environment: CLI 0.18.0 and 0.23.1, macOS, authenticated as `Jason-Vaughan`.

smyx-sunjinhui:
we also get the problem same as you
root@arkclaw:~/.openclaw/workspace/dist# clawhub publish --name "Intelligent Video Search & Retrieval Analysis Skill | 视频搜索检索智能分析技能" --slug smyx-video-search-analysis --version 1.0.6 --tags latest /root/.openclaw/workspace/dist/smyx_video_search_analysis
✖ API response: skillId: invalid value; versionId: invalid value
[...]

PR #2979, opened 2026-07-06
feat: stage publishes behind prepublication checks

Patrick-Erichsen:
## Summary
- stage skill and plugin publishes as private `publishAttempts` when `CLAWHUB_STAGED_PREPUBLICATION_PUBLISHES=1`
- add worker-facing TruffleHog/ClawScan claim + completion flow that finalizes publicly only after both checks pass
- keep secret-positive attempts private and blocked, delete their uploaded storage blobs, and add local-auth e2e proof for skill/plugin happy paths plus secret-positive skill blocking
- preserve VirusTotal as post-finalization telemetry; it is not part of the pre-publication gate

## Validation
- `bun run ci:static`
- `bunx vitest run convex/publishAttempts.test.ts convex/lib/skillPublish.test.ts convex/packages.public.test.ts src/__tests__/plugins-publish-route.test.tsx src/__tests__/skills-publish-route.test.tsx` — 352 passed
- `bun run ci:unit` — 4403 passed, 1 skipped
- `bun run ci:types-build`
- `.agents/skills/autoreview/scripts/autoreview --mode branch --base origin/main` — clean, no accepted/actionable findings
- `PLAYWRIGHT_LOCAL_AUTH_CONVEX_URL=http://127.0.0.1:3410 PLAYWRIGHT_LOCAL_AUTH_CONVEX_SITE_URL=http://127.0.0.1:3411 PLAYWRIGHT_PORT=3300 bun run test:pw:local-auth -- --project=chromium e2e/local-auth/publish-skill-lifecycle.pw.test.ts e2e/local-auth/plugin-inspector-findings.pw.test.ts --grep "skill publishers can create|mocked TruffleHog|plugin publish stays private"` — 3 passed (3.6m)

Refs CLAW-467.

PR #2985, opened 2026-07-07
fix: add emergency bypass for staged prepublication when worker is unavailable

zrr000212-netizen:
## Problem

`CLAWHUB_STAGED_PREPUBLICATION_PUBLISHES=1` in production routes ALL publishes through async staging (TruffleHog + ClawScan). When the prepublication worker is not processing attempts — due to backlog, transient infra failure, or deployment ordering — publishers have zero recourse. Skills that would pass all checks are stuck in `pending_checks` indefinitely.

This is NOT about skipping legitimate checks. It is about providing an emergency escape hatch when the async pipeline itself is the bottleneck, not the content.

Affected: 10 skills from @huaweiclouddev that pass all local validation but are stuck in `pending_checks` because the worker hasn"t claimed them (#2984).

## Solution

Add optional `stagePrePublicationChecks` boolean to the publish payload. When explicitly `false`, bypasses the async staging pipeline and uses the synchronous `insertVersion` path. Omitted/`true` uses existing server-side default — staging remains ON by default for everyone.

This is an **emergency bypass**, not a recommended workflow. Publishers who use it are still subject to VirusTotal (post-publish) and moderation — they just skip the pre-publication TruffleHog + ClawScan gate when that gate is non-functional.

## Changes (5 files, +16 lines)

- `packages/schema/src/schemas.ts` — add to `CliPublishRequestSchema`
- `packages/clawhub/src/schema/schemas.ts` — same
- `convex/httpApiV1/shared.ts` — `parsePublishBody` + `parseMultipartPublish`
- `convex/httpApiV1/skillsV1.ts` — `publishSkillPayloadForApiUser`
- `convex/httpApi.ts` — CLI handler + `parsePublishBody`

## Related

- #2979 — introduced staged prepublication
- #2981 — prepublication worker
- #2983 — CLI schema mismatch from pending response
- #2984 — 10 skills stuck despite passing all checks

### DIFF
--- a/convex/httpApi.ts
+++ b/convex/httpApi.ts
@@ -226,7 +226,7 @@ async function cliPublishHandler(ctx: ActionCtx, request: Request) {
     if (!hasAcceptedLegacyLicenseTerms(args.acceptLicenseTerms)) {
       return text("MIT-0 license terms must be accepted to publish skills", 400);
     }
-    const { ownerHandle, sourceOwnerHandle, migrateOwner, ...publishPayload } = args;
+    const { ownerHandle, sourceOwnerHandle, migrateOwner, stagePrePublicationChecks, ...publishPayload } = args;
     const target = ownerHandle
       ? ((await ctx.runMutation(internal.publishers.resolvePublishTargetForUserInternal, {
           actorUserId: userId,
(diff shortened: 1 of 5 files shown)


Patrick-Erichsen:
Thanks for the quick incident-response proposal. Closing this as not planned.

Production recovery used the server-controlled rollback flag, paused the worker, reconciled and drained the queue, and removed the temporary repair tooling. Allowing ordinary publishers to send `stagePrePublicationChecks: false` would put the emergency control on the wrong side of the security boundary and let callers bypass TruffleHog and ClawScan.

The durable async hardening and eventual controlled re-enable remain tracked in #2990 / CLAW-480.

PR #3005, opened 2026-07-08
feat(feeds): add dormant catalog signing foundation

giodl73-repo:
## Summary

- add an **optional** DSSE/Ed25519 representation for the stored `clawhub-official` catalog publication
- preserve the existing unsigned JSON representation as the default on `/api/v1/feeds/plugins`
- sign the exact immutable stored payload bytes and expose representation-specific validators
- provide a tested prepare/activate rotation script and operator runbook

## Merge and activation contract

**Merging this PR does not enable signing and does not require production keys.**

The route selects DSSE only when a client explicitly sends `Accept: application/vnd.dsse+json`. Without that opt-in, the existing unsigned response remains available even when `CLAWHUB_FEED_SIGNING_CONFIG` is missing or malformed. An opted-in signed request fails closed with `503 no-store` until a valid signer is installed.

This is the dormant signing foundation for the large-feed stack. Production activation is a later operator action after the matching OpenClaw trust anchor is released.

## Review order

1. **This PR — signing foundation and activation boundary**
2. openclaw/clawhub#3149 — durable revision/change history
3. openclaw/clawhub#3160 — signed paged query and delta transport
4. openclaw/clawhub#3163 — sharded full snapshots
5. openclaw/openclaw#110250 — verified sharded-feed consumer

OpenClaw trust-anchor work in openclaw/openclaw#101981 can be reviewed and landed in parallel, but must ship before production signing is activated.

## Signing configuration

ClawHub owns one atomic Convex secret:

- `CLAWHUB_FEED_SIGNING_CONFIG`: strict JSON containing exactly `keyId` and PKCS#8 Ed25519 `privateKey`

OpenClaw receives only the matching public PEM and key id. There is intentionally no feed-adjacent public-key bootstrap endpoint. The complete provisioning, activation, rotation, and emergency-revocation runbook is in `specs/hosted-catalog-feed.md`.

## Cache behavior

The signed representation uses its envelope ETag as the conditional validator and deliberately omits `Last-Modified`; signer rotation changes envelope bytes without changing the stored publication timestamp. The unsigned representation retains its existing validators. Responses include `Vary: Accept`.

## Validation

- `bun x vitest run convex/httpApiV1.catalogFeedSigning.test.ts scripts/provision-catalog-feed-signing-key.test.ts` — 25 tests passed
- regression coverage proves missing or malformed signer config cannot break ordinary unsigned requests
- exact stored bytes and DSSE Ed25519 signatures verified with generated key pairs
- matching ETag returns 304; signed responses omit `Last-Modified`
- `git diff --check origin/main...HEAD`
- direct `codex review --uncommitted` — no actionable findings

`bun run ci:static` now clears the dependency audit after upstream #3234; it currently stops on formatting errors in two files already present on `main` (`.agents/skills/autoreview/CLAUDE.md` and `CLAUDE.md`). Production key provisioning and public-endpoint proof are intentionally deferred to activation.

### DIFF
--- a/convex/httpApiV1/catalogFeedSigning.ts
+++ b/convex/httpApiV1/catalogFeedSigning.ts
@@ -3,6 +3,7 @@ import { internal } from "../_generated/api";
 import type { ActionCtx } from "../_generated/server";
 import { httpAction } from "../functions";
 import {
+  catalogFeedV1Handler,
   catalogFeedResponseHeaders,
   catalogFeedUnavailableResponse,
   matchesEtag,
@@ -189,4 +190,41 @@ export async function signedCatalogFeedV1Handler(
   return new Response(signed.body, { status: 200, headers });
 }
 
-export const signedCatalogFeedV1Http = httpAction(signedCatalogFeedV1Handler);
+function acceptsSignedCatalogFeed(request: Request) {
+  return (request.headers.get("Accept") ?? "").split(",").some((value) => {
+    const [mediaType, ...parameters] = value.split(";").map((part) => part.trim());
+    const disabled = parameters.some((parameter) => /^q=0(?:\.0*)?$/u.test(parameter));
+    return mediaType?.toLowerCase() === "application/vnd.dsse+json" && !disabled;
+  });
+}
+
+function addAcceptVary(response: Response) {
+  const headers = new Headers(response.headers);
+  const vary = headers.get("Vary");
+  const values = new Set(
+    (vary ?? "")
+      .split(",")
+      .map((value) => value.trim())
+      .filter(Boolean),
+  );
+  values.add("Accept");
+  headers.set("Vary", [...values].join(", "));
+  return new Response(response.body, {
+    status: response.status,
+    statusText: response.statusText,
+    headers,
+  });
+}
+
+export async function negotiatedCatalogFeedV1Handler(
+  ctx: ActionCtx,
+  request: Request,
+  env: Record<string, string | undefined> = process.env,
+) {
+  const response = acceptsSignedCatalogFeed(request)
+    ? await signedCatalogFeedV1Handler(ctx, request, env)
+    : await catalogFeedV1Handler(ctx, request);
+  return addAcceptVary(response);
+}
+
+export const signedCatalogFeedV1Http = httpAction(negotiatedCatalogFeedV1Handler);
--- a/convex/httpApiV1/catalogFeedSigning.ts
+++ b/convex/httpApiV1/catalogFeedSigning.ts
@@ -176,6 +176,7 @@ export async function signedCatalogFeedV1Handler(
     catalogFeedResponseHeaders(publication, {
       representationSha256: signed.sha256,
       additionalHeaders: {
+        "Content-Type": "application/vnd.dsse+json; charset=utf-8",
         "X-Catalog-Payload-SHA256": publication.payloadSha256,
         "X-OpenClaw-Feed-Signing-Key-ID": signingConfig.keyId,
       },
--- a/convex/httpApiV1/catalogFeedSigning.ts
+++ b/convex/httpApiV1/catalogFeedSigning.ts
@@ -11,14 +11,14 @@ import {
 export const OPENCLAW_CATALOG_FEED_PAYLOAD_TYPE =
   "openclaw.official-external-plugin-catalog-feed.v1";
 
-type FeedSigningConfig = {
+export type FeedSigningConfig = {
   keyId: string;
   privateKey: CryptoKey;
 };
 
 type SignedFeedEnvelope = {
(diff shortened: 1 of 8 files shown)


specs/hosted-catalog-feed.md
```diff
@@ -140,3 +148,75 @@ schema version.
...
+   clients may fall back to their bundled catalog when they first encounter the
+   envelope; they must not reinterpret it as unsigned feed content.
+8. Securely delete operator-machine private-key files after the approved secret
+   backup and recovery process is complete.
+
+### Normal rotation
+
+1. Generate a new dedicated key and key id.
+2. Bundle the new public key in OpenClaw while the old key remains trusted.
+3. After that trust update is available, update the two Convex signing variables
```
Patrick-Erichsen: [P1] Make the signer pair transition atomic. These are two independently mutable Convex variables, so the documented rotation cannot actually update them "together": after changing either one, the handler can return a cacheable 200 whose `keyId` selects the new public key while the signature was made by the old private key (or vice versa). OpenClaw filters trusted keys by `keyId`, so that envelope is unverifiable, and `s-maxage=300, stale-while-revalidate=86400` can preserve it after the second update. Use one atomic config value, a verified batch update, or an explicit disable-first sequence that cannot expose a mixed pair, and cover the transition in the runbook/tests.

convex/httpApiV1/catalogFeedSigning.ts
```diff
@@ -0,0 +1,169 @@
...
+  const headers = catalogFeedResponseHeaders(publication, {
+    representationSha256: signed.sha256,
+    additionalHeaders: {
+      "X-Catalog-Payload-SHA256": publication.payloadSha256,
+      "X-OpenClaw-Feed-Signing-Key-ID": signingConfig.keyId,
+    },
+  });
+  if (
+    matchesEtag(request, etag) ||
+    (!request.headers.has("if-none-match") && matchesLastModified(request, publication.publishedAt))
```
Patrick-Erichsen: [P2] Make `Last-Modified` change with the signed representation. Rotation changes `signed.body` and its ETag without changing `publication.publishedAt`; a client that sends only the previous `If-Modified-Since` therefore receives 304 and keeps the old envelope even though the selected representation changed. RFC 0009 explicitly permits conditional refresh with either ETag or Last-Modified. Track signer activation/representation time, omit Last-Modified for this route, or atomically publish a fresh representation when the key changes, with a rotation regression test.

Patrick-Erichsen (changes requested):
This revision now signs the canonical stored `clawhub-official` publication bytes, preserves the payload digest/sequence/expiry fields, and matches the merged OpenClaw-specific DSSE/Ed25519 envelope contract. That is the right behavior and trust boundary.

LOC: +462/-33 (5 files)

Two rotation/cache findings remain below.

Best-fix verdict: acceptable architecture, but not merge-ready until signer transitions cannot publish a mixed key pair and all validators describe the selected signed representation.

Alternatives considered: storing the signed envelope alongside each publication would make representation revisions naturally immutable, but it couples publication to signing-secret availability and expands the storage contract. The narrower fix is an atomic signer configuration/activation revision plus rotation-aware validators.

Code/contracts read: ClawHub publication/store/HTTP/proxy paths and tests; current OpenClaw envelope verifier, loader, snapshot rollback path, and 1 MiB response bound; Convex's upstream Ed25519 WebCrypto implementation; DSSE protocol/envelope docs; RFC 0009 corrected trust/feed drafts.

CI: exact-head `pr-gates`, `static`, `unit`, `types-build`, `packages`, and `e2e-http` are green. The only failing status is external Vercel team authorization.

Expected deployment blockers, separate from these findings: the production key and live endpoint proof do not exist yet; OpenClaw's default ClawHub trust/feed-id PR and rotation follow-up remain open; the corrected RFC contract is also still open. Current OpenClaw main also does not yet enforce the signed payload's `expiresAt`, so end-to-end production trust is not complete even after this producer lands.

giodl73-repo:
@Patrick-Erichsen Re-review when convenient: your two requested producer fixes are both present at exact head c62a50640ae8ca60f17e3f7672f8e1f7dd7d2071. Signer id/key activation is atomic, and Last-Modified now varies with the selected signed representation. The substantive static/unit/types/packages/e2e checks are green; the remaining failures are the unrelated moderation-star Playwright lane and Vercel authorization.

giodl73-repo:
@Patrick-Erichsen The current restacked head \15b8dfba\ addresses both requested changes: signer key/id activation is atomic, and signed representation validators no longer reuse a stale Last-Modified value. It also keeps signing dormant behind explicit activation, preserving the unsigned canonical response until rollout. Substantive CI is green; only the external Vercel authorization status remains. Re-review when convenient.\n\n@clawsweeper re-review

Issue #2925, opened 2026-06-28
Proposal: staged web performance audit and SSR auth plan


vyctorbrzezowski:
## Summary

ClawHub already SSRs public content, but auth is client-driven and `/plugins/` deliberately returns loading data from its loader before fetching the catalog after mount. This issue proposes measured, incremental fixes and keeps SSR auth behind an explicit architecture decision.

## Confirmed from the code

- `src/server.ts` streams SSR output but does not establish auth context.
- `src/lib/useAuthStatus.ts` resolves `useConvexAuth()` and then `api.users.me` on the client.
- Header avatar rendering waits for `users.me`. Its later `api.publishers.getMyProfileHandle` query only controls the optional Profile menu item; it does **not** block the avatar.
- `@convex-dev/auth/react` stores tokens in localStorage. SSR cannot read them.
- The installed `@convex-dev/auth` package implements its complete cookie/proxy/refresh/provider flow specifically for Next.js. TanStack Start has no drop-in equivalent here.
- `/plugins/` uses `createPluginsLoadingData()` in the route loader and calls `loadPluginsPageData()` in an effect after mount.
- Skill detail fetches `getBySlug` before `getReadme`. This cannot be directly parallelized because `getReadme` needs the returned latest-version id.
- `/$owner/skills/$slug` has no `pendingComponent`; the router default only shows route pending UI after 1 second.

## Recommended sequence

1. Add a small benchmark covering TTFB, SSR markers, catalog request counts, and browser timings.
2. Replace the `/plugins/` stub with a real loader. Include search `loaderDeps`, propagate the navigation abort signal, sync loader data into pagination state, and prove there is only one catalog request per navigation. View-only search changes should not refetch catalog data.
3. Optionally add the existing `SkillDetailSkeleton` as the skill route `pendingComponent`. Treat this as perceived-UX polish, not a latency improvement.
4. Measure the skill README contribution to TTFB. Keep SSR README if it is acceptable; otherwise defer to the existing client action, accepting that README content will no longer be in the initial response.
5. Measure `MarkdownPreview` before changing its loading boundary; Shiki and Mermaid are already lazy internally.

Do not add `users.session` as a quick win. Combining `users.me` with publisher visibility reads would broaden the avatar query's read/invalidation set and can make the critical path slower.

## Maintainer-gated SSR auth

A safe TanStack-compatible auth adapter must cover the whole contract, not only set cookies:

[...]

PR #3062, opened 2026-07-12
perf: restore SSR plugin catalog loading

vyctorbrzezowski:
## Summary

- restore the proposed `/plugins/` catalog request to the real TanStack route loader after #2586 moved it behind hydration
- key loader invalidation to request-affecting search state, while keeping view changes and search-only client sorting out of the cache key
- forward navigation cancellation and cap catalog loading at five seconds so a stalled backend cannot block SSR indefinitely
- keep this public browse surface explicitly anonymous so request-scoped private package visibility is never serialized into HTML or mixed into pagination
- anchor load-more state to the active loader result, abort stale pagination, and skip the redundant public-count subscription when loader data already has the total

This closes the confirmed `/plugins/` loader item from #2925. The broader SSR auth, skill README, and Markdown rendering work remain explicitly out of scope.

## Validation

- `bunx vitest run src/__tests__/packages-route.test.tsx src/lib/packageApi.test.ts` (100 passed)
- red/green regressions: effective search loader deps and stale pagination both fail against the previous implementation and pass in this PR
- `bun run ci:static`
- `bun run ci:unit` (4,605 passed, 1 skipped; coverage 85.59% statements / 75.35% branches / 86.79% functions / 88.51% lines)
- `bun run ci:types-build`
- `git diff --check`
- real browser proof at `http://localhost:3000/plugins` with the seeded local public corpus: SSR HTML contained all 3 fixture plugins; warm local TTFB was 0.28s; Chromium observed 0 catalog requests after hydration, 1 request for a category navigation, and 0 requests for a list/grid-only navigation

### DIFF
--- a/src/__tests__/packages-route.test.tsx
+++ b/src/__tests__/packages-route.test.tsx
@@ -48,7 +48,13 @@ let loaderDataMock:
 
 vi.mock("@tanstack/react-router", () => ({
   createFileRoute:
-    () => (config: { loader?: unknown; component?: unknown; validateSearch?: unknown }) => ({
+    () =>
+    (config: {
+      loader?: unknown;
+      loaderDeps?: unknown;
+      component?: unknown;
+      validateSearch?: unknown;
+    }) => ({
       __config: config,
       useNavigate: () => navigateMock,
       useSearch: () => searchMock,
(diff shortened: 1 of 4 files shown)


vyctorbrzezowski:
@clawsweeper review

vyctorbrzezowski:
## ClawHub UI Proof

Status: `passed`
Mode: `feature`
Scenario: `Rendered plugin catalog reference; SSR request measurements remain in the PR body`
Provider: `real Chromium against public ClawHub`
Baseline: not run for feature proof.
Candidate: `merged PR #3062 visual reference`

**Public plugin catalog rendered with live catalog data**

Public plugin catalog rendered with live catalog data

Raw proof files: https://github.com/openclaw/clawhub/tree/qa-artifacts/clawhub-ui-proof/pr-3062/ssr-plugin-catalog-reference

PR #3074, opened 2026-07-13
feat: forward package catalog metadata in publish workflow

sergiopesch:
## What Problem This Solves

The ClawHub CLI already accepts `--changelog`, `--categories`, and `--topics` for package publishes, but the official reusable GitHub Actions workflow does not expose them. Trusted OIDC callers therefore cannot set release/catalog metadata without replacing the supported workflow or falling back to a long-lived token.

This surfaced in [`sergiopesch/raspberry-pi-maker#4`](https://github.com/sergiopesch/raspberry-pi-maker/pull/4): a direct CLI workflow could carry the fields, but ClawHub's trusted-publisher verifier rejected that route. [`sergiopesch/raspberry-pi-maker#5`](https://github.com/sergiopesch/raspberry-pi-maker/pull/5) restored the official reusable workflow and left the catalog metadata gap unresolved.

## Why This Change Was Made

- Keep the official reusable workflow as the single supported OIDC publishing path.
- Expose the CLI's existing metadata contract without adding permissions, secrets, or a second publish implementation.
- Preserve every existing caller: all three inputs are optional and omitted values add no CLI flags.
- Keep parsing and validation in the CLI, which already owns category/topic semantics.

## User Impact

Plugin repositories can publish changelog text, category slugs, and topics through the supported secretless workflow. Existing callers that omit the inputs behave exactly as before.

## Summary

- Add optional string inputs for `changelog`, `categories`, and `topics`.
- Forward non-empty values through the existing environment-backed, `shlex`-quoted command builder.
- Strengthen the workflow contract test around the exact input → environment → conditional CLI flag path.
- Add a copy-ready documentation example and explicitly document omitted-input behavior.

## Behavioural Proof

Exact revision: `6f34a65f328f938326c7826463a2311edc325d72`

- [Secretless Raspberry Pi Maker proof run](https://github.com/sergiopesch/raspberry-pi-maker/actions/runs/29866192179)
  - Real reusable-workflow dry run with all three metadata inputs.
  - Real reusable-workflow dry run with all three inputs omitted.
  - Both calls pin the ClawHub PR and Raspberry Pi package by full commit SHA and hardcode `dry_run: true`.
  - The run logs show the supplied inputs and the generated package-publish command; the control omits the three flags.
- [Raspberry Pi Maker caller CI](https://github.com/sergiopesch/raspberry-pi-maker/actions/runs/29866191172)
- [Existing Raspberry Pi Maker package workflow](https://github.com/sergiopesch/raspberry-pi-maker/actions/runs/29866192661)

The temporary evidence harness was [`sergiopesch/raspberry-pi-maker#8`](https://github.com/sergiopesch/raspberry-pi-maker/pull/8). It used no repository secrets and is now closed with its remote branch removed; the immutable run evidence remains available.

## Security / Trust Impact

- [x] No security/trust impact

No permissions or secret handling changed. Inputs cross the workflow boundary as strings, enter the existing Python argument list through environment variables, and are shell-escaped by the existing `shlex.quote` path. The live proof uses only `actions: read`, `contents: read`, and the workflow's existing `id-token: write` permission.

## Data / Deploy Impact

- [x] No data/deploy impact

No schema, migration, configuration, or rollout change. The workflow remains additive and backward-compatible for callers that omit the new inputs.

## Evidence

- `bun run docs:list`
- `bunx vitest run src/__tests__/package-publish-workflow.test.ts` (5 passed)
- Deterministic `ci:static` components: peer checks, generated-doc check, formatting, lint, dead-code checks, and release-workflow action-pin checks
- `bun run ci:unit` (4,873 passed, 1 skipped)
- `bun run ci:packages` (468 passed across package suites)
- Clean structured autoreview of the full branch: no accepted/actionable findings
- [Fork secret scan for the exact PR revision](https://github.com/sergiopesch/clawhub/actions/runs/29865683102)

`bun run ci:static` currently stops at `bun audit` on two newly reported advisories in dependencies unchanged by this PR (`dompurify` GHSA-c2j3-45gr-mqc4 and `fast-uri` GHSA-4c8g-83qw-93j6). The proof run records that audit result and runs every deterministic static component separately.

## Verification

- [ ] `bun run ci:static` (dependency audit blocker documented above)
- [x] Focused tests for touched behavior
- [x] `bun run ci:unit`
- [x] Broader package gate: `bun run ci:packages`
- [x] Real reusable-workflow behavior proof
- [x] Screenshots/recordings: N/A (workflow-only change)

## AI Assistance

This contribution was AI-assisted. The implementation was manually traced through the reusable workflow, CLI option declarations, package payload builder, adjacent tests, GitHub reusable-workflow/OIDC contracts, and a real secretless Raspberry Pi Maker caller run. The author can maintain the resulting three-input workflow extension.

### DIFF
--- a/.github/workflows/package-publish.yml
+++ b/.github/workflows/package-publish.yml
@@ -67,10 +67,20 @@ on:
         description: Optional comma-separated plugin category slugs.
         required: false
         type: string
+      clear_categories:
+        description: Clear existing plugin categories. Cannot be combined with categories.
+        required: false
+        type: boolean
+        default: false
       topics:
         description: Optional comma-separated catalog topics.
         required: false
         type: string
+      clear_topics:
+        description: Clear existing catalog topics. Cannot be combined with topics.
+        required: false
+        type: boolean
+        default: false
       source_repo:
         description: Optional source repo override for local-folder publishes.
         required: false
@@ -326,7 +336,9 @@ jobs:
           INPUT_TAGS: ${{ inputs.tags }}
           INPUT_CHANGELOG: ${{ inputs.changelog }}
           INPUT_CATEGORIES: ${{ inputs.categories }}
+          INPUT_CLEAR_CATEGORIES: ${{ inputs.clear_categories }}
           INPUT_TOPICS: ${{ inputs.topics }}
+          INPUT_CLEAR_TOPICS: ${{ inputs.clear_topics }}
           INPUT_SOURCE_REPO: ${{ inputs.source_repo }}
           INPUT_SOURCE_COMMIT: ${{ inputs.source_commit }}
           INPUT_SOURCE_REF: ${{ inputs.source_ref }}
@@ -495,7 +507,13 @@ jobs:
           tags = os.environ["INPUT_TAGS"].strip()
           changelog = os.environ["INPUT_CHANGELOG"].strip()
           categories = os.environ["INPUT_CATEGORIES"].strip()
+          clear_categories = os.environ["INPUT_CLEAR_CATEGORIES"].strip().lower() == "true"
           topics = os.environ["INPUT_TOPICS"].strip()
+          clear_topics = os.environ["INPUT_CLEAR_TOPICS"].strip().lower() == "true"
+          if categories and clear_categories:
+              raise SystemExit("categories and clear_categories cannot be combined")
+          if topics and clear_topics:
+              raise SystemExit("topics and clear_topics cannot be combined")
           if owner:
               cmd += ["--owner", owner]
           if family:
@@ -513,8 +531,12 @@ jobs:
               cmd += ["--changelog", changelog]
           if categories:
               cmd += ["--categories", categories]
+          elif clear_categories:
+              cmd += ["--categories", ""]
           if topics:
               cmd += ["--topics", topics]
+          elif clear_topics:
+              cmd += ["--topics", ""]
           source_repo = os.environ["INPUT_SOURCE_REPO"].strip()
           source_commit = os.environ["INPUT_SOURCE_COMMIT"].strip()
           source_ref = os.environ["INPUT_SOURCE_REF"].strip()
--- a/.github/workflows/package-publish.yml
+++ b/.github/workflows/package-publish.yml
@@ -59,6 +59,18 @@ on:
         required: false
         type: string
         default: latest
+      changelog:
+        description: Optional release changelog shown on ClawHub.
+        required: false
+        type: string
+      categories:
+        description: Optional comma-separated plugin category slugs.
+        required: false
+        type: string
+      topics:
+        description: Optional comma-separated catalog topics.
+        required: false
+        type: string
       source_repo:
         description: Optional source repo override for local-folder publishes.
         required: false
@@ -312,6 +324,9 @@ jobs:
           INPUT_FAMILY: ${{ inputs.family }}
           INPUT_VERSION: ${{ inputs.version }}
           INPUT_TAGS: ${{ inputs.tags }}
+          INPUT_CHANGELOG: ${{ inputs.changelog }}
+          INPUT_CATEGORIES: ${{ inputs.categories }}
+          INPUT_TOPICS: ${{ inputs.topics }}
           INPUT_SOURCE_REPO: ${{ inputs.source_repo }}
           INPUT_SOURCE_COMMIT: ${{ inputs.source_commit }}
           INPUT_SOURCE_REF: ${{ inputs.source_ref }}
@@ -478,6 +493,9 @@ jobs:
           family = os.environ["INPUT_FAMILY"].strip()
           version = os.environ["INPUT_VERSION"].strip()
           tags = os.environ["INPUT_TAGS"].strip()
+          changelog = os.environ["INPUT_CHANGELOG"].strip()
+          categories = os.environ["INPUT_CATEGORIES"].strip()
+          topics = os.environ["INPUT_TOPICS"].strip()
           if owner:
               cmd += ["--owner", owner]
           if family:
@@ -491,6 +509,12 @@ jobs:
               cmd += ["--version", version]
           if tags:
               cmd += ["--tags", tags]
+          if changelog:
+              cmd += ["--changelog", changelog]
+          if categories:
+              cmd += ["--categories", categories]
+          if topics:
+              cmd += ["--topics", topics]
           source_repo = os.environ["INPUT_SOURCE_REPO"].strip()
           source_commit = os.environ["INPUT_SOURCE_COMMIT"].strip()
           source_ref = os.environ["INPUT_SOURCE_REF"].strip()
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -797,12 +797,12 @@ jobs:
       clawhub_token: ${{ secrets.CLAWHUB_TOKEN }}
 ```
 
-To attach release and catalog metadata, pass the matching CLI values to either
-job:
+To attach release and catalog metadata, add the matching CLI values to the
+job's existing `with` block. Keep `dry_run: true` on pull-request jobs; use
+`dry_run: false` only on the trusted publish job shown above.
 
 ```yaml
 with:
-  dry_run: false
   changelog: "Describe the changes in this release."
   categories: "tools"
   topics: "automation,productivity"
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -797,11 +797,26 @@ jobs:
       clawhub_token: ${{ secrets.CLAWHUB_TOKEN }}
 ```
 
+To attach release and catalog metadata, pass the matching CLI values to either
+job:
+
+```yaml
+with:
+  dry_run: false
+  changelog: "Describe the changes in this release."
+  categories: "tools"
+  topics: "automation,productivity"
+```
+
 Notes:
 
 - The reusable workflow defaults `source` to the caller repo.
 - For monorepos, pass `source_path` so the workflow publishes the plugin
   package folder, for example `source_path: extensions/codex`.
+- `changelog`, `categories`, and `topics` are optional. When present, the
+  workflow forwards them to the matching package publish CLI flags. Categories
+  and topics use comma-separated values; omitting them preserves the existing
+  workflow behavior.
 - Pin the reusable workflow to a stable tag or full commit SHA. Do not run release publishing from `@main`.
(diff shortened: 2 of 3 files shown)


sergiopesch:
@clawsweeper re-review

The requested real behavior proof is now attached to the PR body for exact head `6f34a65f328f938326c7826463a2311edc325d72`.

- [Metadata-present reusable-workflow job](https://github.com/sergiopesch/raspberry-pi-maker/actions/runs/29866192179/job/88754910551) completed the real inspector + `package publish --dry-run` path. Its resolved command contains `--changelog 'Proof: forward release metadata through the reusable workflow.' --categories tools --topics raspberry-pi,gpio,sensors,python,maker`.
- [Omitted-input control job](https://github.com/sergiopesch/raspberry-pi-maker/actions/runs/29866192179/job/88754910604) completed the same path, and its resolved command contains none of those three flags.
- [Exact-head validation](https://github.com/sergiopesch/raspberry-pi-maker/actions/runs/29866192179/job/88754910391) passed the focused test, deterministic static gates, 4,873 unit tests, and 468 package tests. The audit-only step records two current advisories in untouched dependencies, documented in the PR body.
- [Caller CI](https://github.com/sergiopesch/raspberry-pi-maker/actions/runs/29866191172), [existing package workflow](https://github.com/sergiopesch/raspberry-pi-maker/actions/runs/29866192661), and [exact-head secret scan](https://github.com/sergiopesch/clawhub/actions/runs/29865683102) are green.

Both proof calls pin immutable ClawHub and Raspberry Pi Maker SHAs, hardcode `dry_run: true`, and use no repository secrets.

docs/cli.md
```diff
@@ -750,11 +750,26 @@ jobs:
...
+To attach release and catalog metadata, pass the matching CLI values to either
+job:
+
+```yaml
+with:
+  dry_run: false
+  changelog: "Describe the changes in this release."
+  categories: "tools"
+  topics: "automation,productivity"
+```
```
Yigtwxx: This snippet is copy-pasteable in a way that can turn a dry run into a real publish.

The prose says "pass the matching CLI values to **either** job", but the example above it (the one ending at line 751) defines two jobs — a `dry-run:` job with `dry_run: true` and a `publish:` job with `dry_run: false`. Pasting this block into the `dry-run:` job, which is what "either job" invites, silently flips it to a real publish. That directly contradicts the note four lines further down: "`pull_request` should use `dry_run: true` so CI stays non-polluting."

`dry_run` isn't one of the three keys this PR is documenting, so dropping it keeps the snippet about the new inputs and removes the trap:

````suggestion
To attach release and catalog metadata, add the matching CLI values to the
publishing job:

```yaml
with:
  changelog: "Describe the changes in this release."
  categories: "tools"
  topics: "automation,productivity"
```
````

(For what it's worth I checked `tools` against `PLUGIN_CATEGORY_SLUGS` in `packages/schema/src/catalogMetadata.ts` — it is a real slug, so the example value itself is fine.)

File: .github/workflows/package-publish.yml [hunk 2]
```diff
@@ -470,6 +488,12 @@ jobs:
               cmd += ["--version", version]
           if tags:
               cmd += ["--tags", tags]
+          if changelog:
+              cmd += ["--changelog", changelog]
+          if categories:
+              cmd += ["--categories", categories]
+          if topics:
+              cmd += ["--topics", topics]
```
Yigtwxx, line 537: `categories` and `topics` are forwarded with no pre-flight validation, which is a notable asymmetry with the input three lines above.

`family` validates inline against a literal allowlist right here in the same step (`valid_families` at `:484`), and the workflow also has a dedicated fail-fast `Validate publish mode inputs` step that rejects bad values before any work starts.

I traced where a bad category actually surfaces, and it's late: the CLI does not validate locally (`packages/clawhub/src/cli/commands/packages.ts` just runs the value through `parseCsv` into the payload), so the first rejection is server-side in `convex/packages.ts`, where `resolvePluginCategories` throws `Unknown plugin category slug "…"`. By then the job has installed the CLI, written config, downloaded and extracted the artifact and run a full `package validate` pass — inside a 75-minute budget. A single typo (`developer-tools` instead of `tools`) costs the whole run, and the same applies to exceeding `CATALOG_TOPIC_LIMIT`.

The repo already owns the allowlist: `PLUGIN_CATEGORY_SLUGS` / `isPluginCategorySlug` in `packages/schema/src/catalogMetadata.ts`, and `normalizeCatalogTopics` enforces `CATALOG_TOPIC_LIMIT` and `CATALOG_TOPIC_MAX_LENGTH` in the same file.

Either mirror the `valid_families` pattern for categories, or — if the deliberate position is that validation belongs to the CLI and server — say so in the `docs/cli.md` note. Right now that note only mentions the comma-separated format, which reads as though anything comma-separated is accepted.

.github/workflows/package-publish.yml
```diff
@@ -49,6 +49,18 @@ on:
...
+        required: false
+        type: string
+      categories:
+        description: Optional comma-separated plugin category slugs.
+        required: false
+        type: string
+      topics:
+        description: Optional comma-separated catalog topics.
+        required: false
+        type: string
```
Yigtwxx, line 78: The placement here is right — these sit next to `tags`, the other catalog-ish string input, and the env and Python halves land in the matching spots, so the diff reads as a continuation of the existing structure rather than a bolt-on.

One question about scope: `skill publish` accepts the identical `--changelog` / `--categories` / `--topics` trio, but `.github/workflows/skill-publish.yml` forwards only `owner` and `tags`. After this lands, plugin publishers can set catalog metadata through the supported OIDC path and skill publishers can't. The step structure over there is the same shape, so it's roughly the same twenty lines — worth deciding whether that belongs in this PR or a follow-up, rather than leaving the gap undocumented.

Minor, while you're in this block: every string input in `skill-publish.yml` declares `default: ""` explicitly and these three don't. Behaviour is identical (an unset `workflow_call` string input is `""`, and `.strip()` keeps it falsy), and this file is already mixed on the convention, so this is only worth a line if you want the two workflows to converge.

src/__tests__/package-publish-workflow.test.ts
```diff
@@ -168,4 +168,32 @@ describe("package publish workflow", () => {
...
+    const inputs = workflow.on?.workflow_call?.inputs;
+    const resolveStep = workflow.jobs?.publish?.steps?.find(
+      (step) => step.name === "Resolve publish command",
+    );
+
+    for (const name of ["changelog", "categories", "topics"]) {
+      const envName = `INPUT_${name.toUpperCase()}`;
+      expect(inputs?.[name]).toMatchObject({ required: false, type: "string" });
+      expect(resolveStep?.env?.[envName]).toBe(`\${{ inputs.${name} }}`);
+      expect(resolveStep?.run).toContain(`${name} = os.environ["${envName}"].strip()`);
```
Yigtwxx, line 213: This is a stronger contract test than most of the file — asserting the whole chain (input declaration, then the exact env expression string, then the Python read, then the conditional append) rather than a single `toContain`. The `if ${name}:\n    cmd += [...]` assertion in particular pins the post-YAML-dedent indentation, so it would catch a reindent of the block scalar that a looser substring check would sail past.

One small thing: `scripts/security/package-publish-workflow.test.ts` already parses this same workflow and does `job.steps.find((step) => step.name === "Resolve publish command")` with a `WorkflowStep` type alias at the top. This test re-declares an inline structural type and repeats the lookup. Not worth a round trip on its own, but if you end up pushing anyway, sharing that alias would keep the two files from drifting on the step name.

Yigtwxx (commented):
Read through this against `main` at `a9d04bb0`. The mechanism is right: the three flags go through the existing env-var to Python-list to `shlex.quote` path rather than being interpolated into a shell line, which is what makes a free-text `changelog` safe to pass at all — a multi-line or quote-bearing value can't break the generated script. Backward compatibility is real rather than asserted, since an unset `workflow_call` string input is `""` and the `.strip()` guards keep it falsy, so existing callers get no new flags.

Also worth noting for anyone rebasing: I checked every hunk's context against current `main`, and although the base is ~109 commits behind (the `wait_for_publication` work in `1a3ee6e0` touched both files since), all the surrounding lines are still byte-identical, so this should still apply with line offsets only.

Four inline notes, in rough order of how much they'd cost someone: the docs snippet can flip a dry-run job to a real publish, `categories`/`topics` reach the server unvalidated after a long job, `skill-publish.yml` is left without the same inputs, and a small test-duplication nit.

Patrick-Erichsen (changes requested):
This exposes the package CLI’s existing changelog/category/topic fields through the supported OIDC reusable workflow. The workflow implementation and real dry-run proof are strong, but the copy-ready documentation can turn a pull-request dry run into a real publish.

LOC: +67/-0 (3 files)

Finding: do not present `dry_run: false` as a metadata block that applies to either job. Remove that override or place the metadata fields only in an explicitly trusted publishing-job example, keeping pull-request examples hardcoded to `dry_run: true`.

Best-fix verdict: best API shape with a blocking docs safety defect. Optional workflow inputs forwarded through the existing Python argv builder preserve the single OIDC publish path and existing CLI validation.

Alternatives considered: asking callers to replace the official workflow with direct CLI publishing breaks trusted-publisher verification; duplicating validation in YAML would create drift.

Code read: `.github/workflows/package-publish.yml`, current publish-command builder/wait flow, `package-publish-workflow.test.ts`, `docs/cli.md`, and the immutable real caller proof.

Remaining uncertainty: rebase onto current main and rerun the focused workflow contract after the documentation correction.

docs/cli.md
Yigtwxx: The review above lists "rebase onto current main" as the remaining uncertainty, so I rechecked that part against `main` at `f9ea25e1`.

`git merge-tree --write-tree origin/main 6f34a65f` returns a tree and no conflicts, so the branch still merges cleanly even though `wait_for_publication` has since touched both `docs/cli.md` and `.github/workflows/package-publish.yml`. A rebase looks optional here rather than blocking.

The finding survives that merge unchanged. On current `main` the caller example this snippet follows is still the same two jobs — `dry-run:` with `dry_run: true` (`docs/cli.md:761`) and `publish:` with `dry_run: false` (`docs/cli.md:767`) — and the note at `docs/cli.md:785` still reads "`pull_request` should use `dry_run: true` so CI stays non-polluting". So the snippet lands in the same position against the same example after a rebase, and the copy-paste path into the dry-run job is unchanged.

The suggestion block above still applies cleanly to the current diff if you want to take it as-is.

PR #3073, opened 2026-07-13
perf(publishers): reduce profile and catalog read amplification (#3064 / #2925)

vyctorbrzezowski:
## Summary

- In this PR, publisher profiles and settings use denormalized counters when they are complete, with the existing table reads retained as a fallback for legacy rows.
- Publisher memberships return lightweight summaries, while the exact deletion inventory is loaded only after the destructive dialog opens.
- Publisher catalog pagination selects one resource kind per request, uses the matching index, and keeps legacy numeric cursors resumable without mixing cursor modes.
- This is a partial step for #3064: `getPublishedDisplayManifest` remains unbounded, and this PR does not add a `starredCount` migration or a mixed skill/plugin cursor.

## Linked Issue

- Refs #3064
- Related #2925

## Screenshots

- [x] Real ClawHub profile/catalog and account-deletion inventory proof attached in the PR proof comment.

## Behavioural Proof

- [x] In the proposed candidate, a seeded publisher profile rendered 12 of 14 skills on the first page, then all 14 after pagination, without duplicate or missing items.
- [x] As the seeded `@local` owner, Settings → Account deletion loaded the exact 2-skill/3-plugin inventory only after the dialog opened; the dialog was cancelled and no deletion was confirmed.

## Security / Trust Impact

- [x] No security/trust impact. Existing publisher visibility, ownership authorization, and deletion mutations are unchanged.

## Data / Deploy Impact

- [x] No data migration or production deploy. Legacy rows continue through the fallback path.

## Verification

- [x] `bun run ci:static`
- [x] Focused tests: `bun run test -- convex/publishers.test.ts src/routes/-settings.test.tsx src/__tests__/user-profile-route.test.tsx src/components/PublisherListItem.test.tsx src/__tests__/home-popular-publishers-section.test.tsx src/lib/slugRoute.test.ts`
- [x] `bun run ci:unit`
- [x] `bun run ci:types-build`
- [x] `bunx convex codegen`
- [x] `git diff --check`

Issue #3064, opened 2026-07-13
perf: bound publisher profile, membership, and catalog reads


vyctorbrzezowski:
## Summary

Several publisher-profile and authenticated publisher paths read or serialize complete resource catalogs even when the caller only needs identity, role, counters, or one page. These paths are currently bounded only at the response layer, not at the database-read layer.

Related umbrella: #2925.

## Confirmed source evidence

- `publishers.getProfileByHandle` calls `toPublisherListItem` with `forceComputedStats`, `includePublishedItems`, and `includeStarredCount`. This can collect all active skills/packages and all stars. `src/routes/user/$handle.tsx` does not consume `publisher.publishedItems`, and it invokes the profile query once in the loader and again as a subscription.
- `publishers.listMine` collects the user's memberships, then requests all published items for every publisher. There are eight runtime subscribers; only the destructive summary in Settings consumes `publishedItems`.
- `publishers.listPublishedPage` collects and sorts every active skill and package before applying its offset cursor with `slice`.
- `publishers.getPublishedDisplayManifest` is a sibling profile path that also collects the complete published catalog for GitHub-backed publishers.

These are source-proven unbounded read shapes. This issue does not claim production p50/p95 or billing numbers because Convex production Insights were unavailable during the audit.

## Proposed outcome

1. Make the profile summary use complete denormalized counters when available, with the existing computed fallback for legacy incomplete publishers. Do not include `publishedItems`.
2. Add a lightweight membership summary for the eight `listMine` consumers.
3. Load deletion inventory only while the relevant destructive dialog is open. Keep authorization in the deletion mutation.
4. Paginate the published catalog directly through the existing publisher-scoped index for the requested `kind`.
5. Explicitly bound, redesign, or document the remaining `getPublishedDisplayManifest` behavior before claiming the profile path is fully bounded.

## Non-goals

- Changing profile reactivity solely for performance.
- Adding a migrated `starredCount` counter in the first change.
- Solving alternate sorting of starred skills. That requires workload data and a product/consistency decision.
- Building a mixed skill/plugin cursor: the only runtime catalog caller currently supplies `kind`.

## Acceptance criteria

- With complete publisher counters, profile and membership summary tests fail if `skills` or `packages` are queried.
- Missing-counter fixtures preserve the existing computed fallback and visibility rules.
- The membership summary contains no `publishedItems`; deletion inventory is skipped until its dialog opens.
[...]

PR #3078, opened 2026-07-14
fix: settings page fails to load when the deployed backend predates the listMine args change

mattmillerai:
## What Problem This Solves

Signed-in users currently get the global error boundary on `/settings`:

```
Something went wrong
[CONVEX Q(publishers:listMine)] Server Error Called by client
```

#3073 made the settings route pass `{ includePublishedItems: false }` to `publishers:listMine`, but Convex validators reject unknown arguments. Whenever the deployed web app is newer than the deployed Convex backend — true in production right now — the whole settings route error-boundaries for every signed-in user.

Verifiable from anywhere (unauthenticated):

```bash
# what the settings route sends today → error on the live backend
curl -s -X POST https://wry-manatee-359.convex.cloud/api/query \
  -H 'Content-Type: application/json' \
  -d '{"path":"publishers:listMine","args":{"includePublishedItems":false},"format":"json"}'
# → {"status":"error","errorMessage":"[Request ID: …] Server Error"}

# legacy args → success (deployed validator still has args: {})
curl -s -X POST https://wry-manatee-359.convex.cloud/api/query \
  -H 'Content-Type: application/json' \
  -d '{"path":"publishers:listMine","args":{},"format":"json"}'
# → {"status":"success","value":[]}
```

## What This Changes

The settings route omits the argument and passes `{}`, which both backend generations accept — the new backend defaults `includePublishedItems` to `true`, so rendering is unchanged and the page works regardless of deploy order. A brief comment marks the rollout hazard. The lightweight-list optimization from #3073 can be reintroduced once the new validator is verifiably live in production.

Even after the backend catches up, this keeps `/settings` resilient to the frontend-ahead-of-backend deploy window that any future arg addition reopens.

Tests: `src/routes/-settings.test.tsx` updated (24/24 passing); biome clean.

Related: #3064

### DIFF
--- a/src/routes/-settings.test.tsx
+++ b/src/routes/-settings.test.tsx
@@ -365,7 +365,7 @@ describe("Settings", () => {
     render(<Settings />);
 
     expect(getLastQueryArgs("tokens:listMine")).toEqual({});
-    expect(getLastQueryArgs("publishers:listMine")).toEqual({ includePublishedItems: false });
+    expect(getLastQueryArgs("publishers:listMine")).toEqual({});
     expect(getLastQueryArgs("publishers:getDeletionInventory")).toBe("skip");
     useQueryMock.mockClear();
     fireEvent.click(screen.getByRole("button", { name: "Delete account" }));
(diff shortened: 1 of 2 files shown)


Patrick-Erichsen:
Resolved by the production Convex rollout on July 14, 2026: the deployed `publishers:listMine` validator now accepts `includePublishedItems`, and live production accepts both `{}` and `{ includePublishedItems: false }`. Current main intentionally retains the bounded-read argument, so this compatibility rollback is now obsolete. Thanks for catching the real deploy-skew window.

PR #3172, opened 2026-07-18
fix: owner-qualified skill reports for ambiguous slugs

### LINKED ISSUE #3111: Listing reports are impossible when a slug collides with another publisher's (report handler ignores owner)
## Summary

A listing whose slug collides with another publisher's **cannot be reported at all** — neither via the HTTP API nor the CLI. The only working path is the signed-in web dialog. Per SECURITY.md this is a bug in "report handling".

The collision that surfaced it is also, itself, the listing I am trying to report (details at the bottom).

## Root cause

`reportSkillForUserInternal` (`convex/skills.ts:~4365`) resolves with the **non**-owner-qualified resolver:

```ts
const resolved = await resolveSkillBySlugOrAlias(ctx, args.slug);
const skill = resolved.skill;
if (!skill || skill.softDeletedAt || skill.moderationStatus === "removed") {
  throw new ConvexError("Skill not found");
}
```

When two publishers share a slug, `resolved.ambiguous` is set and `resolved.skill` is null, so it throws **"Skill not found"** — which is misleading, because the skill exists.

Note the sibling helper `resolveUnambiguousSkillForLegacySlug` already handles this properly:

```ts
if (resolved.ambiguous) {
  throw new ConvexError("Slug is used by multiple publishers. Use an owner-qualified skill URL.");
}
```

…and `resolveSkillBySlugOrAliasForOwner` (`convex/skills.ts:~2525`) exists. The report path just doesn't use either.

The HTTP route (`convex/httpApiV1/skillsV1.ts`, `segments.length === 2 && action === "report"`) only ever forwards a bare `slug` — there is no owner parameter to forward, even though `GET /api/v1/skills/{slug}?owner={handle}` accepts one.

By contrast the web dialog works, because it passes an unambiguous id:

```ts
const reportSkill = useMutation(api.skills.report);
await reportSkill({ skillId: skill._id, reason: trimmedReason });
```

## Reproduction

Two publishers currently share the slug `wp-multi-tool` (`@marcindudekdev`, `@kingaiwork`).

```bash
# GET disambiguates fine:
curl -H "Authorization: Bearer $TOKEN" \
  "https://clawhub.ai/api/v1/skills/wp-multi-tool?owner=kingaiwork"      # -> 200, correct skill

# POST report cannot:
curl -X POST -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
  -d '{"reason":"..."}' \
  "https://clawhub.ai/api/v1/skills/wp-multi-tool/report?owner=kingaiwork"
# -> 400 Uncaught ConvexError: Skill not found   (?owner= is ignored)
```

URL-encoding the owner into the slug (`kingaiwork%2Fwp-multi-tool`, `%40kingaiwork%2Fwp-multi-tool`) also returns "Skill not found".

`clawhub install wp-multi-tool` is affected by the same collision — it errors `AMBIGUOUS_SKILL_SLUG` and suggests using `@marcindudekdev/wp-multi-tool`, but the CLI then rejects that form as `Invalid slug`, and there is no `--owner` flag. So the error message recommends a syntax the CLI does not accept.

## Suggested fix

1. Accept an owner qualifier on the report route (`?owner=<handle>`, mirroring GET), or accept `skillId`.
2. Failing that, surface the ambiguous case with the existing message ("Slug is used by multiple publishers. Use an owner-qualified skill URL.") rather than "Skill not found", so the caller can tell "doesn't exist" from "can't disambiguate".
3. Make the `AMBIGUOUS_SKILL_SLUG` hint match a form `clawhub install` actually accepts.

## The listing I am trying to report

`@kingaiwork/wp-multi-tool` — https://clawhub.ai/kingaiwork/skills/wp-multi-tool

- It reproduces the **display name** of my skill verbatim: "WP Multitool — WordPress Optimization Toolkit" (@marcindudekdev/wp-multi-tool).
- It reproduces my SKILL.md **description verbatim**: "WordPress site health audit, performance optimization, database cleanup, autoload tuning, slow query detection, wp-config management, image size control, …".
- It presents itself as a **"Free Tier / 免费版" of WP Multitool**. WP Multitool is my commercial plugin and has **no free tier** (Lite is $9 one-time, Pro $79/year), so the listing states something untrue about a paid third-party product.
- It advertises "Pro / Advanced / Enterprise" upgrades at kingai.work under my product's name. I have no affiliation with them.
- Its actual commands are unrelated boilerplate (`iptables -L INPUT`, `ss -tlnp`, `lastb`, `systemctl is-active fail2ban`) — nothing to do with the product it names, so an agent installing it under my brand gets unrelated shell commands.

SECURITY.md categories: misleading metadata, impersonation, trademark misuse.

Happy to move the listing part to the proper report flow once it can target the listing — which is the point of this issue.

pgm-norbert:
## Summary
- Fixes skill listing reports for ambiguous slugs (SECURITY.md: report handling).
- `POST /api/v1/skills/{slug}/report` now accepts `owner` / `ownerHandle` from the query or JSON body, matching GET disambiguation.
- Bare-slug collisions now return the standard message (`Slug is used by multiple publishers. Use an owner-qualified skill URL.`) instead of misleading `Skill not found`.
- Keeps the URL slug plus owner qualifier authoritative; the endpoint does not accept a conflicting direct skill ID.

Fixes #3111

## Test plan
- [x] `VITE_CONVEX_URL=https://example.invalid bunx vitest run convex/httpApiV1.handlers.test.ts -t "skill report"` (4 passed)
- [x] Authenticated local Convex duplicate-slug proof:
  - merge base `aaa73625ed41`: owner-qualified and bare requests both returned `400 Skill not found`
  - PR head `34ef2ade`: owner-qualified request returned `200` and persisted the requested publisher's exact skill ID; bare request returned `400` with owner-qualified guidance
- [x] `bun run ci:unit` (4,799 passed, 1 skipped)
- [x] `bun run ci:types-build`
- [x] `bun run ci:static`
- [x] `$autoreview` (clean; no actionable findings)

## Notes
Maintainer follow-up `34ef2ade` removed the unnecessary HTTP `skillId` override so the report target cannot disagree with the route.

### DIFF
--- a/convex/httpApiV1.handlers.test.ts
+++ b/convex/httpApiV1.handlers.test.ts
@@ -3519,6 +3519,53 @@ describe("httpApiV1 handlers", () => {
     );
   });
 
+  it("skill report forwards owner query/body and skillId for ambiguous slugs", async () => {
+    vi.mocked(requireApiTokenUser).mockResolvedValue({
+      userId: "users:reporter",
+      user: { _id: "users:reporter", role: "user" },
+    } as never);
+    const runMutation = vi.fn(async (_mutation: unknown, args: Record<string, unknown>) => {
+      if (isRateLimitArgs(args)) return okRate();
+      return {
+        ok: true,
+        reported: true,
+        alreadyReported: false,
+        reportId: "skillReports:2",
+        skillId: "skills:owner-scoped",
+        reportCount: 1,
+      };
+    });
+
+    const response = await __handlers.skillsPostRouterV1Handler(
+      makeCtx({ runMutation }),
+      new Request(
+        "https://example.com/api/v1/skills/wp-multi-tool/report?owner=kingaiwork",
+        {
+          method: "POST",
+          headers: { Authorization: "Bearer clh_test" },
+          body: JSON.stringify({
+            reason: "impersonation",
+            ownerHandle: "kingaiwork",
+            skillId: "skills:owner-scoped",
+          }),
+        },
+      ),
+    );
+
+    expect(response.status).toBe(200);
+    expect(runMutation).toHaveBeenCalledWith(
+      (internal as unknown as { skills: Record<string, unknown> }).skills
+        .reportSkillForUserInternal,
+      {
+        actorUserId: "users:reporter",
+        slug: "wp-multi-tool",
+        reason: "impersonation",
+        ownerHandle: "kingaiwork",
+        skillId: "skills:owner-scoped",
+      },
+    );
+  });
+
   it("skill report triage posts moderator decisions", async () => {
     vi.mocked(requireApiTokenUser).mockResolvedValue({
       userId: "users:moderator",
--- a/convex/httpApiV1.handlers.test.ts
+++ b/convex/httpApiV1.handlers.test.ts
@@ -3519,7 +3519,7 @@ describe("httpApiV1 handlers", () => {
     );
   });
 
-  it("skill report forwards owner query/body and skillId for ambiguous slugs", async () => {
+  it("skill report forwards owner query for ambiguous slugs", async () => {
     vi.mocked(requireApiTokenUser).mockResolvedValue({
       userId: "users:reporter",
       user: { _id: "users:reporter", role: "user" },
@@ -3538,18 +3538,13 @@ describe("httpApiV1 handlers", () => {
 
     const response = await __handlers.skillsPostRouterV1Handler(
       makeCtx({ runMutation }),
-      new Request(
-        "https://example.com/api/v1/skills/wp-multi-tool/report?owner=kingaiwork",
-        {
-          method: "POST",
-          headers: { Authorization: "Bearer clh_test" },
-          body: JSON.stringify({
-            reason: "impersonation",
-            ownerHandle: "kingaiwork",
-            skillId: "skills:owner-scoped",
-          }),
-        },
-      ),
+      new Request("https://example.com/api/v1/skills/wp-multi-tool/report?owner=kingaiwork", {
+        method: "POST",
+        headers: { Authorization: "Bearer clh_test" },
+        body: JSON.stringify({
+          reason: "impersonation",
+        }),
+      }),
     );
 
     expect(response.status).toBe(200);
@@ -3561,7 +3556,6 @@ describe("httpApiV1 handlers", () => {
         slug: "wp-multi-tool",
         reason: "impersonation",
         ownerHandle: "kingaiwork",
-        skillId: "skills:owner-scoped",
       },
     );
   });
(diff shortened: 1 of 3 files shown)


Patrick-Erichsen:
Maintainer proof at exact head `34ef2ade` (base `aaa73625ed41`).

I exercised the real authenticated HTTP path against disposable local Convex state containing two publishers with the same slug, a hashed API token, and moderation-report readback.

**Unpatched merge base**

```json
{"mode":"unpatched","scopedStatus":400,"scopedMessage":"Skill not found","bareStatus":400,"bareMessage":"Skill not found"}
```

**Candidate**

```json
{"mode":"candidate","scopedStatus":200,"persistedTargetMatchesOwner":true,"bareStatus":400,"bareGuidance":true}
```

The candidate's owner-qualified POST returned `200`; the stored report's `skillId` matched the requested publisher's listing. The same bare slug returned `400` with `Slug is used by multiple publishers. Use an owner-qualified skill URL.`

I also added `34ef2ade` to keep the route slug and owner authoritative. The original optional HTTP `skillId` could disagree with the URL and silently target another listing; the signed-in web dialog uses its separate public mutation and did not require that bypass.

Validation:
- focused report tests: 4 passed
- full unit coverage: 4,799 passed, 1 skipped
- `ci:types-build`: passed
- `ci:static`: passed
- required schema and CLI TypeScript checks: passed
- autoreview: clean, no actionable findings

Best-fix verdict: **best**. This reuses the established owner-aware resolver at the report mutation boundary, preserves the shared bare-slug ambiguity guidance, and avoids widening the HTTP target contract.

Alternatives considered:
- Parse `owner/slug` inside the slug path: rejected because existing GET and sibling write routes already use `ownerHandle`.
- Accept a direct HTTP `skillId`: rejected because it can conflict with the route and is unnecessary for the web dialog.

@clawsweeper re-review

Patrick-Erichsen (approved):
This fixes the authenticated reporting path for owner-scoped slug collisions and preserves a truthful error for unresolved bare slugs.

LOC: +58/-1 (3 files)

No blocking findings.

Best-fix verdict: **best**. The HTTP handler forwards the established owner qualifier to the existing owner-aware resolver, and the mutation owns the ambiguity decision. Maintainer commit `34ef2ade` removes the unnecessary direct-ID override so the URL target cannot disagree with the persisted report.

Evidence checked:
- exact-head authenticated local Convex proof with duplicate publisher slugs, hashed API token, POST, and report-list readback
- identical merge-base replay demonstrating `400 Skill not found` before the fix
- focused handler tests, full unit coverage, static/types/build gates, and clean autoreview
- issue #3111, surrounding resolver/report code, CLI caller, web-dialog caller, CI, discussion, and current review state

Remaining uncertainty: the fork cannot create an authorized Vercel preview, but this is a backend-only change and local exact-head runtime proof covers the affected HTTP/Convex path.

PR #3174, opened 2026-07-18
fix: show one year instead of 12 months in relative timestamps

Yigtwxx:
## What Problem This Solves

Fixes an issue where listings last updated between 360 and 364 days ago displayed
"Updated 12mo ago" instead of "Updated 1y ago".

`timeAgo` measures months as 30 days but years as 365 days. Those two units do not
tile: 360-364 days still divides into twelve whole 30-day months while remaining
below the 365-day year threshold, so the month branch renders a twelve-month value
that should already have rolled over into years.

The affected surfaces are the skill list rows, skill header, browse results, plugin
detail page, dashboard catalog and needs-attention views, and the GitHub sync
timestamp in settings.

## Why This Change Was Made

The repository already contains the intended semantics. `formatRelativeUpdatedAt`
in `src/routes/user/$handle.tsx` derives years from whole months and caps the month
branch at 11:

```ts
const months = Math.floor(days / 30);
if (months < 12) return `${months}mo ago`;
return `${Math.floor(months / 12)}y ago`;
```

`timeAgo` is now aligned with that rule, so the two relative-time renderers agree at
the one-year boundary. Deriving years from months also removes the unit mismatch at
its source rather than special-casing the twelve-month output.

- Consolidating `timeAgo` and `formatRelativeUpdatedAt` into a single helper. They
  are now consistent; merging them is a separate refactor.
- Making the 30-day month calendar-aware.
- Future-dated timestamps. Both functions already render those as "just now", so
  clamping the difference would be a no-op here.

## User Impact

Relative timestamps no longer show a twelve-month value. A listing that has not been
updated for just under a year now reads "1y ago", matching what the publisher profile
route has always shown for the same listing.

## Evidence

Added `src/lib/timeAgo.test.ts`, which had no test coverage before this change even
though `src/lib/**` is inside the `vitest.config.ts` coverage `include` list.

The new boundary case fails on the parent commit and passes with the fix:

```
$ bunx vitest run src/lib/timeAgo.test.ts        # before the fix
FAIL  src/lib/timeAgo.test.ts > timeAgo > rolls the last days before a year over into years
AssertionError: expected '12mo ago' to be '1y ago'
 ❯ src/lib/timeAgo.test.ts:54:28
     54|     expect(agoByDays(360)).toBe("1y ago");

Test Files  1 failed (1)
     Tests  1 failed | 5 passed (6)

$ bunx vitest run src/lib/timeAgo.test.ts        # with the fix
Test Files  1 passed (1)
     Tests  6 passed (6)
```

The other five cases pass on both sides, so the existing minute, hour, day, week and
month branches are unchanged.

Rendered `SkillListItem` output, reading `.skill-list-item-meta-item.is-updated` from
the mounted component with the clock fixed at 2026-06-23T12:00:00Z:

| updatedAt | before | after |
| --- | --- | --- |
| now - 330 days | Updated 11mo ago | Updated 11mo ago |
| now - 359 days | Updated 11mo ago | Updated 11mo ago |
| now - 360 days | Updated 12mo ago | Updated 1y ago |
| now - 364 days | Updated 12mo ago | Updated 1y ago |
| now - 365 days | Updated 1y ago | Updated 1y ago |
| now - 400 days | Updated 1y ago | Updated 1y ago |
| now - 730 days | Updated 2y ago | Updated 2y ago |

Regression check on the consumers and on the sibling renderer:

```
$ bunx vitest run src/lib/timeAgo.test.ts src/components/SkillListItem.test.tsx \
    src/__tests__/user-profile-route.test.tsx
Test Files  3 passed (3)
     Tests  29 passed (29)
```

Gates run locally on Windows:

- `bun run format:check -- src/lib/timeAgo.ts src/lib/timeAgo.test.ts` — clean
- `bun run lint` — clean
- `bunx tsc --noEmit -p tsconfig.json` — clean
- `bun run ci:unit` — 4722 passed. The 22 failures are pre-existing on this platform
  and unrelated to this change: they are `spawn bun ENOENT` in the worker and
  security-dataset suites plus the design-contract and management route suites. I
  confirmed they fail identically on the unmodified parent commit, so I have left
  them alone. CI on Linux is the authoritative signal here.

other fork pull requests.

### DIFF
--- a/src/lib/timeAgo.test.ts
+++ b/src/lib/timeAgo.test.ts
@@ -1,0 +1,63 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { timeAgo } from "./timeAgo";
+
+const MINUTE = 60_000;
+const HOUR = 60 * MINUTE;
+const DAY = 24 * HOUR;
+
+const NOW = Date.UTC(2026, 5, 23, 12, 0, 0);
+
+function agoByDays(days: number) {
+  return timeAgo(NOW - days * DAY);
+}
+
+describe("timeAgo", () => {
+  beforeEach(() => {
+    vi.useFakeTimers();
+    vi.setSystemTime(NOW);
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  it("reports sub-minute gaps as just now", () => {
+    expect(timeAgo(NOW)).toBe("just now");
+    expect(timeAgo(NOW - 59_999)).toBe("just now");
+  });
+
+  it("formats minutes and hours", () => {
+    expect(timeAgo(NOW - MINUTE)).toBe("1m ago");
+    expect(timeAgo(NOW - 59 * MINUTE)).toBe("59m ago");
+    expect(timeAgo(NOW - HOUR)).toBe("1h ago");
+    expect(timeAgo(NOW - 23 * HOUR)).toBe("23h ago");
+  });
+
+  it("formats days and weeks", () => {
+    expect(agoByDays(1)).toBe("1d ago");
+    expect(agoByDays(6)).toBe("6d ago");
+    expect(agoByDays(7)).toBe("1w ago");
+    expect(agoByDays(29)).toBe("4w ago");
+  });
+
+  it("formats months", () => {
+    expect(agoByDays(30)).toBe("1mo ago");
+    expect(agoByDays(60)).toBe("2mo ago");
+    expect(agoByDays(180)).toBe("6mo ago");
+  });
+
+  // Regression: 30-day months divided into a 365-day year yielded "12mo ago"
+  // for gaps of 360-364 days. Months must roll over into years at 12.
+  it("rolls the last days before a year over into years", () => {
+    expect(agoByDays(330)).toBe("11mo ago");
+    expect(agoByDays(359)).toBe("11mo ago");
+    expect(agoByDays(360)).toBe("1y ago");
+    expect(agoByDays(364)).toBe("1y ago");
+  });
+
+  it("formats years", () => {
+    expect(agoByDays(365)).toBe("1y ago");
+    expect(agoByDays(400)).toBe("1y ago");
+    expect(agoByDays(730)).toBe("2y ago");
+  });
+});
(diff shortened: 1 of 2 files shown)


Patrick-Erichsen (approved):
This fixes the shared relative-time helper so ClawHub never renders `12mo ago`; at 12 whole 30-day months it now rolls over to `1y ago`, matching the publisher-profile formatter.

LOC: +70/-7 (2 files)

No blocking findings.

**Best-fix verdict:** best. The change updates the owner helper used by skill rows, skill/plugin details, browse results, dashboard views, and GitHub sync settings. Deriving years from the same 30-day month convention is the narrowest fix that also agrees with `formatRelativeUpdatedAt`.

**Boundary review:** `359d -> 11mo`, `360d -> 1y`, `364d -> 1y`, `365d -> 1y`, and later 12-month rollovers remain internally consistent. Future timestamps still return `just now`. The output remains the existing locale-independent compact English format; this PR does not add or remove localization behavior.

**Alternatives considered:** clamping only the literal `12mo` case would preserve two competing year units and let the helpers disagree; consolidating both formatters would be a broader refactor than this bug requires.

**Code read:** `src/lib/timeAgo.ts`, its new boundary tests, all production callers, `src/routes/user/$handle.tsx`, and the adjacent profile/list-item tests.

**Validation on `39f424e9b06ab5509aae32a9705cb5417eca62e3`:**
- focused consumers: 3 files, 29 tests passed
- `bun run ci:static`: passed
- `bunx tsc --noEmit -p tsconfig.json`: passed
- `bun run ci:unit`: 4,803 passed, 1 skipped, with one unrelated `package-detail-route` responsive-layout timeout; that test passes immediately in isolation on both the exact PR head and a clean synthetic merge with current `main`
- synthetic merge with current `main` `57d1e1530b5517d01c916a032efd1653abe691c9`: conflict-free; the same 29 focused tests passed

**Provenance:** the mixed 30-day-month/365-day-year helper was introduced by Vincent Koc in commit `150470820838` through #1567 (PR author `BunsDev`, merged April 7, 2026). Current fix author: `Yigtwxx`.

**Remaining uncertainty:** GitHub's Vercel status is red only because the fork author is not authorized for the OpenClaw Foundation team preview. No preview is needed to validate this deterministic text-only helper change.

Issue #3186, opened 2026-07-20
ClawScan timeouts leave staged publications stuck in pending_checks


Patrick-Erichsen:
## Summary

Multiple users are successfully creating immutable skill or package releases, but the pre-publication ClawScan step times out. The releases remain hidden in `pending_checks`, cannot be republished because the version already exists, and are not visible through normal owner or public read surfaces.

This issue tracks the shared production fix and recovery of every currently reported publication.

## Linked reports

- #3183
  - `octo@1.1.0` and `octo@1.1.1` (`caster-q`): 238 files each
  - `clawseccheck@3.54.0` (`gl0di`): 79 files
- #3182
  - `api-gateway@1.0.130`, `1.0.131`, and `1.0.132` (`byungkyu`): 231 files each
- #3175
  - `@alan512/experienceengine@0.5.3` (`Alan-512`): 3,636 files / approximately 16.4 MB unpacked

## Confirmed production state

All listed releases exist server-side and are stuck in `pending_checks` after `clawscan timed out`. VirusTotal and/or TruffleHog completed successfully where present, so these reports are not currently explained by those scanners.

The affected artifacts range from 79 to 3,636 files and include both skills and code-plugin packages. This is therefore not isolated to one owner, package family, or only extremely large artifacts.

## Done when

- [ ] Identify and fix the cause of the ClawScan timeouts in the pre-publication worker.
- [ ] Confirm normal retries finish without leaving releases indefinitely in `pending_checks`.
- [ ] Recover or rerun every release listed above.
- [ ] Verify each version is successfully published and visible through its public page/API and owner-facing CLI surfaces.
- [ ] Confirm the broader `pending_checks` timeout backlog is draining normally.
- [ ] Close #3183, #3182, and #3175 after their reported versions are verified in production.

Patrick-Erichsen:
## Resolved

Root cause: the pre-publication ClawScan subprocess was capped at 240 seconds, but legitimate larger artifacts can require longer. The proof canary (`clawseccheck@3.54.0`) completed successfully in 266 seconds once the cap was raised.

Shipped:
- Timeout raised to 900 seconds in code and production configuration: https://github.com/openclaw/clawhub/pull/3190
- Successful >240-second canary: https://github.com/openclaw/clawhub/actions/runs/29768218596
- Recovery runs: https://github.com/openclaw/clawhub/actions/runs/29769125244 and https://github.com/openclaw/clawhub/actions/runs/29771411812
[...]

PR #3198, opened 2026-07-21
fix: require undelete before republishing deleted entries

### LINKED ISSUE #3177: Deleted skill slug cannot be republished; new tag publishes but is missing from dashboard
## Summary

After deleting one of my skills, it appears that the skill was not fully removed.

When I try to publish a skill with the same slug and the same tag, publishing fails because the tag already exists. If I use a new tag, the publish command succeeds, but the skill does not appear on my dashboard.

## Steps to reproduce

1. Publish a skill.
2. Delete the skill.
3. Try to publish a skill using the same slug and the previous tag.
4. Observe an error indicating that the tag already exists.
5. Publish again using the same slug but a new tag.
6. Observe that publishing succeeds, but the skill is not visible on the dashboard.

## Actual behavior

- The deleted skill is no longer visible as expected.
- Its existing tag/version still blocks republishing.
- Publishing with a new tag reports success.
- The republished skill does not appear on the dashboard.

This makes it unclear whether the skill is soft-deleted, restored, hidden, or only partially recreated.

## Expected behavior

ClawHub should either:

- allow the owner to recreate/reuse a deleted slug and make the republished skill visible; or
- explicitly document that deletion is a soft delete and provide a restore/undelete flow instead of accepting a publish that remains invisible.

The CLI/API should not report a successful publish if the resulting skill will remain hidden from the owner's dashboard.

## Questions

1. Can an owner recreate or reuse a deleted skill slug?
2. If deletion is soft deletion, should publishing a new tag restore the skill, or must the owner use an undelete command?
3. How can I recover the skill that accepted a new tag but is missing from the dashboard?

## Related issue

This appears related to #922, which concerned a deleted slug that could not be reused. That issue was closed after the specific production record was no longer reserved, but the general lifecycle behavior and the successful-yet-invisible republish case do not appear to be covered.

Patrick-Erichsen:
## Summary

- reject publishing a new skill version when the owner-deleted parent is still hidden, with the exact `clawhub undelete @owner/slug --yes` recovery command
- reject publishing a new plugin/package release when its package is deleted, with `clawhub package undelete <name> --yes`
- create no pending version or release for these rejected attempts, and keep moderator/security-hidden skills on their existing moderation path
- safely drain pending skill attempts created by the earlier auto-restore implementation without overwriting a newer delete or moderation decision

## Behavior

Skill versions and package releases remain immutable, so republishing the same version still fails. For a new version or release, owners must explicitly restore the deleted parent first; ClawHub no longer accepts a publish that remains invisible on the dashboard.

## Root Cause

The staged skill-publish path could insert a new pending version under a soft-deleted parent. The publish command reported success, but owner dashboard queries continued excluding the deleted skill. Package publishing already rejected deleted parents, but did not provide the actionable CLI command.

## Validation

- `bunx vitest run convex/skills.backportLatest.test.ts convex/skills.undeleteGate.test.ts convex/packages.public.test.ts convex/publishAttempts.test.ts` (371 passed)
- `PLAYWRIGHT_LOCAL_AUTH_CONVEX_URL=http://127.0.0.1:3220 PLAYWRIGHT_LOCAL_AUTH_CONVEX_SITE_URL=http://127.0.0.1:3221 PLAYWRIGHT_PORT=4299 bun run test:pw:local-auth -- --project=chromium e2e/local-auth/publish-skill-lifecycle.pw.test.ts -g "publishing another version of a deleted skill requires an explicit restore"` (1 passed against isolated real Convex)
- `bun run ci:static`
- `bun run ci:unit` (4,838 passed, 1 skipped)
- `bun run ci:types-build`
- `bun run ci:packages`
- `bun run ci:e2e-http` (unrelated GitHub dry-run fixture timed out; exact failed case passed immediately in isolation)
- `.agents/skills/autoreview/scripts/autoreview --mode local` (clean)

Fixes #3177

This PR is intentionally separate from #3186 and does not include the reversible version-deletion work tracked separately.

### DIFF
--- a/convex/packages.public.test.ts
+++ b/convex/packages.public.test.ts
@@ -7793,7 +7793,10 @@ describe("packages public queries", () => {
         files: [],
         integritySha256: "abc123",
       }),
-    ).rejects.toThrow("Restore it before publishing another release");
+    ).rejects.toThrow(
+      'Package "demo-plugin" is hidden/deleted. Run "clawhub package undelete demo-plugin --yes" before publishing another release.',
+    );
+    expect(ctx.insert).not.toHaveBeenCalledWith("packageReleases", expect.anything());
   });
 
   it("rejects final package publish inserts when the actor was banned mid-publish", async () => {
--- a/convex/packages.ts
+++ b/convex/packages.ts
@@ -9765,7 +9765,7 @@ export const insertReleaseInternal = internalMutation({
     const nextNameLabel = typeof args.name === "string" ? args.name : "<unknown>";
     if (existing?.softDeletedAt) {
       throw new ConvexError(
-        `Package "${nextNameLabel}" was deleted. Restore it before publishing another release or choose a new package name.`,
+        `Package "${nextNameLabel}" is hidden/deleted. Run "clawhub package undelete ${nextNameLabel} --yes" before publishing another release.`,
       );
     }
     const nextChannel = derivePackagePublisherChannel({
--- a/convex/publishAttempts.test.ts
+++ b/convex/publishAttempts.test.ts
@@ -1028,31 +1028,75 @@ describe("publishAttempts", () => {
     });
   });
 
-  it("keeps existing skill parents when TruffleHog blocks a pending new version", async () => {
+  it("restores an owner-delete tombstone when TruffleHog blocks its pending republish", async () => {
+    const deletedAt = 1_700_000_000_000;
+    const reservedUntil = deletedAt + 30 * 24 * 60 * 60 * 1000;
     const ctx = {
       db: {
-        get: vi
-          .fn()
-          .mockResolvedValueOnce({
-            _id: "publishAttempts:demo",
-            kind: "skill",
-            status: "pending_checks",
-            userId: "users:publisher",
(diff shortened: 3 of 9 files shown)


steipete:
Closing without merge after exact-current review.

The user-visible report is reproducible and the proposed linear lifecycle works: the focused Convex suite passed 74/74, and the existing exact-head hosted checks were green. A fresh full-branch AutoReview found two security/state-machine races that make this implementation unsafe to land:

- successful finalization can overwrite a moderator, security, ban, or malware action applied while checks are pending;
- pending versions that predate the owner deletion do not share the rollback state, so cleanup can leave the parent permanently visible-but-hidden or later reverse the deletion.

Repair attempts exposed a broader missing contract rather than a local fix: the workflow needs a stable skill-ID-indexed active-attempt lookup, an explicit restore generation/marker proving which attempt owns the temporary restore, and backward-compatible handling for already in-flight rows. Slug-based lookup, general `updatedAt`, or version-history filtering are not safe substitutes.

Issue #3177 should remain open for that bounded design. No production Convex mutation or deployment was performed.

PR #3225, opened 2026-07-22
feat: add GitHub Skill Sync repository preview

Patrick-Erichsen:
## Summary

- add first-class GitHub Skill Sync settings for a verified publisher to select one of its public repositories
- authorize personal and organization publishers through immutable GitHub identities and fresh organization-admin membership
- resolve canonical repository redirects, discover skills directly from repository archives, and preview new destinations, Hosted Skill replacements, unavailable entries, and ownership conflicts
- show one repository-level replacement warning and retain the existing source-removal flow
- ignore stale preview responses when the selected repository or publisher changes

## Dependency boundary

Activation remains disabled until the canonical GitHub Skill Sync engine exposes the repository enrollment contract. This PR does not enumerate skills from skills.sh, create another mirror, run scans, change shared schema, alter ranking, or mutate production.

## Tests

- `bunx vitest run src/routes/-settings.test.tsx convex/githubSkillSyncSettings.test.ts convex/lib/githubSkillSyncSettings.test.ts` (39 passed)
- `bun run ci:unit` (5,096 passed, 1 skipped)
- `bun run ci:static`
- `bun run ci:types-build`
- `bunx convex dev --once --typecheck=disable` against local `local-amantus-clawdhub-2080`
- real local Convex action calls proved both public endpoints reject unauthenticated requests before GitHub access
- final autoreview: no accepted/actionable findings


convex/githubSkillSyncSettings.ts
```diff
@@ -0,0 +1,550 @@
...
+    },
+  )) as PublisherContext;
+}
+
+async function buildGitHubSettingsHeaders(fetcher: typeof fetch) {
+  return await buildGitHubApiHeaders({
+    userAgent: "clawhub/github-skill-sync-settings",
+    fetchImpl: fetcher,
+  });
+}
```
Yigtwxx, line 293: This sends a GitHub App installation token to endpoints that can't use one.

`buildGitHubApiHeaders` treats `useGitHubApp` as opt-out, not opt-in — `convex/lib/githubAuth.ts:48` is `if (options.useGitHubApp !== false)`. Since this call omits the flag, every request built from it (`/user/{id}`, `/organizations/{id}`, `/users/{login}/repos`, `/orgs/{login}/repos`, `/repos/{owner}/{name}`, and the archive fetch behind the snapshot) authenticates as the installation.

The sibling helper 90 lines away in `convex/githubSkillSync.ts:3199-3210` disables it explicitly and states why in a comment: *"Installation tokens are repository-scoped and cannot reliably read an arbitrary public repository selected by a publisher."* That's the same constraint this file is under — a publisher picking any of their public repos is precisely the case where the installation may not be present.

In a deployment where an installation token is available, repository listing and preview would 403/404 or silently return a truncated repo set for anything outside the installation. The unit tests can't catch it because they inject a `fetchMock`, so the header contents never reach a real endpoint.

```suggestion
async function buildGitHubSettingsHeaders(fetcher: typeof fetch) {
  return await buildGitHubApiHeaders({
    userAgent: "clawhub/github-skill-sync-settings",
    fetchImpl: fetcher,
    // Installation tokens are repository-scoped and cannot reliably read an
    // arbitrary public repository selected by a publisher.
    useGitHubApp: false,
    useOAuthAppClientCredentials: true,
  });
}
```

Reusing `buildGitHubSkillSourceHeaders` directly would work too, and would keep the two from drifting.

convex/githubSkillSyncSettings.ts
```diff
@@ -0,0 +1,550 @@
...
+    .query("githubOrgMemberships")
+    .withIndex("by_user_and_github_org", (q) =>
+      q.eq("userId", args.userId).eq("githubOrgId", githubOwnerId),
+    )
+    .unique();
+  const now = args.now ?? Date.now();
+  if (
+    !membership ||
+    membership.role !== "admin" ||
+    now - membership.syncedAt > GITHUB_ORG_MEMBERSHIP_VERIFICATION_MAX_AGE_MS
```
Yigtwxx, line 152: This is a statement-for-statement fork of `getPublicGitHubSkillSourceSetupContextInternal` (`convex/githubSkillSync.ts:493-537` on `main`): same `requirePublisherRole({ allowed: ["admin"] })`, same `publisher.linkedUserId !== userId` → `"Forbidden"`, same `getGitHubProviderAccountId` match, same `githubOrgId` + `githubVerifiedAt` gate, same `githubOrgMemberships.by_user_and_github_org` lookup with `role !== "admin"` and the same `GITHUB_ORG_MEMBERSHIP_VERIFICATION_MAX_AGE_MS` staleness check.

That's the authorization boundary for GitHub source enrollment, which makes it the worst place in the file to keep two copies — a future hardening lands in one and quietly misses the other, and nothing fails loudly when it does. Exporting the existing internal query, or lifting the shared body into `convex/lib/`, would keep one definition of who may enrol a repo.

Worth noting this cuts against the file's own habit elsewhere: slug resolution correctly reuses `getSkillBySlugForPublisher` / `getSkillSlugAliasBySlugForPublisher` rather than re-deriving it, so this reads like drift rather than intent.

convex/githubSkillSyncSettings.ts
```diff
@@ -0,0 +1,550 @@
...
+  assertGitHubSkillSyncRuntimeEnabled();
+  const context = await requireActionPublisherContext(ctx, args.publisherId, authOverride);
+  const metadata = await fetchVerifiedRepositoryMetadata(args.repo, context.githubOwnerId, fetcher);
+  const snapshot = await fetchGitHubSkillSourceSnapshot(
+    {
+      repo: metadata.repo,
+      defaultBranch: metadata.defaultBranch,
+    },
+    fetcher,
+  );
```
Yigtwxx, line 481: This call no longer type-checks against current `main`. `fetchGitHubSkillSourceSnapshot` gained a leading `ctx` parameter (`convex/githubSkillSync.ts:2953-2954`):

```ts
async function fetchGitHubSkillSourceSnapshot(
  ctx: Pick<ActionCtx, "runAction">,
  { repo, defaultBranch }: { repo: string; defaultBranch: string },
  fetcher: typeof fetch = fetch,
)
```

It needs `runAction` for the raster-icon validation that landed with it, and every call site on `main` now passes `ctx` first (`:2585`, `:2779`, `:2865`).

That also propagates to this handler's own signature: `previewGitHubSkillSyncRepositoryHandler` types its ctx as `Pick<ActionCtx, "runQuery">`, which no longer satisfies the callee. It would need `Pick<ActionCtx, "runQuery" | "runAction">`, and the same widening on `requireActionPublisherContext` if you thread it through.

Flagging it because the branch is a couple of weeks old and this is the kind of drift that only surfaces at `ci:types-build` time.

convex/githubSkillSyncSettings.ts
```diff
@@ -0,0 +1,550 @@
...
+function parseGitHubNumericId(value: unknown, message: string) {
+  const normalized =
+    typeof value === "number" && Number.isSafeInteger(value) && value > 0
+      ? String(value)
+      : typeof value === "string" && /^[1-9]\d*$/.test(value.trim())
+        ? value.trim()
+        : "";
+  if (!normalized) throw new ConvexError(message);
+  return normalized;
+}
```
Yigtwxx, line 106: Three canonical helpers get re-implemented in this file. Grouping them here rather than spreading the same note around:

**`parseGitHubNumericId`** is `normalizeGitHubNumericId` (`convex/githubSkillSync.ts:3170`) with a different name — same two branches, same `/^[1-9]\d*$/`. And `parseRepositoryMetadata` (`:317` and `:325` in this file) then inlines that same normalization twice more for `row.id` and `owner.id` instead of calling even this local copy, so the logic exists four times in one file.

**`normalizeRepo`** (`:535`) is a copy of `convex/githubSkillSync.ts:3323` with the regex widened to accept `https?`, `www.`, a bare `github.com/` prefix, and to strip `[?#]`. If that widening is right — and it looks right — it belongs on the canonical function. Two parsers that disagree about which strings are valid repo references, applied to the same user input on different screens, is a bug waiting for the right paste.

**`clampInteger`** (`:547`) is byte-for-byte `convex/githubImport.ts:885`, and semantically `clampInt` at `convex/githubSkillSync.ts:202`.

None of these are wrong on their own; it's the aggregate that concerns me, given the file is otherwise disciplined about reuse.

convex/githubSkillSyncSettings.ts
```diff
@@ -0,0 +1,550 @@
...
+  },
+): Promise<GitHubSkillSyncPreviewItem[]> {
+  const publisher = await ctx.db.get(args.publisherId);
+  if (!publisher || publisher.deletedAt || publisher.deactivatedAt) {
+    throw new ConvexError("Publisher not found");
+  }
+  const source = await ctx.db
+    .query("githubSkillSources")
+    .withIndex("by_repo", (q) => q.eq("repo", args.repo))
+    .unique();
```
Yigtwxx, line 245: This resolves the existing source by the mutable repo string, in the one code path that already knows the repo may have just been renamed — `previewGitHubSkillSyncRepositoryHandler` computes `redirected: metadata.requestedRepo.toLowerCase() !== metadata.repo.toLowerCase()` and passes the post-redirect `full_name` in as `args.repo`.

`main` added `by_github_repository_id` for exactly this (`convex/githubSkillSync.ts:543-546`), resolving by immutable id with the repo string as fallback.

Concrete failure: publisher `@acme` has an enrolled source stored as `acme/skills`; the repo is renamed to `acme/agent-skills`; a *different* publisher previews `acme/agent-skills`. The `by_repo` lookup returns `null`, so `source` is `null`, so `resolvePreviewDestination` can never return `source-conflict` — and the already-enrolled skills are reported as clean `new-destination` / `replacement`. That inverts the guarantee the preview screen exists to give.

`metadata.repositoryId` is already in hand at the call site, so mirroring the `main` lookup order should be a small change.

convex/githubSkillSyncSettings.ts
```diff
@@ -0,0 +1,550 @@
...
+    publisherId: Id<"publishers">;
+    page?: number;
+    perPage?: number;
+  },
+  fetcher: typeof fetch = fetch,
+  authOverride?: { userId: Id<"users"> },
+): Promise<GitHubSkillSyncRepositoryListResult> {
+  assertGitHubSkillSyncRuntimeEnabled();
+  const context = await requireActionPublisherContext(ctx, args.publisherId, authOverride);
+  const login = context.githubLogin ?? (await fetchVerifiedOwnerLogin(context, fetcher));
```
Yigtwxx, line 389: The `??` here creates an asymmetry between the two publisher kinds that's worth making deliberate.

For a personal publisher, `context.githubLogin` is always undefined, so `fetchVerifiedOwnerLogin` runs and asserts `id === context.githubOwnerId` against `/user/{id}` — the login is verified against the immutable owner id on every call.

For an org, `githubLogin` is populated from the stored `githubOrgMemberships.login`, so this short-circuits and the login is **never** re-verified against `githubOwnerId`. GitHub org logins are renameable, and that row is only as fresh as `GITHUB_ORG_MEMBERSHIP_VERIFICATION_MAX_AGE_MS` allows. After a rename, `/orgs/{stale-login}/repos` either 404s or — if the old login has been claimed by someone else — lists a different org's repositories, which then get filtered by `context.githubOwnerId` in `toRepositoryListItem` and come back empty. The ownership filter saves you from showing the wrong repos, but the user just sees "No public repositories were returned for this publisher" with no way to understand why.

The org path is also the one path with no test coverage, which is how the asymmetry stayed invisible. Either drop the `??` so both kinds verify, or keep the shortcut and add a test pinning that a stale login yields a recoverable error rather than an empty list.

Other specific branches with no coverage, while this file is open: the `alias-conflict` arm of `resolvePreviewDestination` is only tested through the pure classifier, never at the DB level; `parseRepositoryMetadata`'s `row.private !== false` / `visibility !== "public"` rejection is untested; and the stale-response race test covers the repo change but not `onPublisherChange`, which bumps the same request id.

convex/githubSkillSyncSettings.ts
```diff
@@ -0,0 +1,550 @@
...
+  return {
+    publisher: {
+      _id: context.publisherId,
+      handle: context.publisherHandle,
+      kind: context.publisherKind,
+    },
+    page,
+    perPage,
+    hasMore: body.length === perPage,
+    repositories,
```
Yigtwxx, line 424: `page` / `perPage` / `hasMore` are computed and returned, but the settings screen calls this once with `{ perPage: 100 }` and renders `result.repositories` only — so a publisher with more than 100 public repos has no way to reach repo 101. Either wire a "Load more" or drop the three fields from the return type until something consumes them; shipping an unused pagination contract in a `ReturnType<typeof action>` is harder to remove later than to not add.

There's also a mismatch between the two: `hasMore` is derived from the raw `body.length`, while `repositories` is the *filtered* list. A page of 100 repos that all fail the `expectedOwnerId` / `private` / `visibility` checks renders zero rows with `hasMore: true`, and the UI shows "No public repositories were returned for this publisher" — which is misleading, since there are more pages. Deriving the empty-state copy from `hasMore` as well as `repositories.length` would separate "you have none" from "none on this page".

convex/githubSkillSyncSettings.ts
```diff
@@ -0,0 +1,550 @@
...
+  const snapshot = await fetchGitHubSkillSourceSnapshot(
+    {
+      repo: metadata.repo,
+      defaultBranch: metadata.defaultBranch,
+    },
+    fetcher,
+  );
+  if (snapshot.skills.length === 0) {
+    throw new ConvexError("No skills were found in that public GitHub repo.");
+  }
```
Yigtwxx, line 484: "This repo contains no skills" is a legitimate answer to "what would happen if I synced this repo", not a failure — but modelling it as `ConvexError` means the settings screen turns it into `toast.error`.

That lands harder than it looks because the repository list auto-selects the first selectable repo, so a publisher whose most-recently-pushed repo happens to have no `SKILL.md` gets an error toast immediately on opening the screen, before touching anything. An empty preview with `total: 0` and the existing summary counts at zero would say the same thing without reading as a fault.

Related: `archived` and `fork` are parsed into `GitHubRepositoryListItem`, asserted `selectable: true` in the backend test, and then never rendered — and `fetchVerifiedRepositoryMetadata` only rejects `disabled`. So a publisher can select an archived fork with no signal anywhere in the flow. Either surface them as badges in the list or drop them from the type, so the data and the UI agree on what matters.

src/routes/settings.tsx
```diff
@@ -339,22 +351,18 @@ export function Settings() {
   const manageablePublishers = (publisherMemberships ?? []).filter(
     (entry) => entry.role !== "publisher",
   );
-  const officialGitHubSourcePublishers = manageablePublishers.filter(
-    (entry) => entry.publisher.official === true,
-  );
+  const githubSourcePublishers = manageablePublishers;
```
Yigtwxx, line 354: This drops the `official` publisher gate, and the PR body doesn't mention it.

Before, this was `manageablePublishers.filter((entry) => entry.publisher.official === true)`, and the query below moved from `listForManageableOfficialPublishers` — which calls `isOfficialPublisher` (`convex/githubSkillSources.ts:182`) — to `listForPublisher`, which gates on the admin role only. Neither of the two new actions checks `official` either.

This may well be correct: the configure action on `main` no longer gates on `official` at all, so the UI filter could be the last holdout of a policy that has already moved. But it's a change to who is eligible to enrol a GitHub source, riding inside a PR titled "repository preview", with no test asserting the new audience. Worth a line in the description and a `-settings.test.tsx` case for a managed non-official publisher, so the next person reading this diff doesn't have to reconstruct whether it was intentional.

Yigtwxx (commented):
Read through this against `main` at `a9d04bb0`. Some of the structure here is genuinely nice: the classifier lives in a pure, dependency-free module with a discriminated-union destination and an `assertNever` in the UI, so the interesting logic is directly unit-testable; `assertGitHubSkillSyncRuntimeEnabled()` is the first statement in both public actions with a test proving no `fetch` and no `runQuery` happen when the rollout is off, which is a well-constructed fail-closed test rather than a nominal one; the `useRef` request-id guard plus the `cancelled` flag is a correct double guard with coverage; and pinning `snapshot.commit` into the preview is the right call for a "what will happen" screen.

Three things I'd want resolved before this leaves draft, in order:

1. `buildGitHubSettingsHeaders` omits `useGitHubApp`, which is opt-out rather than opt-in, so every new GitHub call authenticates as the App installation — the sibling helper in `githubSkillSync.ts` disables that explicitly and documents why. The `fetchMock`-based tests can't see it.
2. `fetchGitHubSkillSourceSnapshot` gained a leading `ctx` parameter on `main`, so this branch won't type-check as-is, and the fix propagates into the handler's own ctx type.
3. The publisher-authorization block is a statement-for-statement fork of `getPublicGitHubSkillSourceSetupContextInternal`. Two copies of an authorization boundary is the one duplication that tends to bite silently.

Six more inline: three smaller helpers duplicated from elsewhere in `convex/`, a source lookup by mutable repo string in the path that already detects renames, an org/personal asymmetry in login verification (plus the specific untested branches), unused pagination whose `hasMore` disagrees with the filtered list, "no skills found" modelled as an error on an auto-selected repo, and the silently dropped `official` publisher gate in `settings.tsx`.

Patrick-Erichsen (commented):
This draft adds a repository-preview flow for GitHub Skill Sync. It should stay draft: current main changed the snapshot contract and public-source credential policy, while this branch would remove the only shipped enrollment control before its replacement is active.

LOC: +1775/-142 (9 files)

Findings: rebase onto the current action-context API; retain existing source enrollment until preview hands off to an implemented activation path; keep installation tokens disabled for arbitrary public repositories; decide Official-only versus all verified publishers; add real-browser proof for selection, results, conflict, and unavailable states.

Best-fix verdict: too broad and stale. Preview is useful, but it must be layered onto the current immutable snapshot/source contract rather than replacing enrollment prematurely.

Alternatives considered: keep preview read-only alongside the existing enrollment control, then remove the old path only in the later activation slice.

Code read: settings preview UI, GitHub request helpers, snapshot discovery action/query boundary, current enrollment flow, and current-main public-source header policy.

Remaining uncertainty: intended eligibility boundary and activation design.

steipete:
@clawsweeper re-review

PR #3232, opened 2026-07-23
fix(deps): update Next.js to 16.2.11

giodl73-repo:
## Summary

- update the existing Next.js override from 16.2.6 to 16.2.11
- refresh the Bun lockfile entries for Next.js and its platform packages
- clear the new Next.js advisories currently blocking the shared PR gate

## Verification

- `bun audit` — no vulnerabilities found
- `git diff --check`

## Context

The ClawHub main lockfile currently resolves Next.js 16.2.6, which is within the vulnerable range reported by `bun audit`. Because `pr-gates` runs that audit, unrelated PRs currently fail their aggregate checks. This keeps the remediation isolated from those feature PRs.


giodl73-repo:
Superseded by #3234, which landed the same Next.js 16.2.11 override and lockfile patch on main. git cherry origin/main d7d4a61f reports this commit as already applied, so this is no longer a dependency of the feed stack.

Issue #3275, opened 2026-07-27
package publish fails on npm 12: 'npm pack did not return a tarball filename' (pack --json is now an object, not an array)


mir-stream:
## Summary

`clawhub package publish` fails on npm 12 with `npm pack did not return a tarball filename`. The CLI reads `npm pack --json` output as an array, but npm 12 changed that payload to an object keyed by package name.

Affects `clawhub@0.23.1` (current latest). Any CI runner that has moved to npm 12 hits this on every publish.

## Cause

`dist/cli/commands/packages.js` (0.23.1):

```js
const result = spawnSync("npm", ["pack", options.sourcePath, "--json", "--ignore-scripts",
                                 "--pack-destination", options.packDestination], { cwd: options.cwd, encoding: "utf8" });
...
npmOutput = JSON.parse(result.stdout);
const filename = npmOutput[0]?.filename;
if (!filename) fail("npm pack did not return a tarball filename");
```

`npmOutput[0]` is `undefined` under npm 12, so the guard fires even though the pack itself succeeded and the tarball is sitting in `--pack-destination`.

## Reproduction

Same package, same flags, only the npm version differs:

```console
$ npm@11 pack ./packages/my-plugin --json --ignore-scripts --pack-destination /tmp/out
[
  {
    "id": "my-plugin@0.4.0",
    "filename": "my-plugin-0.4.0.tgz",
    ...
  }
]

$ npm@12 pack ./packages/my-plugin --json --ignore-scripts --pack-destination /tmp/out
{
  "my-plugin": {
    "id": "my-plugin@0.4.0",
    "filename": "my-plugin-0.4.0.tgz",
    ...
  }
}
```

Evaluating what the CLI evaluates:

| npm | `JSON.parse(stdout)[0]?.filename` |
|---|---|
| 11 | `my-plugin-0.4.0.tgz` |
| 12 | `undefined` |

## Suggested fix

Accept both shapes, e.g.:

```js
const parsed = JSON.parse(result.stdout);
const entry = Array.isArray(parsed) ? parsed[0] : Object.values(parsed)[0];
const filename = entry?.filename;
```

That keeps working on npm 10/11 and fixes npm 12 without pinning.

Worth noting the failure is a little misleading: `npm pack` did produce the tarball, so the message points away from the actual cause. Including the raw stdout in the error, or asserting on the parsed shape, would shorten the next diagnosis.

## Workaround

[...]

PR #3276, opened 2026-07-27
fix(ci): accept npm 12 pack --json object shape in release

SebTardif:
## What Problem This Solves

The clawhub CLI release workflow packs the CLI with `npm pack --json`. npm 11 returns an **array**; npm 12 returns a **package-keyed object**. The workflow only accepted an array (`Array.isArray(parsed) ? parsed[0] : null`), so release packing fails on npm 12 runners even when the tarball is produced.

The product CLI path in `packages/clawhub/src/cli/commands/packages.ts` already dual-parses both shapes (tracked under Unreleased / related to #3275). This PR aligns the **release workflow** with that behavior.

## Evidence

Dual-shape parse (same logic as the workflow step):

```console
$ node --input-type=module -e '
const parse = (s) => {
  const parsed = JSON.parse(s);
  const first = Array.isArray(parsed) ? parsed[0]
    : (parsed && typeof parsed === "object" ? Object.values(parsed)[0] : null);
  if (!first || typeof first.filename !== "string") process.exit(1);
  process.stdout.write(first.filename);
};
console.log(parse(JSON.stringify([{filename:"a.tgz"}])));
console.log(parse(JSON.stringify({pkg:{filename:"b.tgz"}})));
'
a.tgz
b.tgz
```

## Real behavior proof

- **Behavior or issue addressed:** clawhub CLI npm release packing must accept npm 12 object-shaped `npm pack --json` output.
- **Real environment tested:** macOS arm64, Node dual-shape parse of the exact expression used in the workflow.
- **Exact steps or command run after this patch:** node one-liner above with array and object fixtures.
- **Evidence after fix:** both shapes yield a tarball filename.
- **Observed result after fix:** workflow step no longer returns null for object-shaped npm 12 output.
- **What was not tested:** full GitHub Actions release job on npm 12 (requires release secrets).

## Summary

- One-line dual-shape parse in `clawhub-cli-npm-release.yml`
- Refs #3275 (CLI packages.ts already fixed on main; published npm still needs a release for end users)

Closes nothing until a CLI release ships for #3275 user-facing path; this PR only unblocks the release workflow on npm 12.

### DIFF
--- a/.github/workflows/clawhub-cli-npm-release.yml
+++ b/.github/workflows/clawhub-cli-npm-release.yml
@@ -140,7 +140,7 @@ jobs:
           pushd "$PACKAGE_DIR" >/dev/null
           PACK_JSON="$(npm pack --json --ignore-scripts)"
           echo "$PACK_JSON"
-          PACK_PATH="$(printf '%s\n' "$PACK_JSON" | node --input-type=module -e 'const chunks=[]; process.stdin.on("data", (chunk) => chunks.push(chunk)); process.stdin.on("end", () => { const parsed = JSON.parse(Buffer.concat(chunks).toString("utf8")); const first = Array.isArray(parsed) ? parsed[0] : null; if (!first || typeof first.filename !== "string" || !first.filename) process.exit(1); process.stdout.write(first.filename); });')"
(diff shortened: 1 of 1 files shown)


Patrick-Erichsen:
Nice fix, thank you!

PR #2756, opened 2026-06-19
Refresh homepage experience

vyctorbrzezowski:
Preview: https://clawhub-ui-design-refactor-4fk7c9ne7.vercel.app

## Summary

- Refreshes the homepage hero, listing, popular creators, app shortcuts, BYOS/CLI section, and global footer styling.
- Adds homepage skill/plugin listing controls with Top/Trending/New and Official/Top/New tabs, category filtering, search, pagination, and mobile layout polish.
- Standardizes the footer across pages and removes the secondary footer variant.
- Optimizes the footer/BYOS artwork path with smaller WebP assets.

## Tests

- `bun run test -- src/__tests__/home-route.test.tsx src/__tests__/ui-design-contract.test.ts src/__tests__/home-listing-section.test.tsx src/__tests__/packages-route.test.tsx`
- `bun run test -- src/components/Footer.test.tsx`
- `bun run ci:static`
- `bun run ci:unit`
- `bun run build`
- `vercel deploy --yes --force`
- `curl -I -L https://clawhub-ui-design-refactor-4fk7c9ne7.vercel.app/` → `HTTP/2 200`

## Review

- `codex --config service_tier=fast review --base upstream/main` found category/search pagination issues; fixed in this branch.
- Final rerun was attempted, but the Codex CLI hit usage quota before completing.

PR #3281, opened 2026-07-27
fix(home): contain app category scroller on mobile

vyctorbrzezowski:
## Summary

- constrain the mobile app-category strip to the available content width while preserving its internal horizontal scroll
- add a focused CSS contract regression test scoped to the exact mobile rule
- preserve the deliberate Skills Trending behavior from #3270: categories remain available on New, Featured, and Official, but not on the canonical Trending feed

## Root cause

The category strip keeps its desktop `width: fit-content` after becoming a horizontal scroller on mobile. At a 375 px page width, that makes the strip 449 px wide and expands the document to 469 px.

In this PR, the strip is proposed to fill its 335 px content area while its 493 px of category tabs remain scrollable inside it.

## Validation

- `bunx vitest run src/__tests__/ui-design-contract.test.ts` (18 passed)
- mutation check: removing `min-width: 0` fails the scoped contract on that declaration
- `bun run ci:static`
- `bun run ci:unit` (5,878 passed, 2 skipped)
- autoreview: no accepted/actionable findings
- real-browser mobile check at a 390 × 844 viewport: document width 469 px before, 375 px in this PR

## UI proof

See the [published before/after screenshots](https://github.com/openclaw/clawhub/pull/3281#issuecomment-5095721395).

### DIFF
--- a/src/__tests__/ui-design-contract.test.ts
+++ b/src/__tests__/ui-design-contract.test.ts
@@ -19,8 +19,7 @@ function cssRule(css: string, selector: string) {
 function cssMediaContaining(css: string, query: string, required: readonly string[]) {
   let start = css.indexOf(`@media ${query}`);
   while (start >= 0) {
-    const nextMedia = css.indexOf("@media ", start + 1);
-    const block = css.slice(start, nextMedia === -1 ? undefined : nextMedia);
+    const block = balancedCssBlock(css, start, `@media ${query}`);
     if (required.every((snippet) => block.includes(snippet))) return block;
     start = css.indexOf(`@media ${query}`, start + 1);
   }
@@ -34,6 +33,22 @@ function cssBlock(css: string, selector: string) {
   const end = css.indexOf("\n}", start);
   expect(end, `Unclosed CSS block for ${selector}`).toBeGreaterThan(start);
   return css.slice(start, end + 2);
+}
+
+function balancedCssBlock(css: string, start: number, label: string) {
+  const open = css.indexOf("{", start);
+  expect(open, `Missing opening brace for ${label}`).toBeGreaterThanOrEqual(0);
+
+  let depth = 0;
+  for (let index = open; index < css.length; index += 1) {
+    if (css[index] === "{") depth += 1;
+    if (css[index] !== "}") continue;
+
+    depth -= 1;
+    if (depth === 0) return css.slice(start, index + 1);
+  }
+
+  throw new Error(`Unclosed CSS block for ${label}`);
 }
 
 function tokenValue(css: string, selector: string, token: string) {
@@ -386,13 +401,15 @@ describe("restored UI design contract", () => {
   });
 
   it("keeps the mobile app-category scroller within the page width", () => {
-    const css = styles();
-
-    cssMediaContaining(css, "(max-width: 760px)", [
-      ".home-v2-apps-categories {\n    width: 100%;",
-      "min-width: 0;",
-      "overflow-x: auto;",
+    const mobileBlock = cssMediaContaining(styles(), "(max-width: 760px)", [
+      ".home-v2-apps-categories {",
     ]);
+    const scroller = mobileBlock.match(/\.home-v2-apps-categories \{([^}]*)\}/)?.[1];
+
+    expect(scroller, "Missing .home-v2-apps-categories rule").toBeTruthy();
+    expect(scroller).toMatch(/width:\s*100%/);
+    expect(scroller).toMatch(/min-width:\s*0/);
+    expect(scroller).toMatch(/overflow-x:\s*auto/);
   });
 
   it("requires the restored footer columns and mobile section toggles", () => {
--- a/src/__tests__/ui-design-contract.test.ts
+++ b/src/__tests__/ui-design-contract.test.ts
@@ -385,6 +385,16 @@ describe("restored UI design contract", () => {
     expect(cssRule(css, ".home-v2-listing-row::before")).toContain("border-radius: 0");
   });
 
+  it("keeps the mobile app-category scroller within the page width", () => {
+    const css = styles();
+
+    cssMediaContaining(css, "(max-width: 760px)", [
+      ".home-v2-apps-categories {\n    width: 100%;",
+      "min-width: 0;",
+      "overflow-x: auto;",
+    ]);
+  });
+
   it("requires the restored footer columns and mobile section toggles", () => {
     const footerSource = footer();
     const navSource = navItems();
--- a/src/styles.css
+++ b/src/styles.css
@@ -27851,6 +27851,8 @@ a.home-v2-apps-see-all:focus-visible svg {
   }
 
   .home-v2-apps-categories {
+    width: 100%;
+    min-width: 0;
     flex-wrap: nowrap;
     overflow-x: auto;
     overflow-y: hidden;


vyctorbrzezowski:
## ClawHub UI Proof

Status: `pass`
Mode: `before-after`
Scenario: `final rebased mobile validation`
Provider: `local-browser`
Baseline: `https://clawhub.ai/`
Candidate: `6c506f2aac3d6d42ec719087618cd008be5a1cd9`

| Published clawhub.ai after settling: page-level horizontal overflow | Proposed fix after settling: category scroll stays inside content width |
| --- | --- |
| Published clawhub.ai after settling: page-level horizontal overflow | Proposed fix after settling: category scroll stays inside content width |

Raw proof files: https://github.com/openclaw/clawhub/tree/qa-artifacts/clawhub-ui-proof/pr-3281/manual

src/__tests__/ui-design-contract.test.ts
```diff
@@ -394,6 +394,16 @@ describe("restored UI design contract", () => {
...
 
+  it("keeps the mobile app-category scroller within the page width", () => {
+    const css = styles();
+
+    cssMediaContaining(css, "(max-width: 760px)", [
+      ".home-v2-apps-categories {\n    width: 100%;",
+      "min-width: 0;",
+      "overflow-x: auto;",
+    ]);
+  });
```
Yigtwxx, line 413: Two of these three assertions can't fail, which matters because this is the only thing pinning the fix.

`cssMediaContaining` (lines 19-29) checks each snippet independently against the **entire** `@media (max-width: 760px)` block — it returns the block when `required.every((snippet) => block.includes(snippet))`, and that block spans hundreds of lines and dozens of selectors. So `"min-width: 0;"` and `"overflow-x: auto;"` are satisfied by any rule anywhere in it and aren't tied to `.home-v2-apps-categories` at all. Deleting `min-width: 0` from the scroller would very likely leave this test green — and that's the declaration doing the non-obvious half of the fix, since `width: 100%` on its own is still floored by the flex automatic minimum size.

The third assertion has the opposite failure mode: `".home-v2-apps-categories {\n    width: 100%;"` pins the indentation *and* requires `width` to be the first declaration in the rule, so a reformat or an alphabetical reorder breaks a behaviour that still works.

Scoping to the rule fixes both. One gotcha: the file's `cssBlock` helper won't work here — it terminates on `"\n}"` and a rule nested inside a media block closes with `"\n  }"`, so it would run past the rule to the media query's own closing brace.

```suggestion
  it("keeps the mobile app-category scroller within the page width", () => {
    const mobileBlock = cssMediaContaining(styles(), "(max-width: 760px)", [
      ".home-v2-apps-categories {",
    ]);
    const scroller = mobileBlock.match(/\.home-v2-apps-categories \{([^}]*)\}/)?.[1];

    expect(scroller, "Missing .home-v2-apps-categories rule").toBeTruthy();
    expect(scroller).toMatch(/width:\s*100%/);
    expect(scroller).toMatch(/min-width:\s*0/);
    expect(scroller).toMatch(/overflow-x:\s*auto/);
  });
```

src/styles.css
```diff
@@ -27568,6 +27568,8 @@ a.home-v2-apps-see-all:focus-visible svg {
   }
 
   .home-v2-apps-categories {
+    width: 100%;
+    min-width: 0;
```
Yigtwxx, line 27855: Worth naming what these two declarations change beyond containing the overflow: they are what makes this strip a *real* scroller for the first time.

Before, the `width: fit-content` inherited from the base rule (`src/styles.css:27007` on `main`) meant `scrollWidth === clientWidth`, so `overflow-x: auto`, `scroll-snap-type: x proximity`, the `scroll-snap-align: start` on the tabs and the scrollbar hiding just below were all inert — the strip widened the document instead, which is exactly the bug. With `width: 100%` they all become live.

That matters because `HomeAppsSection.tsx:280` renders this container as `role="tablist"` with `aria-selected` on the active tab (`:288`), and there is no `scrollIntoView` anywhere in the component. While the active category is always the first one this is invisible, but any path that mounts with a different `activeCategoryId` — a deep link, a persisted selection, a future non-zero default — now clips the selected tab out of view with no indication it exists, where previously it was at least reachable by scrolling the page. An effect calling `scrollIntoView({ inline: "nearest", block: "nearest" })` when `activeCategoryId` changes would cover it; worth checking it doesn't fight `scroll-snap-type: proximity`.

Separately, and not something I would hold this PR for: `.home-v2-apps-category-tab:focus-visible` (`src/styles.css:27043-27048`) sets `outline: none` and styles focus identically to `:hover`, which is in turn identical to `[aria-selected="true"]` at `:27055-27058` — so a keyboard user can't distinguish focused from selected. If a focus ring is ever restored there, the `overflow-y: hidden` on this rule will now genuinely clip it vertically, since the container has no vertical padding and the tabs carry a `min-height`. Adjacent to your change rather than caused by it, but this PR is what makes the clipping real.

Yigtwxx (commented):
Checked this against `main` at `a9d04bb0` and the diagnosis holds: the base rule sets `width: fit-content` (`src/styles.css:27007`) and the mobile override at `:27747` changes `flex-wrap` to `nowrap` without resetting it, so the shrink-to-fit width equals the full unwrapped row width and the element widens the document. Both declarations are load-bearing — `min-width: 0` defuses the flex automatic minimum size, which `width: 100%` alone does not, so this also holds at 320px and not just the 375/390px widths in the description. RTL is unaffected since none of the three properties are physical. The fix looks right to me, and reaching for the width rather than `overflow-x: hidden` on a wrapper is the reason the internal scroll still works.

Two non-blocking notes inline: the new test's assertions are scoped to the whole media block rather than to the rule, so two of the three can't fail; and now that the strip actually scrolls, the selected tab has no `scrollIntoView`.

src/__tests__/ui-design-contract.test.ts
vyctorbrzezowski: Fixed in 6c506f2a. I made cssMediaContaining return the brace-balanced media block, then scoped the assertions to the nested .home-v2-apps-categories rule. A mutation removing min-width: 0 now fails specifically on that assertion; the restored test passes all 18 cases.

src/styles.css
vyctorbrzezowski: Checked the state path after rebasing: activeCategoryId initializes to the first category and can only change by clicking an already-visible tab; there is no prop, deep link, or persisted non-zero selection today. I kept scrollIntoView and the adjacent focus treatment out of this overflow fix.

vyctorbrzezowski:
This keeps the mobile app-category strip inside the page while preserving the strip's own horizontal scroll. It matters because the existing mobile rule widened the whole document instead of activating the intended internal scroller.

LOC: +31/-2 (2 files)

No blocking findings.

Affected surface: ClawHub homepage app-category strip at widths up to 760 px.

Evidence checked: current `main`, final branch diff, focused red/green mutation, 18-test UI contract suite, full static gate, full unit/coverage gate (5,878 passed, 2 skipped), clean autoreview, refreshed real-browser before/after proof, and exact-head CI on final SHA `6c506f2a`.

Exact-head CI: successful on attempt 3. The `moderation-malicious` shard had two unrelated runner-state failures (`1.0.1` reused after a retry, then a 1s `users:ensure` timeout) before passing unchanged; Vercel remains the expected fork authorization failure without a deployment run.

Best-fix verdict: best. Constraining the scroller itself with `width: 100%` and `min-width: 0` fixes the flex sizing root cause while retaining `overflow-x: auto` and scroll snapping.

Alternatives considered: `overflow-x: hidden` on a page wrapper would mask the symptom and clip content; removing `width: fit-content` globally would change the desktop wrapped layout; adding `scrollIntoView` is unnecessary here because the active category always initializes to the first tab and only changes through clicks on visible tabs.

Code read: `src/styles.css`, `src/components/HomeAppsSection.tsx`, `src/lib/homeApps.ts`, `src/routes/index.tsx`, and `src/__tests__/ui-design-contract.test.ts`.

Provenance: the mobile scroller behavior originated in #2756 (code/PR author @vyctorbrzezowski, merged by @Patrick-Erichsen on 2026-06-20); `width: fit-content`, which created the regression in combination with that mobile rule, came from #3059 (code/PR author and merger @vyctorbrzezowski on 2026-07-11). Current PR author: @vyctorbrzezowski.

Remaining uncertainty: none for the reported overflow. A future non-first externally controlled active category could merit `scrollIntoView`, but no such state path exists today.

PR #3283, opened 2026-07-27
fix(catalog): show Official skills during backend rollout

vyctorbrzezowski:
## What Problem This Solves

Fixes an issue where users opening the Official skills feed would see an empty
state while the frontend and Convex backend were on different catalog rollout
versions.

## Why This Change Was Made

In this PR, the frontend checks the catalog capability before sending the new
Official filter and falls back to a bounded client-side scan for older backends.
The v2 backend treats both skill badges and verified publisher ownership as
Official.

## User Impact

Users can browse Official skills during staggered frontend/backend deploys, and
skills published by verified official organizations appear in the feed.

## Evidence

- Production reproduction in a 390 × 844 browser viewport: the selected Official
  feed renders `No skills found`.
- Candidate proof on a real local ClawHub instance with a legacy rollout backend:
  the Official feed renders the deterministic `@catalog-atlas` fixture.
- `bunx vitest run convex/rolloutCapabilities.test.ts convex/skills.publicListCursor.test.ts src/lib/catalogDiscoveryCapabilities.test.ts src/__tests__/skills-index.claw591.test.tsx src/lib/homeListingData.claw591.test.ts` — 77 passed.
- `bun run ci:static`
- `bun run ci:unit` — 5,598 passed, 2 skipped.
- `bun run ci:types-build`
- `.agents/skills/autoreview/scripts/autoreview --mode local --codex-bin /Applications/ChatGPT.app/Contents/Resources/codex` — no accepted/actionable findings.

| Production: Official feed is empty | Candidate: official publisher skill is visible |
| --- | --- |
| Production Official skills feed showing an empty state on mobile | Candidate Official skills feed showing an official publisher skill on mobile |

[Raw proof files](https://github.com/openclaw/clawhub/tree/qa-artifacts/clawhub-ui-proof/pr-3283/2026-07-27-official-skills-mobile)

### DIFF
--- a/convex/skills.ts
+++ b/convex/skills.ts
@@ -5606,6 +5606,7 @@ export const listPublicPageV4 = query({
       numItems,
       Math.min(MAX_FILTERED_PUBLIC_LIST_SCAN_ROWS, numItems * 12),
     );
+    const officialPublisherCache = new Map<string, boolean>();
 
     for (let pageCount = 0; pageCount < MAX_FILTERED_PUBLIC_LIST_SCAN_PAGES; pageCount += 1) {
       if (remainingRows <= 0) break;
@@ -5639,19 +5640,23 @@ export const listPublicPageV4 = query({
         ) {
           return { page: items, hasMore: false, nextCursor: null };
         }
-        if (
-          (!args.officialOnly || isSkillCatalogOfficial(digest)) &&
+        const passesFilters =
           (typeof args.createdAfter !== "number" || digest.createdAt >= args.createdAfter) &&
           digestPassesPublicListFilters(digest, {
             categorySlug,
             topic,
             categoryKeywords,
             excludeCategoryKeywords,
-          })
+          });
+        if (!passesFilters) continue;
+        if (
+          args.officialOnly &&
+          !(await isOfficialSkillBrowseDigest(ctx, digest, officialPublisherCache))
         ) {
-          const item = await buildPublicSkillEntryFromDigest(ctx, digest);
-          if (item) items.push(item);
+          continue;
         }
+        const item = await buildPublicSkillEntryFromDigest(ctx, digest);
+        if (item) items.push(item);
         if (items.length >= numItems) {
           const nextDigest = result.page[index + 1];
           if (
@@ -6399,6 +6404,21 @@ function decodeSkillCatalogCursor(raw: string | null | undefined): SkillCatalogC
 
 function isSkillCatalogOfficial(digest: Doc<"skillSearchDigest">) {
   return Boolean(digest.badges?.official);
+}
+
+async function isOfficialSkillBrowseDigest(
+  ctx: Pick<QueryCtx, "db">,
+  digest: Doc<"skillSearchDigest">,
+  publisherCache: Map<string, boolean>,
+) {
+  if (isSkillCatalogOfficial(digest)) return true;
+  if (!digest.ownerPublisherId) return false;
+  const publisherId = String(digest.ownerPublisherId);
+  const cached = publisherCache.get(publisherId);
+  if (cached !== undefined) return cached;
+  const official = await hasOfficialPublisherRow(ctx, digest.ownerPublisherId);
+  publisherCache.set(publisherId, official);
+  return official;
 }
 
 function isCuratedSkillDigest(digest: Pick<Doc<"skillSearchDigest">, "badges">) {
--- a/src/lib/badges.ts
+++ b/src/lib/badges.ts
@@ -5,6 +5,7 @@ type BadgeKind = Doc<"skillBadges">["kind"];
 type SkillBadgeMap = Partial<Record<BadgeKind, { byUserId: Id<"users">; at: number }>>;
 
 type SkillLike = { badges?: SkillBadgeMap | null };
+type SkillOwnerLike = object | null | undefined;
 
 type BadgeLabel = "Deprecated" | "Official";
 
@@ -14,6 +15,13 @@ export function isSkillHighlighted(skill: SkillLike) {
 
 export function isSkillOfficial(skill: SkillLike) {
   return Boolean(skill.badges?.official);
+}
+
+export function isOfficialSkillListing(skill: SkillLike, owner?: SkillOwnerLike) {
+  return (
+    isSkillOfficial(skill) ||
+    (owner !== null && owner !== undefined && "official" in owner && owner.official === true)
+  );
 }
 
 export function isSkillDeprecated(skill: SkillLike) {
--- a/src/lib/homeListingData.ts
+++ b/src/lib/homeListingData.ts
@@ -1,5 +1,6 @@
 import { api } from "../../convex/_generated/api";
 import { convexHttp } from "../convex/client";
+import { isOfficialSkillListing } from "./badges";
 import { fetchCatalogDiscoveryCapabilities } from "./catalogDiscoveryCapabilities";
 import { getSkillCategoriesForSkill } from "./categories";
 import { fetchPluginCatalog, type PackageListItem } from "./packageApi";
@@ -53,6 +54,7 @@ export const HOME_NEW_WINDOW_MS = 14 * 24 * 60 * 60 * 1_000;
 
 const PLUGIN_CATALOG_PAGE_LIMIT = 100;
 const LEGACY_NEW_PLUGIN_MAX_REQUESTS = 10;
+const LEGACY_OFFICIAL_SKILL_MAX_REQUESTS = 10;
 // Featured is intentionally a finite editorial feed: the latest 40 badge-history rows.
 const FEATURED_SKILL_LIMIT = 40;
 
@@ -138,7 +140,10 @@ export async function fetchHomeSkillListing(
   // highlightedOnly is a dedicated backend path ordered by skillBadges.by_kind_at;
   // the nominal sort below is ignored for Featured and never chooses its candidate set.
   const capabilities =
-    tab === "new" ? await fetchCatalogDiscoveryCapabilities() : { apiVersion: 1 as const };
+    tab === "new" || tab === "official"
+      ? await fetchCatalogDiscoveryCapabilities()
+      : { apiVersion: 2 as const };
+  const useBackendOfficialFilter = tab === "official" && capabilities.apiVersion >= 2;
   const newestCutoff = Date.now() - HOME_NEW_WINDOW_MS;
   const requestLimit = tab === "featured" ? FEATURED_SKILL_LIMIT : numItems;
   const categoriesToFetch = categorySlugs.length > 0 ? categorySlugs : [null];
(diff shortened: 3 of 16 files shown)


vyctorbrzezowski:
## ClawHub UI Proof

Status: `passed`
Mode: `before-after`
Scenario: `Official skills mobile browse`
Provider: `real browser`
Baseline: `https://clawhub.ai/skills?tab=official`
Candidate: `brzezowski/fix-official-skills-browse`

| Production: Official feed is empty | Candidate: official publisher skill is visible |
| --- | --- |
| Production: Official feed is empty | Candidate: official publisher skill is visible |

Raw proof files: https://github.com/openclaw/clawhub/tree/qa-artifacts/clawhub-ui-proof/pr-3283/2026-07-27-official-skills-mobile

convex/skills.ts
```diff
@@ -6401,6 +6406,21 @@ function isSkillCatalogOfficial(digest: Doc<"skillSearchDigest">) {
...
+) {
+  if (isSkillCatalogOfficial(digest)) return true;
+  if (!digest.ownerPublisherId) return false;
+  const publisherId = String(digest.ownerPublisherId);
+  const cached = publisherCache.get(publisherId);
+  if (cached !== undefined) return cached;
+  const official = await hasOfficialPublisherRow(ctx, digest.ownerPublisherId);
+  publisherCache.set(publisherId, official);
+  return official;
+}
```
Yigtwxx, line 6422: Heads up before you spend more time on this one: `main` moved under it in a way that makes this function unreachable for the feed it targets.

#3314 ("fix: load Official skills from curated index") merged on 2026-07-30, three days after this PR was opened, and added an early return to `listPublicPageV4`:

- `convex/skills.ts:5495` — `const officialCursor = args.officialOnly ? decodeOfficialSkillCursor(args.cursor) : null;`
- `convex/skills.ts:5558` — `if (args.officialOnly && officialCursor?.projection === "curated") { return await listCuratedSkillPage(...) }`

and `decodeOfficialSkillCursor` returns `{ projection: "curated", cursor: null }` for a **missing** cursor (`:5437-5438`). So after a rebase, every first-page Official request returns from the curated branch and never reaches the generic scan loop where `isOfficialSkillBrowseDigest` is called.

That leaves the frontend in the worse of the two positions: `rolloutCapabilities.ts` still advertises `apiVersion: 2`, so the client stops filtering on the strength of a backend guarantee that is no longer being kept.

There's a second layer to it. `listCuratedSkillPage` scans the curated projection and applies the badge-only predicate at `convex/skills.ts:7354` (`!opts.officialOnly || isSkillCatalogOfficial(digest)`), and `convex/lib/skillSearchDigest.test.ts:390` ("removes the curated projection when the last curated badge is cleared") implies a publisher-official-but-unbadged skill isn't in that projection at all — so it can't be surfaced there regardless of which predicate runs.

If the product intent is still "skills owned by official publishers belong in the Official feed", the change probably has to move into how the curated projection is maintained (`convex/lib/skillSearchDigest.ts`) rather than into the scan loop, or `:7354` has to become this predicate. Worth rebasing and re-confirming the production reproduction first — it's possible #3314 already resolved the empty-feed symptom by different means, which would narrow this PR to the staggered-deploy client fallback.

convex/skills.ts
```diff
@@ -5606,6 +5606,7 @@ export const listPublicPageV4 = query({
       numItems,
       Math.min(MAX_FILTERED_PUBLIC_LIST_SCAN_ROWS, numItems * 12),
     );
+    const officialPublisherCache = new Map<string, boolean>();
```
Yigtwxx, line 5609: The reordering just below this — computing `passesFilters` first, `continue`-ing early, and only then doing the async publisher lookup — is the right call, and easy to get backwards. The cheap synchronous filters now gate the DB read instead of running alongside it.

The cache itself may be redundant, though. `buildPublicSkillEntryFromDigest` (`convex/skills.ts:6128-6150`) already calls `addOfficialStatusToOwnerInfo` for **every** entry it returns, and that in turn calls `hasOfficialPublisherRow(ctx, publisherId)` — the same lookup. Since this `Map` is local to `isOfficialSkillBrowseDigest` and isn't threaded into the existing helper, each publisher-official skill in the page costs two `officialPublishers` reads rather than one.

The simplest way out is to drop the pre-filter and filter on the built entry's `entry.owner.official` instead — which is exactly what the frontend fallback in this same PR does via `isOfficialSkillListing(entry.skill, entry.owner)`. That removes this function, the cache, and the duplicate read in one go. If pre-filtering has to stay for scan-budget reasons, then threading the same `Map` into `addOfficialStatusToOwnerInfo` would at least make it pay for itself.

src/lib/badges.ts
```diff
@@ -16,6 +17,13 @@ export function isSkillOfficial(skill: SkillLike) {
   return Boolean(skill.badges?.official);
 }
 
+export function isOfficialSkillListing(skill: SkillLike, owner?: SkillOwnerLike) {
+  return (
+    isSkillOfficial(skill) ||
+    (owner !== null && owner !== undefined && "official" in owner && owner.official === true)
+  );
+}
```
Yigtwxx, line 25: Extracting this is the right instinct, but the PR that establishes the shared predicate leaves the existing copies in place. The same expression is already inlined at:

- `src/components/SkillCard.tsx:41` — `const isOfficial = badges.includes("Official") || owner?.official === true;`
- `src/components/SkillListItem.tsx:33` — character-for-character identical

(`src/components/Header.tsx:1103` carries a third variant of the owner half.)

So after this lands, three call sites compute "is this Official" independently — which is the drift the helper exists to prevent, and it's now load-bearing rather than cosmetic: the feed's membership rule and the card's badge rule have to agree, or a skill appears in the Official tab with no Official badge on it. Since the diff already touches the feed side, folding `SkillCard` and `SkillListItem` onto `isOfficialSkillListing` in the same change would close that gap.

Second, smaller thing: `SkillOwnerLike = object | null | undefined` plus the `"official" in owner` narrowing gives up type safety that's already available. `src/lib/publicUser.ts:20` defines the owner shape with `official?: boolean`, and that's what `buildPublicSkillEntryFromDigest` actually returns, so the parameter can just be typed:

```suggestion
export function isOfficialSkillListing(
  skill: SkillLike,
  owner?: { official?: boolean } | null,
) {
  return isSkillOfficial(skill) || owner?.official === true;
}
```

src/routes/skills/-useSkillsBrowseModel.ts
```diff
@@ -244,6 +255,26 @@ export function useSkillsBrowseModel({
...
+            }
+
+            setListResults((prev) =>
+              cursor ? [...prev, ...legacyOfficialResults] : legacyOfficialResults,
+            );
+            setListCursor(nextCursor);
+            setListAutoLoadPaused(legacyOfficialResults.length === 0 && Boolean(nextCursor));
+            setListStatus(nextCursor ? "idle" : "done");
+            return;
+          }
```
Yigtwxx, line 276: Two things about this block, one of which is the exact scenario the PR exists to handle.

**A cursor can outlive the `apiVersion` that produced it.** When the legacy budget is exhausted, `setListCursor(nextCursor)` stores an untagged cursor minted by the legacy scan. The next "load more" re-enters this fetch and re-calls `fetchCatalogDiscoveryCapabilities()` at `:208` — and during exactly the staggered deploy this PR is written for, it can now hit a v2 backend, flipping `useBackendOfficialFilter` to `true`. It then sends `officialOnly: true` **with that stale untagged cursor**. Post-#3314, `decodeOfficialSkillCursor` classifies an untagged cursor as `projection: "legacy"` (`convex/skills.ts:5439-5442`) and falls through to the generic scan, so the cursor is decoded against a different index than the one that produced it. Capturing the capabilities once per browse session, or storing the `apiVersion` alongside each cursor and resetting to the first page when it changes, would close it.

**Budget exhaustion reproduces the original bug, just later.** `setListAutoLoadPaused(legacyOfficialResults.length === 0 && Boolean(nextCursor))` with `setListStatus("idle")` means that on a legacy backend whose first 200 newest skills contain no official ones, the user lands on the same "No skills found" empty state this PR is fixing — after ten serial round-trips. That branch has no coverage: the new tests in `src/__tests__/skills-index.claw591.test.tsx` and `src/lib/homeListingData.claw591.test.ts` both use four pages with a hit on the last, comfortably inside the budget of ten. A test with 10+ pages and no official skill would pin whatever affordance you want there — "still searching" reads very differently from "nothing exists".

Worth saying that the bound itself is right: a fixed `legacyOfficialMaxRequestsPerFetch` rather than an open `while (true)` means a legacy backend with zero official skills can't spin.

src/lib/homeListingData.ts
```diff
@@ -138,7 +140,10 @@ export async function fetchHomeSkillListing(
   // highlightedOnly is a dedicated backend path ordered by skillBadges.by_kind_at;
   // the nominal sort below is ignored for Featured and never chooses its candidate set.
   const capabilities =
-    tab === "new" ? await fetchCatalogDiscoveryCapabilities() : { apiVersion: 1 as const };
+    tab === "new" || tab === "official"
+      ? await fetchCatalogDiscoveryCapabilities()
+      : { apiVersion: 2 as const };
+  const useBackendOfficialFilter = tab === "official" && capabilities.apiVersion >= 2;
```
Yigtwxx, line 146: This puts an extra serial round-trip on the Official tab's critical path: `fetchCatalogDiscoveryCapabilities()` is awaited before the first listing query even starts, and `src/lib/catalogDiscoveryCapabilities.ts` has no module-level memoization, so every activation of the tab pays it again. The same pattern lands at `-useSkillsBrowseModel.ts:208`.

Since the capabilities response is per-deployment rather than per-request, caching the promise at module scope (with whatever invalidation you think a staggered deploy needs) would make this free after the first call, and it would also give you the single per-session `apiVersion` that the cursor issue in `-useSkillsBrowseModel.ts` wants.

Unrelated but worth calling out as correct: sending `...(useBackendOfficialFilter ? { officialOnly: true } : {})` rather than `officialOnly: undefined` is the part that actually makes this safe against a legacy backend, since Convex validators reject unknown args outright — and pinning it with `expect(args).not.toHaveProperty("officialOnly")` plus a mock that throws `"Object contains extra field \`officialOnly\`"` is a genuinely good way to test it. Same for the `apiVersion` widening being an explicit `2 → 1 → 0` ladder instead of a numeric cast, with a case pinning that a v1 backend still reports `1`.

Yigtwxx (commented):
Read this against `main` at `a9d04bb0`. The client-side half is well built — the bounded scan is genuinely bounded, omitting `officialOnly` rather than sending `undefined` is the detail that actually keeps a legacy backend from erroring, and the `apiVersion` ladder degrades in a defined way for all three versions instead of collapsing unknown values into the newest.

The main thing I'd flag before more review effort goes in: #3314 merged on 2026-07-30, three days after this PR was opened, and added a curated-projection early return to `listPublicPageV4` that a first-page Official request now always takes. After a rebase the new backend predicate sits in a loop that request no longer reaches, while `rolloutCapabilities` still advertises `apiVersion: 2` — details inline on `isOfficialSkillBrowseDigest`. That may also mean the empty-feed symptom is already fixed on `main` by other means, so a fresh reproduction would be worth having before deciding what's left here.

Four other inline notes: a duplicate `officialPublishers` read, three un-consolidated copies of the Official predicate, a cursor that can survive an `apiVersion` flip mid-session, and an uncached capabilities round-trip on the tab's critical path.

src/lib/badges.ts
Yigtwxx: Correction to my own suggestion above: applying it as-is leaves `type SkillOwnerLike = object | null | undefined;` at `src/lib/badges.ts:8` with no remaining references, since this function was its only consumer. That line should come out in the same edit.

PR #3363, opened 2026-07-31
fix: Japanese searches skip the category and summary result tiers

Yigtwxx:
## What Problem This Solves

Fixes an issue where searching in Japanese returns fewer skills than it should. A query
whose katakana contains the prolonged sound mark `ー` — which covers most of the loanword
vocabulary this registry is full of, `データベース`, `サーバー`, `ユーザーインターフェース`,
`コンピューター` — never reaches the category, topic and summary match tiers, and matches
only on name and slug.

The affected surfaces are skill search on `/search`, the skills and plugins catalog
search, and the search digest that backs the catalog index.

## Why This Change Was Made

`tokenize()` delegates word segmentation to `Intl.Segmenter`, which handles Japanese
correctly. But before the segmenter runs, the text is split on a hand-written character
class, and that class omits `ー` (U+30FC) and `々` (U+3005). Both are therefore treated as
separators, so the word is already in pieces by the time the segmenter sees it:

| input | `tokenize()` on `main` | with this change |
|---|---|---|
| `データベース` | `["デ", "タベ", "ス"]` | `["データベース"]` |
| `データベース管理` | `["デ", "タベ", "ス", "管理"]` | `["データベース", "管理"]` |
| `ユーザーインターフェース` | `["ユ", "ザ", "インタ", "フェ", "ス"]` | `["ユーザー", "インターフェース"]` |
| `コンピューター` | `["コンピュ", "タ"]` | `["コンピューター"]` |
| `人々` | `["人"]` — the `々` is dropped | `["人々"]` |

Two things follow from that.

Exploratory search requires *every* query token to clear a three-character floor
(`EXPLORATORY_SEARCH_MIN_TOKEN_LENGTH`, defined identically in `convex/search.ts`,
`convex/skills.ts` and `convex/packages.ts`, enforced by
`matchesExploratoryTokenPrefixes`). `["デ", "タベ", "ス"]` cannot clear it, so rank tiers 2
and 3 — category/topic and summary — are unreachable for these queries.
`["データベース"]` clears it.

And `getFirstSearchToken` in `convex/lib/skillSearchDigest.ts` is `tokenize(value)[0]`.
Its result is stored as `normalizedDisplayNameFirstToken` and `normalizedSlugFirstToken`,
which are indexed columns used as range-scan bounds in `convex/search.ts`. A skill named
`データベース管理` indexes under the single character `デ` instead of `データベース`, so the
bound stops being selective.

The same file already disagrees with itself about this. `detectCJKLanguage`, 53 lines
below, counts katakana with the full Unicode block `/[゠-ヿ]/`, which does match
`ー` — that is how the text is correctly routed to the Japanese segmenter in the first
place. The pre-split then excludes the character the language detector just counted. This
change makes the pre-split agree with it.

I deliberately added only `ー` and `々` rather than widening to the full katakana
block. Widening produces identical `tokenize()` output — `Intl.Segmenter` already reports
`・` (U+30FB) as not word-like — but it would also pull `・`, `゠` and `ヿ` into the
`segmentCJKByChar` fallback as standalone tokens. The narrow change avoids that.

Non-goals:

- Replacing the hand-written ranges with Unicode property escapes. That is a larger
  change to a hot path and `\p{Script=Katakana}` does not match `ー` either, so it would
  not fix this on its own.
- Re-indexing existing rows. The digest is recomputed on write; this only changes what
  future writes and queries produce.

## User Impact

Japanese-language searches now reach the same result tiers as English and Chinese ones,
so a query like `データベース` can match a skill on its category, topics or summary rather
than only on its name. Japanese skills also index under their actual first word instead
of a single character.

## Evidence

Base commit: `a9d04bb0`. The fix is not on `main` — both character classes there still
end `ァ-ヺ가-힯`.

### Focused tests

`convex/lib/searchText.test.ts` already existed, but its only Japanese assertion was
`expect(tokens.length).toBeGreaterThan(0)`, while the Chinese cases directly above it
assert real segmentation. This raises the Japanese cases to that standard and adds the
regressions. On the parent commit:

```
 FAIL  convex/lib/searchText.test.ts > searchText > CJK tokenization > keeps katakana words that contain a prolonged sound mark intact
AssertionError: expected [ 'デ', 'タベ', 'ス' ] to deeply equal [ 'データベース' ]
 FAIL  convex/lib/searchText.test.ts > searchText > CJK tokenization > keeps the iteration mark attached to the character it repeats
AssertionError: expected [ '人' ] to deeply equal [ '人々' ]
 FAIL  convex/lib/searchText.test.ts > searchText > CJK tokenization > lets katakana queries reach the exploratory match tiers
AssertionError: expected false to be true

 Test Files  1 failed (1)
      Tests  3 failed | 15 passed (18)
```

With the change:

```
 Test Files  1 passed (1)
      Tests  18 passed (18)
```

The 15 assertions that already passed still pass, including every Chinese, Korean and
ASCII case — `tokenize("中文搜索")`, `tokenize("안녕하세요")`, `tokenize("React 组件开发")`
and `tokenize("Minimax Usage /minimax-usage")` all produce identical output before and
after.

### The two consequences, measured

```
matchesExploratoryTokenPrefixes(tokenize("データベース"), ["データベース管理ツール"], 3)
  on main          : false
  with this change : true

getFirstSearchToken("データベース管理")     // = tokenize(value)[0]
  on main          : "デ"
  with this change : "データベース"
```

The intra-file inconsistency, checked directly:

```
/[゠-ヿ]/.test("ー")   // detectCJKLanguage  -> true
/[ァ-ヺ]/.test("ー")   // CJK_RE on main     -> false
```

### Deployment consequence: rows written before this tokenizer

`skillSearchDigest.normalizedSlugFirstToken` and `normalizedDisplayNameFirstToken` are
produced by this tokenizer (`getFirstSearchToken` is `tokenize(value)[0]`), and
`convex/search.ts:1327-1364` uses them as range-index bounds. Those rows are recomputed
only when their skill is written, so after deploying this change an untouched row keeps
the old one-character token while the search computes the new longer one. The row stays
on disk and drops out of that first-token range lookup.

Measured on the samples this branch is about - two rows that do not move are kept in as
the control:

```
DRIFT  データベース管理        main="デ"        branch="データベース"
DRIFT  ユーザーインターフェース  main="ユ"        branch="ユーザー"
DRIFT  人々の記録             main="人"        branch="人々"
DRIFT  時々のレポート          main="時"        branch="時々"
DRIFT  コードレビュー          main="コ"        branch="コード"
DRIFT  サーバー監視ツール       main="サ"        branch="サーバー"
same   こんにちは世界          main="こんにちは"  branch="こんにちは"
same   中文文档管理            main="中文"       branch="中文"
drifted 6 / 8
```

`maintenance.ts` gains `backfillSkillSearchDigestFirstTokens`, a cursor-paginated
internal mutation that self-schedules until done, plus the admin-gated action that starts
it. It is written against the shape the file already uses for
`backfillSkillSearchDigestModerationVerdicts` directly above it: same args, same
`clampInt(batchSize, 10, 200)`, same `dryRun`, same return record.

### Batch spacing between backfill pages

`skillSearchDigest` is read by live catalog-search subscriptions, and
`.agents/skills/clawhub-convex/SKILL.md:102` asks for a delay between backfill batches
that write reactively subscribed tables. The mutation takes an optional `delayMs`,
clamped the way the batch size already is, and hands it to the `runAfter` that schedules
the next page:

```ts
const delayMs = clampInt(args.delayMs ?? 500, 0, 60_000);
```

That follows `repairLegacyPublisherOwnershipForUserHandler`
(`convex/maintenance.ts:3272`), which is the one backfill in this file that already
spaces its own batches. Two assertions cover it - the 500 ms default, and an explicit
2000 ms plus the clamp that turns 600000 ms into 60000 ms. Restoring `runAfter(0, ...)`
turns both of them red:

```
control (unmutated)      : 37 passed (37)
mutant: runAfter(0, ...) : 2 failed - repairs digest rows whose stored first tokens
                           predate the tokenizer; spaces the next batch by the
                           requested delay and clamps it
restored                 : 37 passed (37)
```

### Real Convex runtime: a pre-existing row, before and after

A browser capture cannot show this one. Search runs inside Convex, so a locally changed
tokenizer is not what a deployed frontend talks to. The equivalent real setup is
`convex-test`, which runs the actual Convex query engine and the real schema against an
in-memory database - the same harness `convex/catalogFeed.runtime.test.ts` and
`convex/publishAttempts.runtime.test.ts` already use.

`convex/skillSearchDigestFirstTokens.runtime.test.ts` inserts a real `skills` row and a
real `skillSearchDigest` row holding the pre-change token `デ`, then queries the index the
search actually queries (`by_active_normalized_display_name_first_token`) with the bounds
the search computes:

```
stored row  : normalizedDisplayNameFirstToken = "デ"
query bounds: gte "データベース"  lt "データベーソ"

recall before backfill : 0 rows
backfill               : { scanned: 1, patched: 1, missingSkills: 0 }
recall after backfill  : 1 row
```

The second run of the backfill over the same data reports `patched: 0`, so it is
idempotent and safe to re-run.

### Live local Convex: the admin backfill path end to end

The runtime test above proves the index behavior. This is the same migration driven
through the CLI against a real Convex backend - an anonymous local deployment created by
`bunx convex dev`, cloud port 3210, site port 3211.

This transcript was recorded at `8d376a48`, before the confirmation guard below existed.
The two dry runs at steps 4 and 7 behave identically today. The applying call at step 5
is now a preview as written and needs
`{"batchSize":10,"delayMs":15000,"dryRun":false,"confirm":"backfill-skill-search-digest-first-tokens"}`
to do what its output shows. What a page does once it starts - the paging, the 15 s
spacing, the idempotence - is what that commit leaves untouched.

`fixtures/public-corpus/corpus.jsonl` holds no record whose slug or display name carries
`ー` (U+30FC) or `々` (U+3005) - zero of its 1250 rows - so the 13 seeded skills are
constructed: twelve Japanese display names carrying those marks, plus one Latin control.
They are seeded while `a9d04bb0` is the deployed code, so their digest rows are written by
the old tokenizer, and this branch is deployed over them afterwards. Cursor strings are
cut from the output below; nothing else is edited.

```
## 4. dry run: how many stored rows the new tokenizer disagrees with
$ bunx convex run maintenance:backfillSkillSearchDigestFirstTokensInternal {"dryRun":true}
{
  "dryRun": true,
  "isDone": true,
  "missingSkills": 0,
  "patched": 12,
  "scanned": 13
}
## 5. one page, spaced 15s before the next one is scheduled
[11:02:59] starting page 1
$ bunx convex run maintenance:backfillSkillSearchDigestFirstTokensInternal {"batchSize":10,"delayMs":15000}
{
  "dryRun": false,
  "isDone": false,
  "missingSkills": 0,
  "patched": 10,
  "scanned": 10
}
## 6. poll until the scheduled continuation lands
[11:03:02] poll 01 -> "patched":2
[11:03:03] poll 02 -> "patched":2
[11:03:05] poll 03 -> "patched":2
[11:03:07] poll 04 -> "patched":2
[11:03:08] poll 05 -> "patched":2
[11:03:10] poll 06 -> "patched":2
[11:03:11] poll 07 -> "patched":2
[11:03:13] poll 08 -> "patched":2
[11:03:14] poll 09 -> "patched":2
[11:03:16] poll 10 -> "patched":2
[11:03:18] poll 11 -> "patched":0
## 7. idempotence: a further pass patches nothing
$ bunx convex run maintenance:backfillSkillSearchDigestFirstTokensInternal {"dryRun":true}
{
  "dryRun": true,
  "isDone": true,
  "missingSkills": 0,
  "patched": 0,
  "scanned": 13
}
```

Twelve of the thirteen stored rows disagree with the new tokenizer and the Latin control
does not. The first page patches ten and returns `isDone: false`; the remaining two stay
stale for the whole 15 s the scheduled continuation waits, then land between `11:03:16`
and `11:03:18`. A further dry run patches nothing, so a re-run is safe.

What this run does not show is a user-visible recall change. `convex run
search:searchSkills` for `データベース`, `サーバー`, `人々` and `Deploy`, before and after
deploying this branch over the stale rows, returned the same one row each time. The
first-token range lookup is one of several candidate sources inside `nativeSkillSearch`,
and full-text search on `displayName` still matched on a catalog this small. The recall
this backfill restores is specific to that range lookup, which is what the runtime test
isolates.

### The mirrored skills.sh catalog carries the same keys

`skillsShMirrorDigests` persists its own `normalizedSlugFirstToken` and
`normalizedDisplayNameFirstToken`, and `convex/search.ts:1073-1099` range-scans both when
it collects external candidates. They come out of the same tokenizer this PR widens -
`convex/skillsShMirror.ts` had its own `firstSearchToken`, `tokenize(value)[0] ??
normalizedSearchText(value)` - so mirrored rows drift exactly the way native digest rows
do, and the native backfill above cannot reach them because it pages `skillSearchDigest`.

`backfillSkillsShMirrorDigestFirstTokens` is the mirror-side counterpart: same args, same
`clampInt(batchSize, 10, 200)`, same `delayMs`, same `dryRun`, same admin-gated action.
The two copies of the first-token rule are now one - `getMirrorFirstSearchToken` in
`convex/lib/skillSearchDigest.ts`, next to the `getFirstSearchToken` the native digest
already used - so they cannot drift apart again.

Both halves are covered, and each is load-bearing:

```
control (unmutated)                              : 39 passed (39)
mutant A: stop recomputing the display-name token: 2 failed
mutant B: runAfter(0, ...) instead of delayMs     : 1 failed
restored                                         : 39 passed (39)
```

On the live local deployment the mutation reaches the real table - seeding the one
mirrored row `devSeed:seedCanonicalSearchFixture` creates and running the backfill over it
reports `{ scanned: 1, patched: 0 }`, correct for a Latin display name. The drift case
itself is proved in `convex/maintenance.test.ts` rather than live: nothing in the local
dev-seed path produces a mirrored row with a Japanese display name, and the importer
protocol that would create one needs a run, a stored source page and a batch lease.

### The no-Segmenter fallback

Admitting the two marks into `CJK_RE` also reaches `segmentCJKByChar`, the path taken when
`Intl.Segmenter` is missing. It emits one token per character, so the marks came out
standing alone - one-character tokens that exploratory matching discards at its
three-character floor. Measured before the fix:

```
segmentCJKByChar("データベース") -> ["デ","ー","タ","ベ","ー","ス"]
segmentCJKByChar("人々")       -> ["人","々"]
```

They are now attached to the character they extend (`["デー","タ","ベー","ス"]`,
`["人々"]`), which is the same rule the widened class states. Removing that branch turns
exactly one test red.

### Preview by default, confirmation to apply

Both backfills first landed here with `dryRun` defaulting to `false`, and their public
actions forward omitted arguments to the mutation unchanged. A bare
`npx convex run maintenance:backfillSkillSearchDigestFirstTokens` therefore patched
`skillSearchDigest` and scheduled every remaining page: a table-wide write against a
reactively subscribed table, started by a command that reads like an inspection.

They now follow the contract `resyncPluginCatalogMetadataDigestsBatchInternal` already
uses in this file:

```ts
const dryRun = args.dryRun !== false;
if (!dryRun && args.confirm !== SKILL_SEARCH_DIGEST_FIRST_TOKEN_BACKFILL_CONFIRM) {
  throw new ConvexError(
    `Pass confirm="${SKILL_SEARCH_DIGEST_FIRST_TOKEN_BACKFILL_CONFIRM}" to apply.`,
  );
}
```

Three details beyond copying that shape:

- The scheduled continuation carries `confirm` with it. Without that, page two throws on
  the guard page one passed, and a confirmed run stalls after its first batch.
- The native and mirror paths take different tokens
  (`backfill-skill-search-digest-first-tokens` and
  `backfill-skills-sh-mirror-digest-first-tokens`), so neither one starts the other
  table's rewrite.
- A preview returns `confirmRequired`, so the token comes out of the dry run an operator
  already has to do rather than out of the source.

Four new tests cover it, two per path: omitted arguments preview without writing or
scheduling, and an apply with no token or the other path's token throws before the first
`paginate`. The three existing tests that exercise a real page now pass the token and
assert it in the `runAfter` payload. The runtime test drives the guard through a Convex
backend as well - the unconfirmed call reports `patched: 1` and leaves the row holding
its stale `デ`, and only the confirmed call rewrites it.

### Live local Convex: the guard itself

Same setup as the transcript above, rebuilt: an anonymous local deployment, thirteen
constructed rows seeded while `a9d04bb0` is deployed so their digest rows carry the old
tokenizer's output, then this branch deployed over them. Eleven of the thirteen drift.
Cursor strings are cut; nothing else is edited.

```
## 1. omitted arguments
$ bunx convex run maintenance:backfillSkillSearchDigestFirstTokensInternal {}
{
  "confirmRequired": "backfill-skill-search-digest-first-tokens",
  "dryRun": true,
  "isDone": true,
  "missingSkills": 0,
  "patched": 11,
  "scanned": 13
}
## 2. apply with no token
$ bunx convex run maintenance:backfillSkillSearchDigestFirstTokensInternal {"dryRun":false}
Uncaught ConvexError: Pass confirm="backfill-skill-search-digest-first-tokens" to apply.
    at handler (../convex/maintenance.ts:2713:8)
exit=1
## 3. apply with the mirror path's token
$ ... {"dryRun":false,"confirm":"backfill-skills-sh-mirror-digest-first-tokens"}
Uncaught ConvexError: Pass confirm="backfill-skill-search-digest-first-tokens" to apply.
exit=1
## 4. neither rejected call wrote anything
$ bunx convex run maintenance:backfillSkillSearchDigestFirstTokensInternal {}
  "dryRun": true, "patched": 11, "scanned": 13
## 5. confirmed apply
$ ... {"dryRun":false,"confirm":"backfill-skill-search-digest-first-tokens"}
  "dryRun": false, "isDone": true, "missingSkills": 0, "patched": 11, "scanned": 13
## 6. idempotence
$ bunx convex run maintenance:backfillSkillSearchDigestFirstTokensInternal {}
  "dryRun": true, "patched": 0, "scanned": 13
```

The mirror action rejects the same way, including the native path's token, on a mirror
table this fixture leaves empty (`scanned: 0`), so that pair shows the guard rather than
a migration:

```
$ bunx convex run maintenance:backfillSkillsShMirrorDigestFirstTokensInternal {"dryRun":false}
Uncaught ConvexError: Pass confirm="backfill-skills-sh-mirror-digest-first-tokens" to apply.
    at handler (../convex/maintenance.ts:2815:8)
exit=1
$ ... {"dryRun":false,"confirm":"backfill-skill-search-digest-first-tokens"}
Uncaught ConvexError: Pass confirm="backfill-skills-sh-mirror-digest-first-tokens" to apply.
exit=1
$ ... {"dryRun":false,"confirm":"backfill-skills-sh-mirror-digest-first-tokens"}
  "dryRun": false, "isDone": true, "patched": 0, "scanned": 0
```

The token has to survive the scheduler or a confirmed run applies its first page and then
throws on the guard that page just passed. Re-seeded, then one confirmed page of ten with
the continuation 15 s out, polled by dry run:

```
[22:07:35] page 1
$ ... {"batchSize":10,"delayMs":15000,"dryRun":false,"confirm":"backfill-skill-search-digest-first-tokens"}
  "dryRun": false, "isDone": false, "missingSkills": 0, "patched": 10, "scanned": 10
[22:07:39] poll 1 -> "patched":1
[22:07:41] poll 2 -> "patched":1
[22:07:43] poll 3 -> "patched":1
[22:07:45] poll 4 -> "patched":1
[22:07:47] poll 5 -> "patched":1
[22:07:49] poll 6 -> "patched":1
[22:07:51] poll 7 -> "patched":1
[22:07:53] poll 8 -> "patched":1
[22:07:55] poll 9 -> "patched":0
[22:07:57] poll 10 -> "patched":0
```

The eleventh row stays stale for the whole delay and then lands. A continuation that lost
the token would have thrown instead, and that row would still read `"patched":1` at poll
10.

### Rollout

Both backfills are admin-gated and are meant to run once after this deploys, native
first, then mirror. Both preview by default, so this form counts rows without writing:

```
npx convex run maintenance:backfillSkillSearchDigestFirstTokens --prod
npx convex run maintenance:backfillSkillsShMirrorDigestFirstTokens --prod
```

Applying takes the token that preview returns as `confirmRequired`:

```
npx convex run maintenance:backfillSkillSearchDigestFirstTokens \
  '{"dryRun": false, "confirm": "backfill-skill-search-digest-first-tokens"}' --prod
npx convex run maintenance:backfillSkillsShMirrorDigestFirstTokens \
  '{"dryRun": false, "confirm": "backfill-skills-sh-mirror-digest-first-tokens"}' --prod
```

Both accept `{"delayMs": N}` if the default 500 ms between pages is too tight. Both are
idempotent, so a re-run after a partial pass is safe.

### CI

- `bun run lint`, `bun run deadcode:ci`, `bun run ci:types-build` — all pass.
- `bunx tsc -p packages/schema/tsconfig.json --noEmit` and
  `bunx tsc -p packages/clawhub/tsconfig.json --noEmit` — both clean.
- `bun run format:check` fails on `CLAUDE.md` and
  `.agents/skills/autoreview/CLAUDE.md`. That is pre-existing: the same two files, and
  only those two, fail on `main` at `a9d04bb0`. The two files in this PR are formatted
  with `oxfmt`.
- `VITE_CONVEX_URL=https://example.invalid bun run ci:unit` on this branch:
  `14 failed | 431 passed | 1 skipped (446)` files, `24 failed | 5790 passed | 9 skipped`
  tests. On `main` at `a9d04bb0`: `14 failed | 430 passed | 1 skipped (445)` files and
  `24 failed | 5773 passed | 9 skipped` tests. The sorted list of failing files is
  identical on both sides — they are the suites that shell out to `bun`, plus
  `convex/lib/githubAccount.test.ts` and `src/routes/-management.test.tsx`. The failure
  count is the same 24 on both sides. This branch adds one file and seventeen passing
  tests and introduces no new failure.

The Vercel check on this PR stays pending OpenClaw Foundation authorization, as it does
on every fork PR.

Written with AI assistance. I ran the focused tests, the drift measurement and the
Convex runtime proof above myself,
diffed the full unit suite against `main`, and can maintain this code.

### DIFF
--- a/convex/lib/searchText.test.ts
+++ b/convex/lib/searchText.test.ts
@@ -94,8 +94,31 @@ describe("searchText", () => {
     });
 
     it("handles Japanese text", () => {
-      const tokens = tokenize("こんにちは世界");
-      expect(tokens.length).toBeGreaterThan(0);
+      expect(tokenize("こんにちは世界")).toEqual(["こんにちは", "世界"]);
+    });
+
+    it("keeps katakana words that contain a prolonged sound mark intact", () => {
+      expect(tokenize("データベース")).toEqual(["データベース"]);
+      expect(tokenize("データベース管理")).toEqual(["データベース", "管理"]);
+      expect(tokenize("ユーザーインターフェース")).toEqual(["ユーザー", "インターフェース"]);
+    });
+
+    it("keeps the iteration mark attached to the character it repeats", () => {
+      expect(tokenize("人々")).toEqual(["人々"]);
+      expect(tokenize("時々")).toEqual(["時々"]);
+    });
+
+    it("matches Japanese query tokens against Japanese skill names", () => {
+      const queryTokens = tokenize("データベース");
+      expect(matchesExactTokens(queryTokens, ["データベース管理ツール"])).toBe(true);
+    });
+
+    it("lets katakana queries reach the exploratory match tiers", () => {
+      // Exploratory tiers require every query token to clear a three-character floor.
+      const queryTokens = tokenize("データベース");
+      expect(matchesExploratoryTokenPrefixes(queryTokens, ["データベース管理ツール"], 3)).toBe(
+        true,
+      );
     });
 
     it("handles Korean text", () => {
--- a/convex/lib/searchText.test.ts
+++ b/convex/lib/searchText.test.ts
@@ -108,6 +108,14 @@ describe("searchText", () => {
       expect(tokenize("時々")).toEqual(["時々"]);
     });
 
+    it("keeps the marks attached when Intl.Segmenter is unavailable", () => {
+      // segmentCJKByChar is the no-Segmenter fallback. Emitting ー or 々 on their own
+      // would leave one-character tokens that exploratory matching discards.
+      expect(__test.segmentCJKByChar("データベース")).toEqual(["デー", "タ", "ベー", "ス"]);
+      expect(__test.segmentCJKByChar("人々")).toEqual(["人々"]);
+      expect(__test.segmentCJKByChar("時々の記録")).toEqual(["時々", "の", "記", "録"]);
+    });
+
     it("matches Japanese query tokens against Japanese skill names", () => {
       const queryTokens = tokenize("データベース");
       expect(matchesExactTokens(queryTokens, ["データベース管理ツール"])).toBe(true);
--- a/convex/lib/searchText.ts
+++ b/convex/lib/searchText.ts
@@ -1,4 +1,6 @@
-const CJK_RE = /[\u4e00-\u9fff\u3400-\u4dbf\u3041-\u3096\u30a1-\u30fa\uac00-\ud7af]/;
+// U+30FC (ー) and U+3005 (々) extend the word they follow, so the pre-split in tokenize()
+// must keep them with that word instead of treating them as separators.
+const CJK_RE = /[\u4e00-\u9fff\u3400-\u4dbf\u3041-\u3096\u30a1-\u30fa\u30fc\u3005\uac00-\ud7af]/;
 
 const hasSegmenter = typeof Intl !== "undefined" && "Segmenter" in Intl;
 
@@ -116,7 +118,7 @@ export function tokenize(value: string): string[] {
   const tokens: string[] = [];
 
   const parts = normalized.split(
-    /([^\u4e00-\u9fff\u3400-\u4dbf\u3041-\u3096\u30a1-\u30fa\uac00-\ud7af]+)/g,
+    /([^\u4e00-\u9fff\u3400-\u4dbf\u3041-\u3096\u30a1-\u30fa\u30fc\u3005\uac00-\ud7af]+)/g,
   );
 
   for (const part of parts) {
--- a/convex/lib/searchText.ts
+++ b/convex/lib/searchText.ts
@@ -2,6 +2,10 @@
 // must keep them with that word instead of treating them as separators.
 const CJK_RE = /[\u4e00-\u9fff\u3400-\u4dbf\u3041-\u3096\u30a1-\u30fa\u30fc\u3005\uac00-\ud7af]/;
 
+// The same two marks the class above admits: they extend the preceding character rather
+// than standing on their own, so the per-character fallback must not emit them alone.
+const CJK_EXTENDER_RE = /[\u30fc\u3005]/;
+
 const hasSegmenter = typeof Intl !== "undefined" && "Segmenter" in Intl;
 
 let zhSegmenter: Intl.Segmenter | null = null;
@@ -36,9 +40,12 @@ function getKoSegmenter(): Intl.Segmenter {
 function segmentCJKByChar(text: string): string[] {
   const tokens: string[] = [];
   for (const ch of text) {
-    if (CJK_RE.test(ch)) {
-      tokens.push(ch);
+    if (!CJK_RE.test(ch)) continue;
+    if (CJK_EXTENDER_RE.test(ch) && tokens.length > 0) {
+      tokens[tokens.length - 1] += ch;
+      continue;
     }
+    tokens.push(ch);
   }
   return tokens;
 }
--- a/convex/lib/skillSearchDigest.ts
+++ b/convex/lib/skillSearchDigest.ts
@@ -146,6 +146,12 @@ export function getFirstSearchToken(value: string) {
   return tokenize(value)[0];
 }
 
+// The skills.sh mirror persists the same tokenizer-derived first tokens, but its columns
+// are required, so it falls back to the normalized text when the tokenizer yields nothing.
+export function getMirrorFirstSearchToken(value: string) {
+  return getFirstSearchToken(value) ?? normalizeSkillSearchText(value);
+}
+
 /**
  * Map a digest row to the HydratableSkill shape expected by toPublicSkill /
  * isPublicSkillDoc / isSkillSuspicious.  Fully type-checked: if
--- a/convex/maintenance.test.ts
+++ b/convex/maintenance.test.ts
@@ -36,6 +36,9 @@ vi.mock("./_generated/api", () => ({
       backfillSkillSearchDigestFirstTokensInternal: Symbol(
         "backfillSkillSearchDigestFirstTokensInternal",
       ),
+      backfillSkillsShMirrorDigestFirstTokensInternal: Symbol(
+        "backfillSkillsShMirrorDigestFirstTokensInternal",
+      ),
       getEmptySkillCleanupPageInternal: Symbol("getEmptySkillCleanupPageInternal"),
       applyEmptySkillCleanupInternal: Symbol("applyEmptySkillCleanupInternal"),
       nominateUserForEmptySkillSpamInternal: Symbol("nominateUserForEmptySkillSpamInternal"),
@@ -79,6 +82,7 @@ const {
   backfillLatestVersionSummaryInternal,
   backfillSkillSearchDigestModerationVerdictsInternal,
   backfillSkillSearchDigestFirstTokensInternal,
+  backfillSkillsShMirrorDigestFirstTokensInternal,
   backfillPublisherStatsInternalHandler,
   backfillSkillFingerprintsInternalHandler,
   backfillSkillSummariesInternalHandler,
@@ -2129,7 +2133,95 @@ describe("backfillSkillSearchDigestFirstTokensInternal", () => {
       { cursor: "next-page", batchSize: undefined, delayMs: 600_000, dryRun: false },
     );
   });
+});
 
+describe("backfillSkillsShMirrorDigestFirstTokensInternal", () => {
+  const mirrorPage = () => [
+    {
+      _id: "skillsShMirrorDigests:stale",
+      slug: "database",
+      displayName: "データベース管理",
+      normalizedSlugFirstToken: "database",
+      normalizedDisplayNameFirstToken: "デ",
+    },
+    {
+      _id: "skillsShMirrorDigests:fresh",
+      slug: "deploy",
+      displayName: "Deploy helper",
+      normalizedSlugFirstToken: "deploy",
+      normalizedDisplayNameFirstToken: "deploy",
+    },
+  ];
+
+  it("repairs mirrored rows whose stored first tokens predate the tokenizer", async () => {
+    const paginate = vi.fn().mockResolvedValue({
+      page: mirrorPage(),
+      continueCursor: "next-page",
+      isDone: false,
+    });
+    const query = vi.fn().mockReturnValue({ paginate });
+    const patch = vi.fn().mockResolvedValue(undefined);
+    const runAfter = vi.fn().mockResolvedValue(undefined);
+
+    const result = await (
+      backfillSkillsShMirrorDigestFirstTokensInternal as unknown as { _handler: Function }
+    )._handler(
+      {
+        db: { query, get: vi.fn(), patch, normalizeId: vi.fn() },
+        scheduler: { runAfter },
+      } as never,
+      { cursor: "start", batchSize: 25 },
+    );
+
+    expect(result).toEqual({
+      scanned: 2,
+      patched: 1,
+      cursor: "next-page",
+      isDone: false,
+      dryRun: false,
+    });
+    expect(query).toHaveBeenCalledWith("skillsShMirrorDigests");
+    expect(paginate).toHaveBeenCalledWith({ cursor: "start", numItems: 25 });
+    expect(patch).toHaveBeenCalledTimes(1);
+    expect(patch).toHaveBeenCalledWith("skillsShMirrorDigests:stale", {
+      normalizedSlugFirstToken: "database",
+      normalizedDisplayNameFirstToken: "データベース",
+    });
+    expect(runAfter).toHaveBeenCalledWith(
+      500,
+      internal.maintenance.backfillSkillsShMirrorDigestFirstTokensInternal,
+      { cursor: "next-page", batchSize: 25, delayMs: undefined, dryRun: false },
+    );
+  });
+
+  it("reports would-be patches without writing or scheduling in dry run mode", async () => {
+    const paginate = vi.fn().mockResolvedValue({
+      page: mirrorPage(),
+      continueCursor: "next-page",
+      isDone: false,
+    });
+    const query = vi.fn().mockReturnValue({ paginate });
+    const patch = vi.fn().mockResolvedValue(undefined);
+    const runAfter = vi.fn().mockResolvedValue(undefined);
+
+    const result = await (
+      backfillSkillsShMirrorDigestFirstTokensInternal as unknown as { _handler: Function }
+    )._handler(
+      {
+        db: { query, get: vi.fn(), patch, normalizeId: vi.fn() },
+        scheduler: { runAfter },
+      } as never,
+      { dryRun: true },
+    );
+
+    expect(result.patched).toBe(1);
+    expect(result.dryRun).toBe(true);
+    expect(patch).not.toHaveBeenCalled();
+    expect(runAfter).not.toHaveBeenCalled();
+  });
+});
+
+describe("backfillSkillSearchDigestFirstTokensInternal dry run", () => {
   it("reports would-be patches without writing or scheduling in dry run mode", async () => {
     const paginate = vi.fn().mockResolvedValue({
       page: [
--- a/convex/maintenance.test.ts
+++ b/convex/maintenance.test.ts
@@ -2080,9 +2080,53 @@ describe("backfillSkillSearchDigestFirstTokensInternal", () => {
       normalizedDisplayNameFirstToken: "データベース",
     });
     expect(runAfter).toHaveBeenCalledWith(
-      0,
+      500,
       internal.maintenance.backfillSkillSearchDigestFirstTokensInternal,
-      { cursor: "next-page", batchSize: 25, dryRun: false },
+      { cursor: "next-page", batchSize: 25, delayMs: undefined, dryRun: false },
+    );
+  });
+
+  it("spaces the next batch by the requested delay and clamps it", async () => {
+    const paginate = vi.fn().mockResolvedValue({
+      page: [
+        {
+          _id: "skillSearchDigest:stale",
+          skillId: "skills:stale",
+          normalizedSlugFirstToken: "database",
+          normalizedDisplayNameFirstToken: "デ",
+        },
+      ],
+      continueCursor: "next-page",
+      isDone: false,
+    });
+    const query = vi.fn().mockReturnValue({ paginate });
+    const get = vi.fn().mockResolvedValue({
+      _id: "skills:stale",
+      slug: "database",
+      displayName: "データベース管理",
+    });
+    const patch = vi.fn().mockResolvedValue(undefined);
+    const runAfter = vi.fn().mockResolvedValue(undefined);
+    const handler = (
+      backfillSkillSearchDigestFirstTokensInternal as unknown as { _handler: Function }
+    )._handler;
+    const ctx = {
+      db: { query, get, patch, normalizeId: vi.fn() },
+      scheduler: { runAfter },
+    } as never;
+
+    await handler(ctx, { delayMs: 2_000 });
+    expect(runAfter).toHaveBeenLastCalledWith(
+      2_000,
+      internal.maintenance.backfillSkillSearchDigestFirstTokensInternal,
+      { cursor: "next-page", batchSize: undefined, delayMs: 2_000, dryRun: false },
+    );
+
+    await handler(ctx, { delayMs: 600_000 });
+    expect(runAfter).toHaveBeenLastCalledWith(
+      60_000,
+      internal.maintenance.backfillSkillSearchDigestFirstTokensInternal,
+      { cursor: "next-page", batchSize: undefined, delayMs: 600_000, dryRun: false },
     );
   });
 
--- a/convex/maintenance.test.ts
+++ b/convex/maintenance.test.ts
@@ -2015,6 +2015,11 @@ describe("maintenance empty skill nominations", () => {
   });
 });
 
+const SKILL_SEARCH_DIGEST_FIRST_TOKEN_BACKFILL_CONFIRM =
+  "backfill-skill-search-digest-first-tokens";
+const SKILLS_SH_MIRROR_DIGEST_FIRST_TOKEN_BACKFILL_CONFIRM =
+  "backfill-skills-sh-mirror-digest-first-tokens";
+
 describe("backfillSkillSearchDigestFirstTokensInternal", () => {
   it("repairs digest rows whose stored first tokens predate the tokenizer", async () => {
     const paginate = vi.fn().mockResolvedValue({
@@ -2065,7 +2070,12 @@ describe("backfillSkillSearchDigestFirstTokensInternal", () => {
         db: { query, get, patch, normalizeId: vi.fn() },
         scheduler: { runAfter },
       } as never,
-      { cursor: "start", batchSize: 25 },
+      {
+        cursor: "start",
+        batchSize: 25,
+        dryRun: false,
+        confirm: SKILL_SEARCH_DIGEST_FIRST_TOKEN_BACKFILL_CONFIRM,
+      },
     );
 
     expect(result).toEqual({
@@ -2075,6 +2085,7 @@ describe("backfillSkillSearchDigestFirstTokensInternal", () => {
       cursor: "next-page",
       isDone: false,
       dryRun: false,
+      confirmRequired: undefined,
     });
     expect(query).toHaveBeenCalledWith("skillSearchDigest");
     expect(paginate).toHaveBeenCalledWith({ cursor: "start", numItems: 25 });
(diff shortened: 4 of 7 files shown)


Patrick-Erichsen (changes requested):
This changes Japanese tokenization so category and summary search tiers can use longer tokens. It changes the persisted prefix-index contract without migrating existing rows, and the supported no-Segmenter fallback still fails the stated behavior.

LOC: +29/-4 (2 files)

Findings: add an idempotent rate-controlled `skillSearchDigest` backfill for existing rows, and make/test the no-`Intl.Segmenter` path retain usable Japanese word runs rather than standalone marks or sub-three-character tokens.

Best-fix verdict: too narrow. Tokenizer correctness and derived-index migration are one deployment contract; changing only query/write logic leaves existing catalog rows inconsistent.

Alternatives considered: lazy rewrite on later skill mutation leaves search broken indefinitely; lowering the exploratory floor would weaken all languages instead of fixing Japanese token formation.

Code read: tokenizer/digest generation, prefix-index query path, fallback branch, and focused tests.

Remaining uncertainty: no real catalog/Convex search proof or rollout plan is supplied.

Yigtwxx:
@Patrick-Erichsen both findings are addressed, the second half of the first one in a follow-up commit.

**Idempotent, rate-controlled backfill for existing rows** - `fa8eb54d` adds `backfillSkillSearchDigestFirstTokens`, a cursor-paginated internal mutation plus the admin-gated action that starts it, written against the shape `backfillSkillSearchDigestModerationVerdicts` directly above it uses. Idempotence is proved rather than asserted: `convex/skillSearchDigestFirstTokens.runtime.test.ts` runs the real Convex query engine against the real schema, inserts a `skillSearchDigest` row holding the pre-change token `デ`, and queries the index search actually queries with the bounds search actually computes - recall goes 0 rows -> backfill `{ scanned: 1, patched: 1, missingSkills: 0 }` -> 1 row, and a second run over the same data reports `patched: 0`.

The rate-control half was missing and is now in `815df69a`. `skillSearchDigest` is read by live catalog-search subscriptions, and `.agents/skills/clawhub-convex/SKILL.md:102` asks for a delay between backfill batches that write reactively subscribed tables, so the mutation takes an optional `delayMs` clamped the way the batch size already is - `clampInt(args.delayMs ?? 500, 0, 60_000)` - and hands it to the `runAfter` that schedules the next page. That follows `repairLegacyPublisherOwnershipForUserHandler` at `convex/maintenance.ts:3272`, the one backfill in this file that already spaces its own batches. Restoring `runAfter(0, ...)` turns both of the covering assertions red.

**No-`Intl.Segmenter` path** - `fa8eb54d` also fixes `segmentCJKByChar`, which is the branch taken when `Intl.Segmenter` is missing. Widening `CJK_RE` alone left the two marks standing alone there, as one-character tokens that exploratory matching discards at its three-character floor:

```
before: segmentCJKByChar("データベース") -> ["デ","ー","タ","ベ","ー","ス"]
        segmentCJKByChar("人々")       -> ["人","々"]
after : ["デー","タ","ベー","ス"]
        ["人々"]
```

Each mark is attached to the character it extends, which is the rule the widened class states. Removing that branch turns exactly one test red.

Verification on `815df69a`: `bunx tsc --noEmit` clean, `bunx oxfmt --check` clean on both changed files, and `VITE_CONVEX_URL=https://example.invalid bun run ci:unit` gives `14 failed | 431 passed | 1 skipped (446)` files and `24 failed | 5784 passed | 9 skipped` tests against `14 failed | 430 passed | 1 skipped (445)` and `24 failed | 5773 passed | 9 skipped` on `main` at `a9d04bb0`. The sorted list of the 14 failing files is identical on both sides.

One thing worth flagging rather than hiding: the delay now differs from the ten other self-continuing backfills in `convex/maintenance.ts`, which all schedule with `runAfter(0, ...)` - including `backfillSkillSearchDigestModerationVerdictsInternal`, which patches the same table. I followed the written guidance rather than the surrounding code, but if you would rather this match its neighbours I can drop it.

Also note that this branch now touches `convex/maintenance.ts`, which CODEOWNERS routes to `@openclaw/openclaw-secops`.

Yigtwxx:
The live-deployment proof is now in the PR body under **Evidence -> Live local Convex: the admin backfill path end to end**.

It is the migration driven through the CLI against a real Convex backend - an anonymous local deployment from `bunx convex dev`, cloud port 3210, site port 3211 - not `convex-test`. Thirteen skills are seeded while `a9d04bb0` is the deployed code, so their digest rows are written by the old tokenizer, and this branch is deployed over them afterwards:

```
$ bunx convex run maintenance:backfillSkillSearchDigestFirstTokensInternal {"dryRun":true}
{ "dryRun": true, "isDone": true, "missingSkills": 0, "patched": 12, "scanned": 13 }

[11:02:59] starting page 1
$ bunx convex run maintenance:backfillSkillSearchDigestFirstTokensInternal {"batchSize":10,"delayMs":15000}
{ "dryRun": false, "isDone": false, "missingSkills": 0, "patched": 10, "scanned": 10 }

[11:03:02] poll 01 -> "patched":2
[11:03:07] poll 04 -> "patched":2
[11:03:14] poll 09 -> "patched":2
[11:03:16] poll 10 -> "patched":2
[11:03:18] poll 11 -> "patched":0

$ bunx convex run maintenance:backfillSkillSearchDigestFirstTokensInternal {"dryRun":true}
{ "dryRun": true, "isDone": true, "missingSkills": 0, "patched": 0, "scanned": 13 }
```

Twelve of thirteen rows disagree with the new tokenizer; the Latin control does not. The first page patches ten and returns `isDone: false`, the remaining two stay stale for the whole 15 s the scheduled continuation waits, and a further dry run patches nothing. That is the throttling working on a real deployment rather than only in a mock.

Two things I would rather state than leave for you to find:

- The seeded rows are constructed. `fixtures/public-corpus/corpus.jsonl` has no record whose slug or display name carries `ー` (U+30FC) or `々` (U+3005) - zero of 1250 - so there was nothing in the real corpus to drift.
- This run does not show a user-visible recall change. `convex run search:searchSkills` for `データベース`, `サーバー`, `人々` and `Deploy` returned the same one row before and after the tokenizer swap, because the first-token range lookup is one of several candidate sources in `nativeSkillSearch` and full-text search on `displayName` still matched on a catalog this small. The recall the backfill restores is specific to that range lookup, which is what `convex/skillSearchDigestFirstTokens.runtime.test.ts` isolates. I have softened the wording in the PR body accordingly - it now says the stale row drops out of the first-token range lookup rather than out of recall generally.

Yigtwxx:
The mirror finding is fixed in `8d376a48`.

It is a real gap and the diagnosis is right: `skillsShMirrorDigests` persists its own `normalizedSlugFirstToken` and `normalizedDisplayNameFirstToken`, `convex/search.ts:1073-1099` range-scans both when collecting external candidates, and `convex/skillsShMirror.ts` derived them through its own copy of the rule - `tokenize(value)[0] ?? normalizedSearchText(value)` - so widening the katakana class strands mirrored rows the same way it stranded native ones. The previous backfill paged `skillSearchDigest` only and could never reach them.

`backfillSkillsShMirrorDigestFirstTokens` is the mirror-side counterpart of the native backfill: same args, same `clampInt(batchSize, 10, 200)`, same `delayMs` spacing, same `dryRun`, same admin-gated action wrapper.

The duplicated rule is also gone. `getMirrorFirstSearchToken` now lives in `convex/lib/skillSearchDigest.ts` beside the `getFirstSearchToken` the native digest already used, and `skillsShMirror.ts` imports it, so the two copies cannot drift apart again.

Both halves of the new mutation are load-bearing:

```
control (unmutated)                               : 39 passed (39)
mutant A: stop recomputing the display-name token : 2 failed
mutant B: runAfter(0, ...) instead of delayMs     : 1 failed
restored                                          : 39 passed (39)
```

Verification on `8d376a48`: `bunx tsc --noEmit` clean, `bun run lint` and `bun run deadcode:ci` clean, `bunx oxfmt --check` clean on all four changed files, `convex/skillsShMirror.test.ts convex/skillsShMirrorVisibility.test.ts convex/search.test.ts convex/lib/searchText.test.ts convex/skillSearchDigestFirstTokens.runtime.test.ts` 156 passed, and `bun run ci:unit` gives `14 failed | 431 passed | 1 skipped (446)` files and `24 failed | 5786 passed | 9 skipped` tests against `14 failed | 430 passed | 1 skipped (445)` and `24 failed | 5773 passed | 9 skipped` on `main` at `a9d04bb0`, with an identical list of the 14 failing files.

One limit worth stating plainly: on the live local deployment I could only prove the mutation reaches the real table. Seeding the single mirrored row `devSeed:seedCanonicalSearchFixture` creates and running the backfill over it reports `{ scanned: 1, patched: 0 }`, which is correct for a Latin display name. The drift case itself is covered by `convex/maintenance.test.ts`, not live, because nothing in the local dev-seed path produces a mirrored row with a Japanese display name and the importer protocol that would create one needs a run, a stored source page and a batch lease. If you would rather see it end to end, say so and I will add a small dev-seed fixture for a mirrored Japanese row.

Yigtwxx:
On the fallback finding - *"Skip leading Japanese extenders in the fallback (P2) ... The Segmenter path drops such non-word segments"* - I do not think that premise holds, so I have not made the change. The measurement, on the current head `8d376a48`:

```
input        Intl.Segmenter (isWordLike)      segmentCJKByChar
ーデータ      ["ーデータ"]                      ["ー","デー","タ"]
々人         ["々","人"]                       ["々","人"]
ー人々の記録   ["ー","人々","の","記録"]           ["ー","人々","の","記","録"]
データベース   ["データベース"]                   ["デー","タ","ベー","ス"]
```

Same output on Node v22.20.0 (V8) and on Bun 1.3.14 (JSC).

Two things follow:

1. **The Segmenter does not drop unattached extenders.** On `ー人々の記録` it emits `ー` as its own segment with `isWordLike === true`, so `segmentCJK` keeps it. The fallback emitting `ー` alone there is parity, not divergence.
2. **On `々人` the two paths are already byte-identical.** Skipping unattached extenders would turn the fallback into `["人"]` and make it disagree with the Segmenter on the one input where they currently agree exactly.

Where the two really differ - `データベース` as `["データベース"]` against `["デー","タ","ベー","ス"]` - is the per-character nature of the fallback, which predates this PR and is what the fallback is for. This PR's change to it was narrower: before, `ー` and `々` were excluded from `CJK_RE` entirely, so `segmentCJKByChar("データベース")` returned `["デ","ー","タ","ベ","ー","ス"]` with the marks stranded as one-character tokens that exploratory matching discards at its three-character floor. They are now attached to the character they extend. A mark in leading position has no character to extend, and dropping it would lose the only signal a slug like `ーデータ` carries in that position.

Happy to be wrong here - if the intent is that the fallback should normalise more aggressively than the Segmenter rather than approximate it, that is a different and larger change than this PR, and I would rather do it separately than bolt it on.

On the remaining rollout item: both backfills are admin-gated and must be run once after deploy, native first, then mirror.

```
npx convex run maintenance:backfillSkillSearchDigestFirstTokens --prod
npx convex run maintenance:backfillSkillsShMirrorDigestFirstTokens --prod
```

Both accept `{"dryRun": true}` for a count-only pass and `{"delayMs": N}` if the default 500 ms between pages is too tight for production. I have added that to the PR body so it is not buried in a comment.

Yigtwxx:
The confirmation finding is fixed in `bdd16378`.

That one was mine to fix rather than argue with. `dryRun` defaulted to `false` on both new
backfills and both public actions forward omitted arguments to the mutation unchanged, so
`npx convex run maintenance:backfillSkillSearchDigestFirstTokens --prod` - a command that
reads like an inspection - patched `skillSearchDigest` and scheduled every remaining page.

Both now use the contract `resyncPluginCatalogMetadataDigestsBatchInternal` already has in
this file: `const dryRun = args.dryRun !== false`, and an apply whose `confirm` does not
match throws before the first `paginate`.

Three things beyond copying that shape:

1. **The scheduled continuation carries `confirm`.** Without it a confirmed run applies
   page one and then throws on the guard page one just passed, which would be a worse
   failure than the one being fixed - a half-migrated index and no obvious signal.
2. **The two paths take separate tokens.** `backfill-skill-search-digest-first-tokens`
   does not start the mirror rewrite, and a test pins that direction specifically.
3. **A preview returns `confirmRequired`.** The token comes out of the dry run an operator
   already has to run, not out of the source.

Coverage: four new tests, two per path - omitted arguments preview without writing or
scheduling, and an apply with no token or the other path's token throws. The three
existing tests that drive a real page now pass the token and assert it in the `runAfter`
payload. The runtime test drives the guard through a Convex backend too: the unconfirmed
call reports `patched: 1` and leaves the row holding its stale `デ`, and only the confirmed
call rewrites it.

```
VITE_CONVEX_URL=https://example.invalid bunx vitest run convex/maintenance.test.ts \
  convex/lib/searchText.test.ts convex/skillSearchDigestFirstTokens.runtime.test.ts
Test Files  3 passed (3)
     Tests  65 passed (65)
```

Reverting only `convex/maintenance.ts` and keeping the tests turns 7 of them red.

One thing I did not do: re-record the live local Convex transcript in the PR body. It was
taken at `8d376a48` and I have labelled it as such there. Its two dry runs behave
identically today; its applying call now needs the token to do what its output shows. The
guard sits in front of the paging, the 15 s spacing and the idempotence that transcript
exists to demonstrate, and none of that changed, so re-seeding a local deployment at the
old revision to reproduce it would not test anything the runtime case does not already
cover against a real backend. Say the word if you would rather have it re-recorded anyway.

The rollout section in the PR body now carries both forms, preview and apply.

Yigtwxx:
I said above that I had not re-recorded the live transcript. I have now, for the guard
specifically - it turned out to be cheap. Full text is in the PR body under **Live local
Convex: the guard itself**; the short version, on an anonymous local deployment with
thirteen rows seeded while `a9d04bb0` was deployed and this branch deployed over them:

```
## 1. omitted arguments
$ bunx convex run maintenance:backfillSkillSearchDigestFirstTokensInternal {}
  "confirmRequired": "backfill-skill-search-digest-first-tokens",
  "dryRun": true, "patched": 11, "scanned": 13
## 2. apply with no token
$ ... {"dryRun":false}
Uncaught ConvexError: Pass confirm="backfill-skill-search-digest-first-tokens" to apply.
    at handler (../convex/maintenance.ts:2713:8)
## 3. apply with the mirror path's token
$ ... {"dryRun":false,"confirm":"backfill-skills-sh-mirror-digest-first-tokens"}
Uncaught ConvexError: Pass confirm="backfill-skill-search-digest-first-tokens" to apply.
## 4. neither rejected call wrote anything
$ ... {}   -> "patched": 11, "scanned": 13
## 5. confirmed apply
$ ... {"dryRun":false,"confirm":"backfill-skill-search-digest-first-tokens"}
  -> "dryRun": false, "patched": 11, "scanned": 13
```

The mirror action rejects the same way, including on the native path's token.

The part I most wanted off a mock is the scheduler. Re-seeded, one confirmed page of ten,
continuation 15 s out, polled by dry run:

```
[22:07:35] page 1 -> "isDone": false, "patched": 10, "scanned": 10
[22:07:39] poll 1 -> "patched":1
...
[22:07:53] poll 8 -> "patched":1
[22:07:55] poll 9 -> "patched":0
[22:07:57] poll 10 -> "patched":0
```

The eleventh row sits stale for the whole delay and then lands. A continuation that
dropped `confirm` would have thrown on the guard its own first page had just passed, and
that row would still read `"patched":1` at poll 10 - so the propagation is what that
flip measures, not just the paging.

No code change since `bdd16378`; this is evidence only.

Yigtwxx:
Thanks for merging this.

One thing this PR cannot do for itself, flagged here so it does not get lost: the two
first-token backfills still need to be run once against production, native first, then
mirror. Until they are, rows written before this tokenizer keep their old one-character
token while the search computes the new longer one, so they drop out of the first-token
range lookups in `convex/search.ts` — the native ones at `1327-1364` and the mirrored ones
at `1073-1099`.

Both preview by default, so this form counts the affected rows without writing anything
and returns the token needed to apply:

```
npx convex run maintenance:backfillSkillSearchDigestFirstTokens --prod
npx convex run maintenance:backfillSkillsShMirrorDigestFirstTokens --prod
```

```
npx convex run maintenance:backfillSkillSearchDigestFirstTokens \
  '{"dryRun": false, "confirm": "backfill-skill-search-digest-first-tokens"}' --prod
npx convex run maintenance:backfillSkillsShMirrorDigestFirstTokens \
  '{"dryRun": false, "confirm": "backfill-skills-sh-mirror-digest-first-tokens"}' --prod
```

Both accept `{"delayMs": N}` if the default 500 ms between pages is too tight for
production, both are idempotent, so a re-run after a partial pass is safe, and a final dry
run reporting `patched: 0` is the check that nothing is left. @Patrick-Erichsen — happy to
help verify the counts if you want a second pair of eyes on the output.

PR #2299, opened 2026-05-18
feat: scope skill slugs by owner

Patrick-Erichsen:
## Summary

Make skill slug handling owner-aware across ClawHub while preserving legacy slug-only compatibility where the public v1 API still cannot encode an owner namespace.

This PR keeps the platform invariant explicit: skill slugs are unique per publisher namespace, not globally. First-class publish/update/install flows now carry owner context, while legacy reads either resolve deterministically through the compatibility helper or return a controlled ambiguity response.

## What changed

- Adds owner-aware skill/alias lookup helpers in `convex/lib/skills/slugResolution.ts`.
- Updates publish/versioning paths to resolve existing skills by owner namespace, including migration/source-owner handling.
- Updates slug availability, package fallback, downloads, telemetry, security scan, moderation-adjacent reads, backups, and public read paths away from unsafe global `.unique()` assumptions.
- Adds structured ambiguous slug responses for slug-only v1 read/install surfaces.
- Keeps legacy slug-only compatibility by preferring the `openclaw` publisher when duplicate slug matches exist there, otherwise returning controlled ambiguity instead of crashing.
- Updates ClawHub CLI surfaces to support `@owner/slug`, persist owner metadata, and default publish owner to `/api/v1/whoami` when `--owner` is omitted.
- Updates UI/docs install examples to use `@owner/slug` where the owner handle is known.

## Compatibility decisions

- We did **not** make `/api/v1/skills/<slug>` require owner context because v1 URLs cannot encode owner without breaking existing clients.
- We did **not** break old CLI publish clients. Omitted owner remains the legacy/personal publish behavior on the HTTP endpoint.
- New ClawHub CLI publish does send owner context on the wire by defaulting to the authenticated user handle via `whoami`.
- Slug-only legacy reads are intentionally still supported, but they now avoid global uniqueness crashes and make ambiguity explicit.

## User-facing proof

Proof comment with GitHub-hosted assets: https://github.com/openclaw/clawhub/pull/2299#issuecomment-4735499524

**Skill detail page uses owner-qualified route and install ref**

Skill detail page shows owner-qualified route and install ref

**Install command copy uses `@owner/slug`**

Install command uses @owner/slug

**CLI help output**

```text
$ bun packages/clawhub/src/cli.ts install --help
Usage: clawhub install [options] <skill>

Install a skill into <dir>

Arguments:
  skill                Skill to install, e.g. @openclaw/demo
```

```text
$ bun packages/clawhub/src/cli.ts inspect --help
[...]

PR #3383, opened 2026-08-03
fix: enforce publisher skill slug invariant

Patrick-Erichsen:
## Summary

- block skill-transfer acceptance when the destination publisher already owns the incoming canonical slug or any incoming redirect slug
- leave conflicting transfers pending so publishers can rename or merge and retry
- add a bounded invariant scan for multiple active skills with the same `(ownerPublisherId, slug)`
- add an incident-specific, dry-run-first HeartFlow repair that merges three exact loser IDs into three pinned survivors

## Why

The duplicate HeartFlow rows were created when transfer acceptance changed `ownerPublisherId` without checking the recipient publisher's existing skill and redirect namespace. Convex indexes accelerate lookups but are not unique constraints, so the database accepted both rows. The earlier deterministic resolver selection would only mask that invalid state and could select the wrong lineage.

RCA evidence: https://github.com/openclaw/clawhub/pull/3383#issuecomment-5184666680

## Safety

- conflict reads and ownership writes occur in one Convex mutation, so OCC protects concurrent acceptance/publish races
- duplicate reads remain fail-closed; no newest-wins resolver is added
- the scan excludes cross-publisher reuse and soft-deleted merge/history rows
- HeartFlow apply is confirmation-gated, ID-specific, idempotent, and atomically pins both source and survivor latest-version IDs
- production data will not be changed until the deployed dry-run is reviewed separately

## Validation

- `bunx vitest run convex/skillTransfers.duplicateSlug.runtime.test.ts convex/maintenance.duplicateSlug.runtime.test.ts convex/skillTransfers.test.ts convex/maintenance.test.ts --reporter=dot`
- `bun run ci:static`
- `bun run ci:unit`
- `bun run ci:types-build`
- `bun run ci:packages`
- autoreview clean against `origin/main`

### DIFF
--- a/convex/maintenance.duplicateSlug.runtime.test.ts
+++ b/convex/maintenance.duplicateSlug.runtime.test.ts
@@ -1,0 +1,140 @@
+/* @vitest-environment edge-runtime */
+import { convexTest } from "convex-test";
+import { expect, it } from "vitest";
+import { internal } from "./_generated/api";
+import schema from "./schema";
+
+const modules = import.meta.glob("./**/*.ts");
+
+it("reports only active duplicate slugs within the same publisher", async () => {
+  const t = convexTest(schema, modules);
+  const ids = await t.run(async (ctx) => {
+    const userId = await ctx.db.insert("users", { handle: "owner" });
+    const publisherId = await ctx.db.insert("publishers", {
+      kind: "user",
+      handle: "owner",
+      displayName: "Owner",
+      linkedUserId: userId,
+      createdAt: 1,
+      updatedAt: 1,
+    });
+    const otherPublisherId = await ctx.db.insert("publishers", {
+      kind: "org",
+      handle: "other",
+      displayName: "Other",
+      createdAt: 1,
+      updatedAt: 1,
+    });
+    const makeSkill = (ownerPublisherId: typeof publisherId, softDeletedAt?: number) =>
+      ctx.db.insert("skills", {
+        slug: "same-slug",
+        displayName: "Skill",
+        ownerUserId: userId,
+        ownerPublisherId,
+        tags: {},
+        badges: {},
+        moderationStatus: softDeletedAt ? "hidden" : "active",
+        softDeletedAt,
+        stats: { comments: 0, downloads: 0, stars: 0, versions: 0 },
+        createdAt: 1,
+        updatedAt: 1,
+      });
+    const first = await makeSkill(publisherId);
+    const second = await makeSkill(publisherId);
+    await makeSkill(publisherId, 1);
+    await makeSkill(otherPublisherId);
+    return { publisherId, first, second };
+  });
+
+  const result = await t.action(internal.maintenance.scanActivePublisherSlugDuplicatesInternal, {
+    batchSize: 1,
+    maxBatches: 10,
+  });
+
+  expect(result.isDone).toBe(true);
+  expect(result.duplicateGroupsFound).toBe(1);
+  expect(result.findings).toEqual([
+    expect.objectContaining({
+      ownerPublisherId: ids.publisherId,
+      slug: "same-slug",
+      activeSkillIds: expect.arrayContaining([ids.first, ids.second]),
+    }),
+  ]);
+});
+
+it("merges an explicitly selected same-publisher duplicate by id", async () => {
+  const t = convexTest(schema, modules);
+  const ids = await t.run(async (ctx) => {
+    const userId = await ctx.db.insert("users", {
+      handle: "owner",
+      publishedSkills: 2,
+    });
+    const publisherId = await ctx.db.insert("publishers", {
+      kind: "user",
+      handle: "owner",
+      displayName: "Owner",
+      linkedUserId: userId,
+      createdAt: 1,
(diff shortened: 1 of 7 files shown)


Patrick-Erichsen:
## RCA: same-publisher active duplicates

Reviewed exact head `809aac4581e9faec0725d7f4fc9930ccce969efc`. I refreshed the PR/head immediately before posting this comment.

**Verified root cause:** these rows were created by accepted direct skill transfers, not concurrent publishing, import/sync, migration, merge, squat removal, moderation restore, or retry behavior. On 2026-08-02, skills moved from `@mark-heartflow` to `@yun520-1`. Three transfers moved an active lineage into a recipient namespace that already had an active skill with the same slug:

| Slug | Transferred skill / versions | Existing recipient skill / versions | Accepted at (UTC) |
| --- | --- | --- | --- |
| `heartflow` | `kd7dmapx3bfr9j5capz3s3errh87hytn` / 17 (`1.3.5`–`6.4.1`) | `kd7d384mr4pc7ezmf0apvwmqtn8992b6` / 9 (`3.8.1`–`5.7.7`) | `2026-08-02T08:12:28.665Z` |
| `mark-heartflow-skill` | `kd70mrrjtpgs0cw8vpf0vc0g558a5365` / 15 (`5.8.6`–`6.0.66`) | `kd7bhw61fc55a9jd2yb4yfrsg588wmnx` / 28 (`3.0.2`–`5.8.5`) | `2026-08-02T08:12:34.260Z` |
| `heartflow-engine` | `kd72yfs8takacq1d8kefnmsv7h8aq1dg` / 4 (`6.0.9`–`6.0.22`) | `kd78tjvftk71c2xcce7djd2znd88v17h` / 7 (`3.0.0`–`5.7.3`) | `2026-08-02T08:12:43.260Z` |

Production audit rows contain the matching `skill.transfer.request` / `skill.transfer.accept` events. Version creators and newer `publishAttempts.ownerPublisherId` values establish two separate pre-transfer lineages. A complete read-only scan found 216 transfer rows (210 accepted); only these three accepted transfers currently map to active same-publisher duplicate groups. `@yun520-1` has 46 active skills and exactly these three duplicate-slug groups.

The code path is `convex/skillTransfers.ts::acceptTransferInternal`: it ensures the recipient publisher, then patches the transferred row's `ownerUserId` / `ownerPublisherId` and reparents aliases. It never checks whether the destination publisher already owns the raw slug or any transferred alias. This was sequential and deterministic; no race was needed. The normal publish, explicit-owner migration, admin transfer, rename/merge, and GitHub/skills.sh paths do perform destination collision checks. Cross-era provenance is the combination of direct transfers introduced in `460ad3c13dd…`, publisher/alias reparenting in [PR #1344](https://github.com/openclaw/clawhub/pull/1344), and publisher-scoped uniqueness in [PR #2299](https://github.com/openclaw/clawhub/pull/2299) without a corresponding destination collision check in direct-transfer acceptance.

**Invariant/blast radius:** same bare slugs across *different* publishers are valid after owner-scoped slugs. Soft-deleted/non-addressable merge, redirect, and history rows are also valid when only one live row owns the namespace. Multiple active/addressable rows for the *same publisher + slug* are invalid. `skills.by_owner_publisher_slug` is a normal Convex index, not a unique constraint; `.unique()` only asserted uniqueness while reading. Current live symptoms match the breach: bare versions returns 409 with only `yun520-1` as a choice, while owner-qualified reads for all three slugs return 500.

**Verdict on #3383:** not sufficient as a durable fix. Its deterministic newest-row preference masks corrupt active duplicates without preventing another transfer from creating them. “Newest published version” does not prove intended lineage, and the two rows retain split versions, stats, visibility, moderation, tags, and ownership history. Because this shared resolver participates in publish/preflight, sync, scan, rename/merge, and telemetry paths, later writes may attach to the guessed row. It also inspects only 25 matches. The added tests seed invalid rows and assert newest-row service; they do not exercise or close the creation path.

**Smallest safe fix:** at transfer acceptance, transactionally read the recipient namespace for the skill slug and every transferred alias, then reject/cancel on any other active skill or redirect. Doing this at acceptance (not only request time) lets Convex OCC retry concurrent accepts and closes the race as well as the verified sequential path. Keep the resolver fail-closed when more than one active/addressable same-publisher row exists; only collapse inert retained history when exactly one live owner remains. Repair these three groups explicitly by skill ID only after the publisher confirms the intended lineage, preserving versions, aliases/redirects, tags, stats, visibility, and moderation state. Add a one-off invariant scan/monitor plus an audit event for rejected collisions; invariant assertions on restore/reactivation paths are useful defense in depth.

Reproduction: a focused fixture with one active `same-slug` skill in each publisher accepts a transfer and leaves two destination rows on both `origin/main` and this PR head (`expected length 1, got 2`). The PR-focused resolver suite passes (34 tests), confirming #3383 changes read selection but not creation prevention.

Material proof gap: the oldest initial versions predate `publishAttempts`, so their exact original client/endpoint is not recorded. That does not affect the transfer provenance, which is directly evidenced by transfer records, audits, timestamps, ownership changes, and disjoint version lineages. All production inspection was read-only; no data was changed.

PR #3379, opened 2026-08-03
fix: search prints no results when a skills.sh row matches

shbernal:
## What Problem This Solves

`clawhub search <query>` aborted with no output when a mixed result set included a skills.sh mirror:

```
✖ API response: results.N.downloads: a number
```

The released CLI (`clawhub@0.23.1`) expects `downloads` to be an optional number. Canonical mixed search instead emitted `downloads: null` for skills.sh rows, so one mirrored row invalidated the entire response.

## Fix

ClawHub uses one canonical cross-source `downloads` field:

- native rows: ClawHub downloads
- skills.sh rows: skills.sh installs

This PR maps the mirror's persisted `upstreamInstalls` count into `downloads`, keeps the canonical internal/web result type numeric, and leaves the released optional-number client schemas intact. It also adds regression coverage at the canonical search, HTTP response, and both schema-copy boundaries.

## Why This Is The Best Layer

The skills.sh count is already ingested, persisted, and rendered. The bug was introduced when canonical mixed search discarded that value at projection time. Populating the canonical field repairs both current and future clients without requiring a CLI release.

Alternatives rejected:

- accepting `null` only in a future CLI leaves installed clients broken
- omitting the field at HTTP fixes parsing but discards the canonical count
- sending `0` misstates a known nonzero metric

## Validation

- Focused canonical search, HTTP, and schema tests: 529 passed
- `bun run ci:static`: passed
- `bun run ci:unit`: 5,879 passed, 2 skipped
- `bun run ci:packages`: passed
- `bun run ci:types-build`: passed
- Structured autoreview: clean, no actionable findings
- Real released-client proof: `clawhub@0.23.1` parsed the patched production-shaped response and rendered all 20 `find-skills` results, including skills.sh rows with `2,801,613 downloads` and `3,062 downloads`

PR #3385, opened 2026-08-03
fix: emit plain Markdown in generated skill cards

### LINKED ISSUE #3384: Generated skill-card.md ships literal <br> tags on every line
Generated `skill-card.md` files end nearly every line with a literal `<br>` tag. Example from a published card:

```markdown
## Description: <br>
Remove signs of AI-generated writing from text and help rewrite it so it sounds more natural while preserving meaning and tone. <br>

This skill is ready for commercial/non-commercial use. <br>

## Publisher: <br>
[biostartechnology](https://clawhub.ai/user/biostartechnology) <br>

### License/Terms of Use: <br>
MIT <br>
```

## Root cause

The tags are hardcoded in `scripts/skill-cards/templates/clawhub-skill-card.md.j2` — the output maps line-for-line onto the template. Neither Codex nor NVIDIA's `render_card.py` injects them: Codex only produces `skill-card.context.json`, and the renderer just fills the Jinja template.

They were inherited from NVIDIA's public Skill Card template, per the template's own header comment. NVIDIA's cards render into HTML surfaces where Markdown soft newlines collapse, so the `<br>` are load-bearing there. ClawHub copied the section shape and kept them.

## Why this is a problem here

`skill-card.md` is not an HTML surface — it is a Markdown file that ships **inside the installed skill bundle** (`convex/skillCards.ts:464` splices it into `version.files`). Its consumers see the raw text:

- the agent reading the installed `skill-card.md` off disk — the primary consumer
- the Files tab raw viewer
- `clawhub inspect`
- the HTTP API and skills.sh mirrors

The strongest signal that the tags are noise is that **ClawHub's own renderer strips them back out** before display:

- `src/lib/skillCardSections.ts:11` — `normalizeSkillCardTitle` deletes `<br>` from headings
- `src/components/SkillCardPreview.tsx:30` — `BREAK_RE` strips them from description, risk, and output bodies

So the skill detail page looks fine and every other consumer gets the markup. Separately, `## Description: <br>` places a line break inside an `<h2>`, which is already a block element, so the tag is redundant even in an HTML context.

Note that the current shape is pinned by a test — `scripts/skill-cards/run-skill-card-worker.test.ts:160`, *"keeps the neutral template close to NVIDIA's public card shape"* — so this needs an explicit decision rather than a silent fix. The argument for changing it: fidelity to NVIDIA's *section shape* is worth keeping, but fidelity to its *HTML line-break convention* is not, because ClawHub's card is consumed as Markdown.

## Proposed fix

Drop the trailing `<br>` from the template and separate blocks with blank lines instead, keeping consecutive field lines (risk/mitigation pairs, output fields) as distinct paragraphs. The section headings, ordering, and wording stay identical.

The `<br>` strippers in `SkillCardPreview` / `skillCardSections` should stay as-is for backward compatibility with already-generated cards.

## Existing cards

Cards already generated keep their `<br>` until something re-enqueues generation. Note that `"manual"` is a declared job source (`convex/skillCards.ts:21`) with **no caller anywhere in the repo** — only publish and scan completion trigger generation today, so there is currently no path to refresh an existing version's card.

Proposing we do nothing here for now and let cards converge naturally on the next publish or rescan. If a backfill is wanted later, there are two options:

1. **Re-run generation** — add the missing manual-enqueue path plus a sweep. This burns a Codex run per version and re-rolls previously reviewed card *prose* for what is a purely cosmetic change.
2. **Rewrite the stored Markdown deterministically** — the `<br>` removal is content-preserving, so the stored file can be rewritten without invoking Codex at all.

Option 2 is clearly cheaper and avoids prose churn, and the primitives for it already exist:

- `selectGeneratedSkillCardFile` (`convex/lib/skillCards.ts:44`) identifies generated cards via `generated-bundle` fingerprints, so a migration would skip publisher-authored `skill-card.md` files
- `replaceGeneratedSkillCardFile` + `attachCardAndSucceedJobInternal` (`convex/skillCards.ts:464-490`) already implement the file swap and preserve historical fingerprints, so installs holding an older card keep resolving
- `runPluginManifestSummaryBackfill` (`convex/migrations.ts:1109`) is a working precedent for the required shape — a paged `internalAction` (needed because `ctx.storage.store` is action-only) over an `internalQuery` candidate list, applying an `internalMutation` patch

We are happy to write that migration as a follow-up if maintainers want the backfill.

shbernal:
Fixes #3384.

## Summary

On `main`, generated `skill-card.md` files end nearly every content line with a literal `<br>`. The tags come from `scripts/skill-cards/templates/clawhub-skill-card.md.j2`, inherited from NVIDIA's public Skill Card template, where they support HTML-oriented rendering.

ClawHub ships the generated card inside the installed skill bundle and exposes the raw Markdown through the CLI and HTTP API. This PR removes the HTML line-break tags at the template source and separates semantic blocks with blank lines instead. Section headings, ordering, and wording stay unchanged.

## Before / after

```diff
-## Description: <br>
-Remove signs of AI-generated writing while preserving meaning and tone. <br>
+## Description:
+
+Remove signs of AI-generated writing while preserving meaning and tone.

-## Publisher: <br>
-[biostartechnology](https://clawhub.ai/user/biostartechnology) <br>
+## Publisher:
+
+[biostartechnology](https://clawhub.ai/user/biostartechnology)
```

Risk/mitigation pairs and the four output fields remain separate Markdown paragraphs rather than collapsing into run-on text.

## Compatibility

- The existing `<br>` normalization in `SkillCardPreview` and `skillCardSections` remains for previously generated cards.
- This PR is not expected to change the Skill Card tab visually: old and new output parse into the same 10 sections, risk pairs, and output fields.
- Existing cards remain unchanged until a publish or rescan re-enqueues generation. A historical backfill is intentionally out of scope.
- No schema, API, or job-queue contract changes.

## Testing

- `bunx vitest run scripts/skill-cards/run-skill-card-worker.test.ts` — 16 passed
- `bun run ci:static` — pass
- `bun run ci:unit` — 448 files passed, 1 skipped; 5,902 tests passed, 2 skipped
- Rendered with `jinja2==3.1.6` using populated/default-risk and missing-license contexts; output retained all sections with no `<br>` tags

### DIFF
--- a/scripts/skill-cards/run-skill-card-worker.test.ts
+++ b/scripts/skill-cards/run-skill-card-worker.test.ts
@@ -161,19 +161,30 @@ describe("run-skill-card-worker Codex skill setup", () => {
   it("keeps the neutral template close to NVIDIA's public card shape", async () => {
     const template = await readFile(neutralTemplatePath(), "utf8");
 
-    expect(template).toContain("## Description: <br>");
+    expect(template).toContain("## Description:");
     expect(template).toContain("## Publisher:");
-    expect(template).toContain("### License/Terms of Use: <br>");
+    expect(template).toContain("### License/Terms of Use:");
     expect(template).toContain("license_identifier is defined");
-    expect(template).toContain("## Use Case: <br>");
-    expect(template).toContain("### Deployment Geography for Use: <br>");
-    expect(template).toContain("## Known Risks and Mitigations: <br>");
-    expect(template).toContain("## Reference(s): <br>");
-    expect(template).toContain("## Skill Output: <br>");
-    expect(template).toContain("## Skill Version(s): <br>");
+    expect(template).toContain("## Use Case:");
+    expect(template).toContain("### Deployment Geography for Use:");
+    expect(template).toContain("## Known Risks and Mitigations:");
+    expect(template).toContain("## Reference(s):");
+    expect(template).toContain("## Skill Output:");
+    expect(template).toContain("## Skill Version(s):");
     expect(template).not.toContain("Third-Party Community Consideration");
     expect(template).not.toContain("Provenance");
     expect(template).not.toContain("For Release on NVIDIA Platforms Only");
+  });
+
+  // skill-card.md ships inside the installed skill bundle and is read as plain
+  // Markdown by agents, the Files tab, the CLI, and the HTTP API. NVIDIA's
+  // upstream template ends every content line with <br> because its cards
+  // render into HTML surfaces; keeping those tags leaked literal markup into
+  // every raw consumer, so the neutral template must stay <br>-free.
+  it("emits plain Markdown with no <br> line-break tags", async () => {
+    const template = await readFile(neutralTemplatePath(), "utf8");
+
+    expect(template).not.toMatch(/<br\s*\/?>/i);
   });
 
   it("rejects NVIDIA-only public-card boilerplate", () => {
--- a/scripts/skill-cards/templates/clawhub-skill-card.md.j2
+++ b/scripts/skill-cards/templates/clawhub-skill-card.md.j2
@@ -1,64 +1,80 @@
 {#-
 ClawHub Skill Card template.
 
-This intentionally stays close to NVIDIA's public Skill Card template shape
+This intentionally stays close to NVIDIA's public Skill Card section shape
 while replacing NVIDIA-only ownership, review, and platform boilerplate with
 neutral public-card language.
+
+Unlike NVIDIA's template this emits plain Markdown with no trailing HTML
+line-break tags. NVIDIA's cards render into HTML surfaces where soft line
+breaks collapse; ClawHub's skill-card.md ships inside the installed skill
+bundle and is read as Markdown by agents, the Files tab, the CLI, and the HTTP
+API, so blocks are separated with blank lines instead. Keep consecutive field
+lines blank-line separated so they stay distinct paragraphs.
 -#}
 
-## Description: <br>
-{{ description_sentence }} <br>
+## Description:
+
+{{ description_sentence }}
 
 {% if usage_posture == "commercial" -%}
-This skill is ready for commercial/non-commercial use. <br>
+This skill is ready for commercial/non-commercial use.
 {%- elif usage_posture == "research_dev" -%}
-This skill is for research and development only. <br>
+This skill is for research and development only.
 {%- elif usage_posture == "demonstration" -%}
-This skill is for demonstration purposes and not for production usage. <br>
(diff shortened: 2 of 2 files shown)


vyctorbrzezowski (commented):
This fixes generated `skill-card.md` at the source: the ClawHub-owned Jinja template now emits plain Markdown while the existing UI normalization remains in place for older cards.

LOC: +128/-53 (10 files) on the current head; the intended fix is +74/-47 across 2 files after removing the superseded search commit.

**Finding:** the implementation itself has no blocking code issue, but the branch is not currently landable. It still contains `6eb0af8b`, an earlier version of the search fix that has already landed as #3379, so GitHub reports `mergeable=CONFLICTING` / `mergeStateStatus=DIRTY`.

**Best-fix verdict:** best. Removing the tags in the template is better than stripping them after generation, in storage, or independently in every consumer; those alternatives would leave the raw bundle/API/CLI contract inconsistent. Existing-card backfill remains correctly separate from this forward fix.

**Evidence checked:** issue repro and root cause; template blame/history; worker render path; generated-card persistence; HTTP card endpoint; `clawhub inspect --card`; web section/risk/output parsers. A Jinja 3.1.6 render produced all 10 sections with no `<br>` and preserved the parsed structure. The focused worker test passed 14/14, failed on the old template for the expected `<br>` assertion, and passed again after restoration.

**Provenance:** introduced by Patrick-Erichsen in #2382 on 2026-05-25; current fix authored by Santiago (`@shbernal`) in #3385 on 2026-08-03.

**Remaining uncertainty:** required static/unit checks must run on the final rebased SHA. The current Vercel failure is contributor authorization, not a code failure.

Maintainer verdict: the fix is accepted in shape; branch repair and exact-head validation are required before approval. I am taking the branch through that maintenance pass now. No merge is being performed.

vyctorbrzezowski (approved):
No blocking findings on the final head. The branch is linear on current main, the superseded #3379 search commit is gone, and the intended diff is limited to the skill-card template plus its regression coverage. Best-fix verdict: this is the correct source-level fix; downstream stripping or per-consumer normalization would leave the raw card contract inconsistent. Exact-head CI is green across static, unit, packages, types/build, HTTP e2e, browser, CodeQL, secrets, and Socket checks. The remaining Vercel status is fork preview authorization, not a code failure. Santiago remains the commit author and Vyctor is preserved as co-author for the maintainer adjustments. Existing-card backfill remains intentionally out of scope. Approved; no merge performed.

vyctorbrzezowski:
Thanks @shbernal for tracking this down and fixing it at the template source. The focused regression coverage and compatibility notes made the change straightforward to validate and land.

PR #3396, opened 2026-08-04
fix: report broken worktree source symlinks

vyctorbrzezowski:
## What Problem This Solves

Fixes an issue where contributors running `setup:worktree` against a source worktree with dangling `.env.local` or `.convex` symlinks received only a generic “Could not find a usable worktree source” error.

## Why This Change Was Made

Source selection now distinguishes genuinely absent ignored state from dangling links and reports each broken target in the existing rejected-source diagnostics. `.convex` remains optional for remote deployments; the broken-link rejection applies only after `.env.local` identifies a `local:` deployment.

The managed-worktree environment configuration is unchanged: managed worktrees already copy `.worktreeinclude` state before validation, while Worktrunk’s existing pre-start hook owns the `copy-ignored` plus `--force --prefer-fallback` repair path.

## User Impact

Contributors see the exact broken local-state links that require repair. Remote-backed sources remain usable when an unrelated `.convex` symlink is dangling.

## Evidence

The focused suite now covers both broken-link diagnostics and the remote-source compatibility case:

- broken `.env.local` and `.convex` links are reported with their targets
- a readable local deployment rejects a broken `.convex` link
- a remote deployment accepts the same unused broken `.convex` link

Validation on current main `f9ea25e1`:

- `bunx vitest run scripts/setup-worktree.test.ts` — 19 passed
- `VITE_CONVEX_URL=https://example.invalid bun run ci:unit` — 5,922 passed, 2 skipped

### DIFF
--- a/scripts/setup-worktree.test.ts
+++ b/scripts/setup-worktree.test.ts
@@ -6,6 +6,7 @@ import {
   mkdirSync,
   realpathSync,
   rmSync,
+  symlinkSync,
   writeFileSync,
 } from "node:fs";
 import { tmpdir } from "node:os";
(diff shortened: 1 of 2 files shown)


vyctorbrzezowski:
@clawsweeper review

Patrick-Erichsen (approved):
This makes worktree setup distinguish ordinary missing local state from dangling `.env.local` or `.convex` symlinks, so the operator gets the actual repair target instead of a generic rejected-source result.

LOC: +65/-1 (2 files)

No blocking findings.

Best-fix verdict: best. The diagnostic belongs before source reading, preserves normal fallback for genuinely absent state, and has focused coverage for both broken-link shapes.

Alternatives considered: changing `readSource` would conflate missing state with an invalid candidate; repairing or replacing links automatically would be an unsafe mutation during source discovery.

Code read: `scripts/setup-worktree.ts`, adjacent source validation and fallback flow, and `scripts/setup-worktree.test.ts`. Exact-head required checks are green.

Remaining uncertainty: none material for this developer-only diagnostic.

vyctorbrzezowski:
Addressed the remaining compatibility case in `0e023c86`: `.convex` is now inspected only after `.env.local` identifies a `local:` deployment. A remote-backed source with an unused dangling `.convex` link remains eligible, with a focused regression test. Current-main rebase, 19 focused tests, `ci:static`, and `ci:unit` are green.

PR #3399, opened 2026-08-04
fix(web): bound canonical skills SSR loading

vyctorbrzezowski:
Related: #2925

## What Problem This Solves

The canonical Skills catalog currently renders only a loading shell on the server and repeats its capability and Trending requests after hydration. Moving both remote calls unconditionally onto the document critical path, however, raised measured median TTFB from 9 ms to 1,516 ms.

## Why This Change Was Made

The default, unfiltered Trending page now gets a 250 ms initial-data budget. If the public catalog responds inside that budget, its first 20 rows are included in the HTML and reused during hydration. If either dependency is slow, the loader releases the shell and the existing client fetch takes over. Search, filtered tabs, view-only changes, and pagination retain their existing paths.

The release-owned changelog edit was removed while rebasing onto current main.

## User Impact

Fast catalog responses can render real rows before hydration without a duplicate request. Slow responses no longer hold the entire document open for seconds; users get the existing loading state and a single client Trending request.

## Evidence

Production build against the same public backend used for the original A/B:

| Measurement | current main | proposed `d40c81f4` |
| --- | ---: | ---: |
| Warm browser response start | 5 ms median | 259 ms median |
| Slow-backend HTML | 6 skeleton rows | 6 skeleton rows after the 250 ms budget |
| Hydrated Trending rows | 20 | 20 |
| Browser Trending requests | 1 | 1 |
| First list visible, 5 paired samples | 3,211 ms median | 2,333 ms median |

Seven direct production-server samples put the proposed warm TTFB at 262–273 ms; the cold first request was 553 ms. A real Chrome run confirmed the fallback changes six skeleton rows into 20 real rows with one Trending request. Focused regression coverage also proves the fast path seeds the server rows without a duplicate request and the slow path aborts at the explicit budget.

Validation:

- `bunx vitest run src/__tests__/skills-index.test.tsx src/__tests__/skills-index-load-more.test.tsx src/__tests__/skills-index.claw591.test.tsx src/__tests__/skills-route-default-sort.test.ts src/routes/skills/-params.test.ts` — 71 passed
- `VITE_CONVEX_URL=https://example.invalid bun run ci:unit` — 5,922 passed, 2 skipped
- `bun run ci:types-build` — passed
- production Nitro TTFB sampling and real Chrome fallback verification — passed

### DIFF
--- a/src/__tests__/skills-index.test.tsx
+++ b/src/__tests__/skills-index.test.tsx
@@ -2,7 +2,11 @@
 import { act, fireEvent, render, screen } from "@testing-library/react";
 import type { ReactNode } from "react";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
-import { Route as SkillsRoute, SkillsIndex } from "../routes/skills/index";
+import {
+  Route as SkillsRoute,
+  SKILLS_INITIAL_PAGE_TIMEOUT_MS,
+  SkillsIndex,
+} from "../routes/skills/index";
 import {
   convexHttpMock,
   convexReactMocks,
@@ -12,6 +16,7 @@ import {
 
 const navigateMock = vi.fn();
 const fetchCatalogDiscoveryCapabilitiesMock = vi.fn();
+const fetchCanonicalTrendingPageMock = vi.fn();
 let searchMock: Record<string, unknown> = {};
 let loaderDataMock: unknown = null;
 
@@ -19,6 +24,14 @@ vi.mock("../lib/catalogDiscoveryCapabilities", () => ({
   fetchCatalogDiscoveryCapabilities: (...args: unknown[]) =>
     fetchCatalogDiscoveryCapabilitiesMock(...args),
 }));
+
+vi.mock("../lib/trendingApi", async (importOriginal) => {
+  const original = await importOriginal<typeof import("../lib/trendingApi")>();
+  return {
+    ...original,
+    fetchCanonicalTrendingPage: (...args: unknown[]) => fetchCanonicalTrendingPageMock(...args),
+  };
+});
 
 vi.mock("@tanstack/react-router", () => ({
   createFileRoute: () => (config: { component: unknown; validateSearch: unknown }) => ({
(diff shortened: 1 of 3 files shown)


vyctorbrzezowski:
## ClawHub UI Proof

Status: `pass`
Mode: `before-after`
Scenario: `bounded Skills SSR fallback and hydration`
Provider: `local-production-browser`
Baseline: `upstream/main@f9ea25e14f14c05f9544d3878d1626d0dcec4b7a`
Candidate: `contrib/skills-canonical-ssr@d40c81f4fc887e750ff04f2431eb041b8a56476f`

| Initial shell from slow backend | Initial shell after 250 ms budget |
| --- | --- |
| Initial shell from slow backend | Initial shell after 250 ms budget |

| Hydrated Trending results | Hydrated Trending results |
| --- | --- |
| Hydrated Trending results | Hydrated Trending results |

Raw proof files: https://github.com/openclaw/clawhub/tree/qa-artifacts/clawhub-ui-proof/pr-3399/skills-ssr-budget-d40c81f4

vyctorbrzezowski:
@clawsweeper review

Patrick-Erichsen (dismissed):
This server-renders the initial Trending catalog page so users see skill cards before hydration. The supplied proof shows the visual benefit, but it also moves capability and Trending requests onto the document-response critical path.

LOC: +240/-38 (4 files)

Findings: remove the release-owned changelog edit and resolve the measured latency tradeoff before merge: median server TTFB rises from 9 ms to 1,516 ms in the provided A/B result.

Best-fix verdict: acceptable mitigation, not yet the best fix. Reusing the loader result client-side is good; synchronously blocking the whole HTML response on both remote calls needs an explicit latency policy or a streamed/deferred shape.

Alternatives considered: keep the client fetch (fast shell, slower content), or stream/defer the catalog payload so shell TTFB stays bounded while avoiding the duplicate client request.

Code read: `/skills` loader, browse-model initial-list hydration, Trending/capability clients, tests, and submitted before/after proof.

Remaining uncertainty: the desired product priority between shell TTFB and first visible catalog content has not been decided; the branch is now behind main.

vyctorbrzezowski:
Addressed both requested changes in `d40c81f4`: the release-owned changelog edit is gone, and canonical initial data now has an explicit 250 ms budget. Slow dependencies release the shell and abort the server fetch; the existing client path then loads the page once. Warm response start measured 259 ms median instead of the prior 1,516 ms TTFB, while five paired browser samples showed the first list at 2,333 ms median versus 3,211 ms on main. The refreshed proof comment is on the final SHA.

PR #3471, opened 2026-08-15
fix: abort hung OG metadata fetches and archive metrics

SebTardif:
## What Problem This Solves

Public OG routes (`/og/skill`, `/og/plugin`, `/og/profile`) fetched registry or Convex metadata with no abort. A stalled backend parked the OG function until the platform killed it. Sibling image fetch already aborts at 1.5s, and that deadline has to cover the JSON body, not only response headers.

Archive downloads awaited an untimed metric POST before returning the first zip byte, so a hung Convex metric blocked the download.

## Evidence

terminal output from a local HTTP origin on this machine, head `d20f8fea`:

```console
$ bun /tmp/clawhub-3471-live-proof.ts
OG_BODY_STALL meta=null elapsed_ms=1502 hits=1
ZIP_FIRST_BYTE status=200 bytes=38 elapsed_ms=5 metric_hits=0
```

The origin wrote `200` plus `content-type: application/json` for `/api/v1/skills/gifgrep` and never sent a body. `fetchSkillOgMeta` returned null at 1502ms. In the same process, the first archive zip byte arrived in 5ms while the metric POST was left hanging.

## Real behavior proof

- **Behavior or issue addressed:** Skill and plugin OG metadata lookups abort at 1.5s through JSON decode; archive metric POST cannot delay the first zip byte.
- **Real environment tested:** macOS, Bun, local Node HTTP listener on 127.0.0.1, clawhub worktree `/tmp/oc-impl-clawhub-og` head `d20f8fea`.
- **Exact steps or command run after this patch:**

  ```bash
  bun /tmp/clawhub-3471-live-proof.ts
  ```

- **Evidence after fix:** terminal output from the patched worktree:

  ```console
  $ bun /tmp/clawhub-3471-live-proof.ts
  OG_BODY_STALL meta=null elapsed_ms=1502 hits=1
  ZIP_FIRST_BYTE status=200 bytes=38 elapsed_ms=5 metric_hits=0
  ```

- **Observed result after fix:** A headers-then-stall JSON origin returns null at 1.5s. The first zip byte is 38 bytes at 5ms and does not wait for the metric POST.
- **What was not tested:** A live production Convex outage.

## Change

- `fetchSkillOgMeta` and `fetchPluginOgMeta` parse `response.json()` inside `withOgFetchTimeout`, matching `fetchPublisherOgMeta`.
- Body-stall coverage for both fetchers (headers arrive, JSON never does).
- Spec: the 1.5s deadline covers transport and JSON decode.

### DIFF
--- a/server/og/fetchPluginOgMeta.ts
+++ b/server/og/fetchPluginOgMeta.ts
@@ -21,21 +21,25 @@ export async function fetchPluginOgMeta(
 ): Promise<PluginOgMeta | null> {
   try {
     const url = new URL(`/api/v1/packages/${encodeURIComponent(packageName)}`, apiBase);
-    const response = await withOgFetchTimeout((signal) =>
-      fetch(url.toString(), { headers: { Accept: "application/json" }, signal }),
-    );
-    if (!response.ok) return null;
-    const payload = (await response.json()) as {
-      package?: {
-        name?: string;
-        displayName?: string;
-        summary?: string | null;
-        latestVersion?: string | null;
-        stats?: unknown;
-        verification?: { scanStatus?: string | null } | null;
-      } | null;
-      owner?: { handle?: string | null; image?: string | null } | null;
-    };
+    const payload = await withOgFetchTimeout(async (signal) => {
+      const response = await fetch(url.toString(), {
+        headers: { Accept: "application/json" },
+        signal,
+      });
+      if (!response.ok) return null;
+      return (await response.json()) as {
+        package?: {
+          name?: string;
+          displayName?: string;
+          summary?: string | null;
+          latestVersion?: string | null;
+          stats?: unknown;
+          verification?: { scanStatus?: string | null } | null;
+        } | null;
+        owner?: { handle?: string | null; image?: string | null } | null;
+      };
+    });
+    if (!payload) return null;
     const stats = readStats(payload.package?.stats);
     return {
       name: payload.package?.name ?? null,
--- a/server/og/fetchPluginOgMeta.ts
+++ b/server/og/fetchPluginOgMeta.ts
@@ -1,3 +1,5 @@
+import { withOgFetchTimeout } from "./ogFetchTimeout";
+
 export type PluginOgMeta = {
   name: string | null;
   displayName: string | null;
@@ -19,7 +21,9 @@ export async function fetchPluginOgMeta(
 ): Promise<PluginOgMeta | null> {
   try {
     const url = new URL(`/api/v1/packages/${encodeURIComponent(packageName)}`, apiBase);
-    const response = await fetch(url.toString(), { headers: { Accept: "application/json" } });
+    const response = await withOgFetchTimeout((signal) =>
+      fetch(url.toString(), { headers: { Accept: "application/json" }, signal }),
+    );
     if (!response.ok) return null;
     const payload = (await response.json()) as {
       package?: {
--- a/server/og/fetchSkillOgMeta.ts
+++ b/server/og/fetchSkillOgMeta.ts
@@ -1,4 +1,5 @@
 import { readCanonicalStat, type SkillStatReadable } from "../../convex/lib/skillStats";
+import { withOgFetchTimeout } from "./ogFetchTimeout";
 
 type SkillApiPayload = SkillStatReadable & {
   displayName?: string;
@@ -43,7 +44,9 @@ export async function fetchSkillOgMeta(
     const url = new URL(`/api/v1/skills/${encodeURIComponent(slug)}`, apiBase);
     const owner = ownerHandle?.trim().replace(/^@+/, "");
     if (owner) url.searchParams.set("ownerHandle", owner);
-    const response = await fetch(url.toString(), { headers: { Accept: "application/json" } });
+    const response = await withOgFetchTimeout((signal) =>
+      fetch(url.toString(), { headers: { Accept: "application/json" }, signal }),
+    );
     if (!response.ok) return null;
     const payload = (await response.json()) as {
       skill?: SkillApiPayload | null;
--- a/server/og/fetchSkillOgMeta.ts
+++ b/server/og/fetchSkillOgMeta.ts
@@ -44,20 +44,24 @@ export async function fetchSkillOgMeta(
     const url = new URL(`/api/v1/skills/${encodeURIComponent(slug)}`, apiBase);
     const owner = ownerHandle?.trim().replace(/^@+/, "");
     if (owner) url.searchParams.set("ownerHandle", owner);
-    const response = await withOgFetchTimeout((signal) =>
-      fetch(url.toString(), { headers: { Accept: "application/json" }, signal }),
-    );
-    if (!response.ok) return null;
-    const payload = (await response.json()) as {
-      skill?: SkillApiPayload | null;
-      owner?: { handle?: string | null; image?: string | null } | null;
-      latestVersion?: { version?: string | null } | null;
-      moderation?: {
-        verdict?: "clean" | "suspicious" | "malicious";
-        isSuspicious?: boolean;
-        isMalwareBlocked?: boolean;
-      } | null;
-    };
+    const payload = await withOgFetchTimeout(async (signal) => {
+      const response = await fetch(url.toString(), {
+        headers: { Accept: "application/json" },
+        signal,
+      });
+      if (!response.ok) return null;
+      return (await response.json()) as {
+        skill?: SkillApiPayload | null;
+        owner?: { handle?: string | null; image?: string | null } | null;
+        latestVersion?: { version?: string | null } | null;
+        moderation?: {
+          verdict?: "clean" | "suspicious" | "malicious";
+          isSuspicious?: boolean;
+          isMalwareBlocked?: boolean;
+        } | null;
+      };
+    });
+    if (!payload) return null;
     return {
       displayName: payload.skill?.displayName ?? null,
       summary: payload.skill?.summary ?? null,
--- a/server/og/ogFetchTimeout.ts
+++ b/server/og/ogFetchTimeout.ts
@@ -1,0 +1,31 @@
+// Match fetchImageDataUrl: public OG cards must not wait forever on a hung origin.
+export const OG_FETCH_TIMEOUT_MS = 1_500;
+
+export async function withOgFetchTimeout<T>(
+  work: (signal: AbortSignal) => Promise<T>,
+  timeoutMs = OG_FETCH_TIMEOUT_MS,
+): Promise<T> {
+  const controller = new AbortController();
+  const timeout = setTimeout(() => controller.abort(), timeoutMs);
+  const aborted = new Promise<never>((_, reject) => {
+    const rejectOnAbort = () => {
+      reject(
+        controller.signal.reason instanceof Error
+          ? controller.signal.reason
+          : new DOMException("The operation was aborted.", "AbortError"),
+      );
+    };
+    if (controller.signal.aborted) {
+      rejectOnAbort();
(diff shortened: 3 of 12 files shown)


server/og/ogFetchTimeout.ts
```diff
@@ -0,0 +1,31 @@
...
+      return;
+    }
+    controller.signal.addEventListener("abort", rejectOnAbort, { once: true });
+  });
+  // Mark handled so abort cannot become an unhandled rejection if work() ignores the signal.
+  void aborted.catch(() => undefined);
+  try {
+    return await Promise.race([work(controller.signal), aborted]);
+  } finally {
+    clearTimeout(timeout);
```
Yigtwxx, line 29: The deadline is disarmed as soon as `work()` resolves, which for `fetch` is when the response headers arrive. The body read then runs with no bound at all, so an origin that answers with headers and stalls the body still hangs the card forever.

`fetchImageDataUrl`, which the comment above says this matches, does the opposite: the timer is armed at `server/og/fetchImageDataUrl.ts:73`, `readLimitedImageBody` runs at `:79`, and `clearTimeout` only happens at `:83`, after both phases.

Reproduced against 546b13e6 with a body that stalls but honours the signal, like a real fetch body:

```ts
function stallingBody(signal: AbortSignal | null | undefined) {
  return new Promise<never>((_resolve, reject) => {
    if (!signal) return;
    signal.addEventListener(
      "abort",
      () => reject(new DOMException("The operation was aborted.", "AbortError")),
      { once: true },
    );
  });
}

it("fetchSkillOgMeta: never settles when the body stalls", async () => {
  vi.useFakeTimers();
  vi.stubGlobal(
    "fetch",
    vi.fn(async (_input: unknown, init?: RequestInit) =>
      ({ ok: true, json: () => stallingBody(init?.signal) }) as unknown as Response),
  );

  let settled = false;
  void fetchSkillOgMeta("gifgrep", "https://clawhub.ai").then(() => { settled = true; });

  await vi.advanceTimersByTimeAsync(OG_FETCH_TIMEOUT_MS * 20);
  await Promise.resolve();
  expect(settled).toBe(false); // passes: 30s of virtual time, still pending
});
```

The identical mock driven through `fetchImageDataUrl` resolves `null` at 1.5s. Both assertions pass on this branch, so the two helpers are not equivalent today.

The fix that keeps the stated model is to hold the controller until the caller is done with the response, i.e. put the parse inside the wrapper as well. `fetchPublisherOgMeta` in this PR already gets that right, because the whole Convex query (transport plus decode) sits inside `withOgFetchTimeout`.

server/og/fetchSkillOgMeta.ts
```diff
@@ -43,7 +44,9 @@ export async function fetchSkillOgMeta(
     const url = new URL(`/api/v1/skills/${encodeURIComponent(slug)}`, apiBase);
     const owner = ownerHandle?.trim().replace(/^@+/, "");
     if (owner) url.searchParams.set("ownerHandle", owner);
-    const response = await fetch(url.toString(), { headers: { Accept: "application/json" } });
+    const response = await withOgFetchTimeout((signal) =>
+      fetch(url.toString(), { headers: { Accept: "application/json" }, signal }),
+    );
     if (!response.ok) return null;
     const payload = (await response.json()) as {
```
Yigtwxx: This `response.json()` runs after `withOgFetchTimeout` has already returned and cleared its timer, so the parse is unbounded. `server/routes/og/skill.png.ts:65` awaits `fetchSkillOgMeta` directly, so a stalled body still stalls `/og/skill` for as long as the connection stays open, which is the failure this PR is fixing.

The new test only exercises the header phase: `hangUntilAborted` never resolves a `Response`, so `fetch` itself is what gets aborted. A body-phase case is what would pin the invariant the spec claims.

server/og/fetchPluginOgMeta.ts
```diff
@@ -19,7 +21,9 @@ export async function fetchPluginOgMeta(
 ): Promise<PluginOgMeta | null> {
   try {
     const url = new URL(`/api/v1/packages/${encodeURIComponent(packageName)}`, apiBase);
-    const response = await fetch(url.toString(), { headers: { Accept: "application/json" } });
+    const response = await withOgFetchTimeout((signal) =>
+      fetch(url.toString(), { headers: { Accept: "application/json" }, signal }),
+    );
     if (!response.ok) return null;
     const payload = (await response.json()) as {
```
Yigtwxx: Same shape as the skill fetcher: the timer is cleared when the headers land, and this parse is outside it. Two of the three metadata paths added here bound only the transport; `fetchPublisherOgMeta` bounds transport plus decode. Worth making all three consistent, since `specs/og-routes.md` states one rule for all of them.

Yigtwxx (commented):
Read this against 546b13e6 and ran the affected suites locally.

The archive-metric half holds up. `does not wait for download metrics before emitting zip bytes` drives a genuinely hung metric origin and asserts the first zip chunk arrives anyway, rather than asserting the call shape, which is the kind of proof this change needs. One nit: it races a real 500ms `setTimeout` against `reader.read()`, the only wall-clock dependency in that file; fake timers would keep it deterministic on a loaded runner.

The OG half bounds less than it claims. `withOgFetchTimeout` clears its timer as soon as `work()` resolves, which for `fetch` is the header phase, so `await response.json()` in the skill and plugin fetchers runs with no deadline. `specs/og-routes.md:6` says these fetches match `fetchImageDataUrl`, and `fetchImageDataUrl` holds its timer across the body read (`server/og/fetchImageDataUrl.ts:73-83`). `fetchPublisherOgMeta` is the one path in this PR that covers transport and decode together.

A reproduction that passes on this branch (stalled body, signal honoured, 30s of virtual time, still pending) is in the inline comment on `server/og/ogFetchTimeout.ts`, along with the contrast against `fetchImageDataUrl` under the same mock.

SebTardif:
@Yigtwxx

> One nit: it races a real 500ms `setTimeout` against `reader.read()`, the only wall-clock dependency in that file; fake timers would keep it deterministic on a loaded runner.

Done in e6e70b09. That test now uses `vi.useFakeTimers()` and `vi.advanceTimersByTimeAsync(500)` for the first-byte watchdog, matching the OG hang tests in this PR. Local run: `bunx vitest run server/convexProxy.test.ts` (18 passed).

Your comments on `withOgFetchTimeout` clearing after headers (skill/plugin `response.json()` still unbounded) are still open. I have not changed that path in this commit.

Yigtwxx:
Verified `e6e70b09` on the PR head: `server/convexProxy.test.ts` gives 18 passed here too.

I also ran the negative control, since a fake-timer rewrite can quietly stop asserting anything.
Restoring only `server/convexProxy.ts` from current main (`f98a436a`) while keeping your test file
gives `2 failed | 16 passed`, and the hung-metric case fails with the intended message:

```
FAIL  server/convexProxy.test.ts > Convex HTTP proxy > does not wait for download metrics before emitting zip bytes
Error: zip first byte blocked by download metric
 ❯ server/convexProxy.test.ts:706:15
```

The second failure is the `signal` assertion at `:229`, which is expected — main's metric fetch has
no `signal`. So the rewrite kept its teeth.

One note for whoever reads that test later: `vi.advanceTimersByTimeAsync(500)` resolves
unconditionally rather than only on a hang, so the race is now "the first read wins the
timer/microtask drain" rather than a 500 ms wall-clock bound. That is stricter than the old
version, not looser, which is why the control still fails — but the `500` no longer means 500 ms.

On the open threads: agreed, and I confirmed nothing at the current head touches that path.
`fetchSkillOgMeta.ts:47-51` and `fetchPluginOgMeta.ts:24-28` still resolve `work()` at the header
phase and parse outside the armed window.

The shape that closes it is already in this PR: `fetchPublisherOgMeta.ts:31-33` puts the whole
operation inside `work()`, so its deadline covers decode. The same move for the two JSON fetchers,
with no change to `withOgFetchTimeout` itself:

```ts
const payload = await withOgFetchTimeout(async (signal) => {
  const response = await fetch(url.toString(), { headers: { Accept: "application/json" }, signal });
  if (!response.ok) return null;
  return (await response.json()) as { /* unchanged */ };
});
if (!payload) return null;
```

That would also make all three paths actually match the `fetchImageDataUrl` parity that
`specs/og-routes.md:6` claims.

server/og/ogFetchTimeout.ts
SebTardif: @Yigtwxx

> put the parse inside the wrapper as well. `fetchPublisherOgMeta` in this PR already gets that right

Done in d20f8fea. Skill and plugin now run `fetch` plus `response.json()` inside `withOgFetchTimeout`. No change to the helper itself.

A local HTTP origin that sent headers and stalled the body returned null at 1502ms.

server/og/fetchSkillOgMeta.ts
SebTardif: @Yigtwxx

> The new test only exercises the header phase

Added `aborts when the public skill API returns headers then stalls the JSON body`. Red: the case timed out at 3000ms on the old fetch-only wrapper. Green: null after 1.5s of fake time, signal aborted.

Production parse is inside `work()` in d20f8fea.

server/og/fetchPluginOgMeta.ts
SebTardif: @Yigtwxx

> Two of the three metadata paths added here bound only the transport

Plugin now matches publisher: transport plus decode sit inside `withOgFetchTimeout` (d20f8fea). Same headers-then-stall coverage as the skill fetcher.

SebTardif:
@clawsweeper re-review

Skill and plugin JSON parse now sit inside the 1.5s deadline (d20f8fea). Live local HTTP origin: headers-then-stall returned null at 1502ms; first zip byte arrived in 5ms.

Yigtwxx (commented):
Re-checked at `d20f8fea`. Both points are closed, and I drove red and green locally in both directions rather than reading the description.

**OG decode deadline.** `fetchSkillOgMeta` and `fetchPluginOgMeta` now run `fetch` plus `response.json()` inside `withOgFetchTimeout`, so the controller stays armed across the body phase. All three metadata paths bound transport and decode, and `specs/og-routes.md:6-8` states the rule the code actually follows. The reproduction I posted no longer reproduces.

Green on `d20f8fea`:

```console
$ npx vitest run server/og/fetchSkillOgMeta.test.ts server/og/fetchPluginOgMeta.test.ts server/og/fetchPublisherOgMeta.test.ts server/convexProxy.test.ts
 Test Files  4 passed (4)
      Tests  26 passed (26)
```

Red with only `fetchSkillOgMeta.ts` and `fetchPluginOgMeta.ts` reverted to `546b13e6`, new tests kept:

```console
 FAIL  server/og/fetchSkillOgMeta.test.ts > aborts when the public skill API returns headers then stalls the JSON body
 FAIL  server/og/fetchPluginOgMeta.test.ts > aborts when the public package API returns headers then stalls the JSON body
Error: Test timed out in 4000ms.
```

So the new cases fail for the stated reason instead of passing vacuously.

**Archive metric timers (`e6e70b09`).** The wall-clock dependency is gone and the guard still bites: with `server/convexProxy.ts` reverted to `4117154e` (the awaited `recordMetric()`), `does not wait for download metrics before emitting zip bytes` fails with `zip first byte blocked by download metric` at 14ms. Ten consecutive runs on `d20f8fea` pass.

One observation I am not asking you to act on: `recordMetric` moved from awaited to fire-and-forget, so a manifest with one small entry can finish the zip before the POST lands, and the metric is dropped rather than merely late. `specs/download-metering.md` now calls it best-effort, so this reads as deliberate; noting it only because the previous code did guarantee delivery.

Nothing outstanding from me.

PR #3474, opened 2026-08-17
fix(api): skip duplicate skill export paths

Patrick-Erichsen:
## Summary

- reject duplicate legacy archive paths while planning a skill export
- keep the first file, avoid fetching the duplicate Blob, and record the skipped path in `_errors.json`
- preserve the successful export response and `X-Export-Errors` signal

## Production evidence

Axiom window `2026-08-14T15:01:51.298Z` to `2026-08-17T15:08:28.591Z` recorded 640 duplicate ZIP-path failures, up from 267 in the equal preceding baseline. One observed collision was `skills/humanizer/_skillhub_meta.json`.

## Validation

- focused TDD regression: red before the fix, green after
- `convex/httpApiV1.handlers.test.ts`: 431/431 passed
- `bunx tsc --noEmit`: passed
- `bun run ci:static`: passed
- `bun run ci:unit`: 6,092 passed, 3 skipped; one unrelated UI timeout reran green
- `bun run ci:types-build`: passed
- autoreview: clean, no accepted/actionable findings

## Residual risk

This intentionally does not address the separate export OOM class (1,886 failures in the current window). That requires extending the Nitro signed-manifest streaming architecture to bulk exports and belongs in a separate repair.

### DIFF
--- a/convex/httpApiV1.handlers.test.ts
+++ b/convex/httpApiV1.handlers.test.ts
@@ -942,7 +942,7 @@ describe("httpApiV1 handlers", () => {
     expect(Object.keys(zipEntries).some((path) => path.endsWith("/SKILL.md"))).toBe(false);
   });
 
-  it("skills export logs generation failure context", async () => {
+  it("skills export skips duplicate legacy file paths without failing the page", async () => {
     vi.mocked(requireApiTokenUser).mockResolvedValue({
       userId: "users:actor",
       user: { _id: "users:actor", role: "user" },
@@ -951,8 +951,6 @@ describe("httpApiV1 handlers", () => {
       userId: "users:actor",
       user: { _id: "users:actor", role: "user" },
     } as never);
-    const consoleError = vi.spyOn(console, "error").mockImplementation(() => undefined);
-
     const runQuery = vi.fn(async (_query: unknown, args: Record<string, unknown>) => {
       if ("startDate" in args) {
         return {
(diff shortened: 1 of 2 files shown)


Yigtwxx (commented):
Verified this locally against `origin/main` (`4117154e`). The dedup works and the new test is coupled to it: keeping the PR's test file and restoring only `convex/httpApiV1/skillsV1.ts` from main gives `1 failed | 8 passed` with `Duplicate ZIP path detected: "alice/demo/skills/humanizer/_skillhub_meta.json"` thrown at `convex/lib/skillZip.ts:306`; at the PR head the same `-t "skills export"` selection is `9 passed`.

Two notes, neither blocking the skills fix.

**1. The rewritten test drops the only coverage of the export failure log.**

The case being replaced was the sole place asserting the `skills_export_failed` context produced by `logSkillsExportFailure` (`convex/httpApiV1/skillsV1.ts:3754`), including `expect(JSON.stringify(consoleError.mock.calls)).not.toContain("user-token")` (main, `convex/httpApiV1.handlers.test.ts:1022`). After this PR, `skills_export_failed` appears nowhere in the test suite, so the guarantee that a bearer token never reaches the failure log is unasserted. The logging path itself is unchanged and still reachable from all three call sites (`:3840`, `:3863`, `:4133`) - only the coverage went away. A small standalone case that forces any `build_zip` failure would keep it, since the duplicate-path route into it is exactly what this PR closes.

**2. `exportPluginsV1Handler` has the same failure mode and is untouched.**

`convex/httpApiV1/packagesV1.ts:2255-2338` is the same plan/assemble shape with no dedup. Its meta entry sits under a separate `__clawhub_export/` prefix (`pluginExportMetaPath`, `:2081`), so only file-vs-file collisions apply there - but two files sharing a path in one release still fail the whole page. Reproduced on this PR's head with a temporary case added to the same test file:

```
FAIL convex/httpApiV1.handlers.test.ts > httpApiV1 handlers > PROOF plugins export duplicate release path
Error: Duplicate ZIP path detected: "bundle-plugin/demo-bundle/plugins/humanizer/_skillhub_meta.json"
  at buildMergedExportZip convex/lib/skillZip.ts:306:13
  at Object.exportPluginsV1Handler convex/httpApiV1/packagesV1.ts:2390:22
```

`convex/lib/clawpack.ts:138` already rejects duplicate paths, but only for ClawPack parsing; `artifactKind` is optional in `convex/schema.ts:1826` and `legacy-zip` releases predate it, which is the same legacy window this PR addresses on the skills side. Happy to send the mirror change as a separate PR once this lands, if you would rather keep this one scoped to skills.

PR #3479, opened 2026-08-18
fix: allow republishing hard-deleted packages

Patrick-Erichsen:
## Summary

- ignore package publish attempts whose referenced package or release was hard-deleted
- retain the audit attempt while allowing the deleted package version to be published again
- add a regression test for the exact orphaned-attempt lookup

## Root cause

Hard-deleting a publisher removes its packages and releases but retains publish-attempt audit rows. The version-conflict lookup treated those orphaned rows as live reservations, so a recreated publisher could not republish the same version.

## Validation

- `bunx vitest run convex/publishAttempts.test.ts`
- `bun run ci:static`
- `bun run ci:unit`
- `bun run ci:types-build`
- local Convex function push with `bunx convex dev --once`
- autoreview: clean, no actionable findings

### DIFF
--- a/convex/publishAttempts.test.ts
+++ b/convex/publishAttempts.test.ts
@@ -71,7 +71,10 @@ const findExistingPublishAttemptForArtifactHandler = (
   }
 )._handler;
 
-function makeAttemptLookupCtx(attempts: Array<Record<string, unknown>>) {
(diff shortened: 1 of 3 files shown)


Patrick-Erichsen:
@clawsweeper re-review

PR #3508, opened 2026-08-22
fix: repair orphaned pending skill versions after publish

### LINKED ISSUE #3349: Skill publish returns success and reserves version, but version is absent from latest/version index
## Summary

`clawhub skill publish` reported success and stored the complete version payload,
but the version was never projected onto the skill record. Its files are
retrievable by explicit version parameter, yet the version is absent from the
version index, the `latest` pointer, the tag pointers and the display name.

The result is a version that simultaneously **exists** (a repeat publish is
rejected as a duplicate) and **does not exist** (every metadata read path returns
`404 Version not found`). Because the duplicate check blocks re-publishing the
same version, a publisher cannot repair this from the client side.

## Affected record

| Field | Value |
|---|---|
| Skill slug | `oraclenet-mesh` |
| Owner handle | `tooloracle` |
| Owner display name | `ToolOracle` |
| Target version | `3.0.0` |
| Version ID returned by publish | `k97ddgverr3426bqww55gg184d8bjtad` |
| Previously visible version (still shown as latest) | `2.0.1` |
| Download baseline | `576` |
| Install baseline | `16` |
| Visible version count | `6` |
| Source repository | `ToolOracle/oraclenet-mesh-skill` |
| Source merge commit | `41d44aa8d67669fc9ac7f6e9be010b90b693648b` |
| CLI | `clawhub` v0.22.0 (`743fa3ab`) |
| Observed | 2026-07-31, publish ~10:03Z, diagnosis 10:22Z–10:24Z |

## Publish response

```json
{
  "ok": true,
  "status": "published",
  "slug": "oraclenet-mesh",
  "displayName": "OracleNet — Live Data & Capability Router for AI Agents",
  "version": "3.0.0",
  "latestVersion": "2.0.1",
  "fileCount": 6,
  "fingerprint": "53cf9c16901cd7a35a6c4b02976d6f3bace940f4a2656c07a352da9843858bfa",
  "versionId": "k97ddgverr3426bqww55gg184d8bjtad"
}
```

Exit code 0. No warning, no partial-failure indication.

## Technical evidence

Every read path below was requested twice — once authenticated as the owning
publisher, once anonymously. **Results were identical**, so this is not a
moderation gate and not an owner-only staging state. All requests used
cache-busting parameters; responses carry `cache-control: no-store`,
`cf-cache-status: DYNAMIC`, `age: 0`.

1. **Publish reports success** — `ok: true`, `status: "published"`,
   `fileCount: 6`, and a `versionId`, as shown above.

2. **A repeat publish is rejected as a duplicate:**
   ```
   Version 3.0.0 already exists. Increment the version number and try again.
   ```
   The uniqueness record therefore exists server-side.

3. **`GET /api/v1/skills/oraclenet-mesh/versions/3.0.0` → `404 Version not found`**
   (authenticated and anonymous, 10:23:00Z).
   `…/versions/3.0.0/files` likewise returns `404 Not found`.

4. **The version list still contains only six versions.**
   `GET /api/v1/skills/oraclenet-mesh/versions` (also with `?limit=200`) returns
   `2.0.1, 2.0.0, 1.2.0, 1.1.1, 1.1.0, 1.0.0`, `nextCursor: null`.
   `stats.versions` is still `6`.

5. **`latest` still resolves to `2.0.1`.** `GET /api/v1/skills/oraclenet-mesh`
   reports `latestVersion.version: "2.0.1"`.

6. **All tag pointers still reference older versions.** Ten topic tags and the
   `latest` tag remain bound to `2.0.1`; six further tags remain on `1.x`. None
   moved. The `displayName` also remains `Oraclenet Mesh`, although the publish
   passed `--name "OracleNet — Live Data & Capability Router for AI Agents"` and
   echoed that value back in its own response.

7. **Moderation is `clean`.** `GET /api/v1/skills/oraclenet-mesh/moderation`
   returns `verdict: "clean"`, engine `v2.4.24`.

8. **`isSuspicious` is `false`.**

9. **`isMalwareBlocked` is `false`.**

10. **`reasonCodes` is empty** (`[]`).

11. **No pending or scan record exists for 3.0.0.**
    `GET /api/v1/skills/oraclenet-mesh/scan?version=3.0.0` returns
    `404 Version not found` — not a pending status. The general `/scan` endpoint
    still describes `version: "2.0.1"`. No `pending`, `scanning`, `held`,
    `hidden`, `blocked`, `quarantined`, `in review` or `unlisted` flag appears in
    any response.

12. **All six files of version 3.0.0 are retrievable** via
    `GET /api/v1/skills/oraclenet-mesh/file?version=3.0.0&path=…`:

    | Path | HTTP | Bytes |
    |---|---|---|
    | `SKILL.md` | 200 | 14491 |
    | `scripts/route.py` | 200 | 9363 |
    | `scripts/smoke_test.py` | 200 | 11107 |
    | `references/route-recipes.md` | 200 | 8388 |
    | `references/verification.md` | 200 | 6573 |
    | `references/x402-safety.md` | 200 | 6464 |

    `GET /api/v1/download?slug=oraclenet-mesh&version=3.0.0` also returns a payload.

13. **The six files are SHA-256 identical** to the corresponding files at
    `ToolOracle/oraclenet-mesh-skill@41d44aa8`. The stored content is exactly
    what was submitted.

14. **Control probe.** `scripts/route.py` exists only in 3.0.0, not in 2.0.1.
    The same endpoint returns **404** at `version=2.0.1` and **HTTP 200** at
    `version=3.0.0`. The `version` parameter is therefore genuinely honoured —
    these are not 2.0.1 bytes served under a 3.0.0 label, and the 200s in point
    12 are not a silent fallback to latest.

15. **Conclusion: the content layer is present and correct, the index projection
    is missing.** The write that stores version documents and file blobs
    succeeded; the write that updates the skill document — version index,
    `latestVersionId`, tag pointers, display name, scan queue — did not.

## Impact

- A publish can report success while the public listing stays on the previous
  version. `clawhub install <slug>` continues to serve the older version with no
  indication that a newer one was accepted.
- The failure is silent: exit code 0, `"status": "published"`, and a version ID.
  Nothing in the output distinguishes this from a healthy publish.
- The version is unrecoverable by the publisher, because the duplicate guard and
  the index write disagree about whether it exists.

## Expected behaviour

A publish should be atomic across the content write and the index projection. If
the index update cannot complete, the version reservation should not be retained
— or the publish should return a non-success status so the client can retry.

## Requested operator action

For the affected record, please:

- link the existing `3.0.0` version document (`k97ddgverr3426bqww55gg184d8bjtad`)
  to the existing `oraclenet-mesh` skill;
- restore/update the version index so `3.0.0` is listed (`stats.versions` 6 → 7);
- update `latestVersionId` to point at that version;
- include `3.0.0` in the version list returned by `/versions`;
- move the `latest` tag pointer to `3.0.0`;
- update the remaining relevant tags published with this version —
  `live-data`, `capability-routing`, `mcp`, `agent-infrastructure`, `research`,
  `blockchain`, `verification`, `x402`, `autonomous-agents`, `external-data`;
- **preserve the existing 576 downloads and 16 installs** — these were unaffected
  throughout and must not be reset by the reconcile;
- **do not delete or recreate the skill**;
- **no 3.0.1 workaround should be required** — bumping the version would strand
  an unreachable `3.0.0` in the registry permanently.

Alternatively, releasing the `3.0.0` reservation so the identical payload can be
re-published cleanly would also resolve it.

## Related historical reports

- #137 — Published skill not appearing in search/explore/inspect (closed 2026-04-30)
- #3288 — Bundle plugin releases are created but never become visible (closed 2026-07-30)

Both describe published artifacts failing to become visible, so they are likely
the same class of defect. Two differences are worth noting:

- #3288 concerns the **`bundle-plugin`** family, where the reporter saw
  `package scan: pending` and `latest release: none`. This report concerns the
  **skill** publish path, where moderation is already `clean` and **no scan
  record exists for the version at all**. If #3288's fix covered only the plugin
  release path, the skill path may still carry the same root cause.
- This incident occurred on 2026-07-31, one day after #3288 was closed.

This report is also **more narrowly localised** than either: the new version's
content is fully stored and retrievable by explicit version parameter
(points 12–14), while only the version metadata and index projection are missing.
That narrows the fault to the skill-document update step rather than to upload,
storage, moderation or search indexing — and gives it a specific signature: a
version that satisfies the duplicate check but is absent from the version index
while its files remain fetchable.

## Reproduction notes

Not reproduced deliberately — observed once on a live publish, likely a partial
failure between the version/file write and the skill-document update.

lechiffre-cfo:
Fixes #3349.

This is a clean rebase of the existing orphaned pending skill-version repair line onto current `openclaw/main` (`845c6d3bdb1a`), with the rebase conflicts resolved and the focused proof rerun locally.

The bug: staged skill publish can persist the pending `skillVersion` and files, then fail before projecting that version onto the parent skill document (`latestVersionId`, version list, tags, search state). The publisher then sees a duplicate reservation on republish, while public metadata still returns 404/old latest.

This branch keeps the existing intended fix shape:

- owner-visible diagnostics for pending, blocked, and terminal failed publication states without leaking hidden versions to public callers;
- capped repeated finalization failures so stuck finalizers become operator-repairable instead of looping forever;
- admin-gated orphan repair and dry-run sweep;
- atomic repair path that publishes, closes/clears the associated attempt where safe, clears staged `pendingPublication`, schedules security followups, and writes a repair audit row in the same mutation boundary;
- tests for public non-leakage, org publisher ownership, terminal failed attempts, active-attempt refusal, existing stats preservation, and hidden first-publish cases.

Local proof after rebase:

```text
npx --yes bun x vitest run convex/httpApiV1.handlers.test.ts -t "3349"
# 1 file passed, 10 tests passed

npx --yes bun x vitest run convex/publishAttempts.test.ts -t "findActiveSkillPublishAttempt|3349|orphaned|finalizationFailure|scanner retries|terminalizes a staged release|terminalizes permanent staged OpenClaw"
# 1 file passed, 23 tests passed

npx --yes bun x vitest run convex/maintenance.test.ts -t "3349|orphan|repair"
# 1 file passed, 28 tests passed

npx --yes bun x vitest run convex/publishAttempts.runtime.test.ts convex/skills.pendingProjection.test.ts
# 2 files passed, 5 tests passed
```

I did not run the full CI suite locally.


lechiffre-cfo:
Follow-up pushed at `72915f75`.

Changes after ClawSweeper review:

- rebased PR branch onto current `origin/main` / `fb0ef1d21eab`;
- restored the current-main publish webhook contract in `scheduleSkillPublishFollowups`: `await schedulePublishWebhook(...)` inside `try/catch`, so webhook lookup/scheduling remains ordered but best-effort;
- PR API now reports `mergeable: true`, `mergeable_state: unstable`.

Local validation on the rebased branch:

```text
npx --yes bun x vitest run convex/lib/skillPublish.test.ts -t "webhook"
# 1 file passed, 4 tests passed

npx --yes bun x vitest run convex/publishAttempts.test.ts -t "findActiveSkillPublishAttempt|3349|orphaned|finalizationFailure|scanner retries|terminalizes a staged release|terminalizes permanent staged OpenClaw|targeted"
# 1 file passed, 29 tests passed

npx --yes bun x vitest run convex/httpApiV1.handlers.test.ts -t "3349"
# 1 file passed, 10 tests passed

npx --yes bun x vitest run convex/maintenance.test.ts -t "3349|orphan|repair"
# 1 file passed, 28 tests passed

npx --yes bun x vitest run convex/publishAttempts.runtime.test.ts convex/skills.pendingProjection.test.ts
# 2 files passed, 7 tests passed

npx --yes bun x tsc --noEmit
# passed
```

Real Convex proof status: I attempted `CONVEX_LOCAL_BACKEND_STARTUP_TIMEOUT_SECS=120 npx --yes bun x convex dev --once --typecheck=disable` in this worker, but the local backend did not start on port 3210 within 120s. I am not treating that as product proof. The remaining proof ask still needs a dev/preview Convex environment that can actually start and expose the HTTP API.

@clawsweeper re-review

PR #3510, opened 2026-08-22
fix(publish): ignore duplicate leased recovery dispatches

vincentkoc:
## What Problem This Solves

Fixes an issue where workers recovering an exact staged publish could treat an active lease as a claim failure or fall through to another claim path when `preferRetry` was enabled.

## Why This Change Was Made

Targeted pending-check and ready-finalization claims now return an internal `active_claim` outcome for a live foreign lease, and the public worker action maps that outcome to `null`. Exact `attemptId` requests use status-aware claim ordering, while untargeted batches retain expired-retry preference. The durable security specification records the same no-op and lease-preservation invariant.

## User Impact

Duplicate recovery dispatches now report no available work without disturbing the active worker lease. Mismatched kind, slug, or version filters still fail, and the same claim can still recover its leased attempt.

## Evidence

- Real `convex-test` action calls prove exact pending and ready attempts with `preferRetry: true` return `null` while preserving the foreign claim ID, claim expiry, and complete stored attempt row.
- Focused unit tests prove the public action does not fall through to another claim path.
- The production and test bytes at this head are unchanged from `ce859551`, where these checks passed:
  - `bun run test convex/publishAttempts.test.ts` - 43 passed.
  - `NODE_OPTIONS=--localstorage-file=<tmp> bun run test convex/publishAttempts.runtime.test.ts` - 5 passed.
  - `bun run test scripts/security/run-prepublication-worker.test.ts -t "reserves one batch claim for expired retries, including filtered batches"` - 1 passed.
  - `bun run ci:static` - peer, audit, generated text, formatting, type-aware lint, dead-code, and workflow pin checks passed.
  - `bunx tsc --noEmit` - passed.
- The specification-only exact-head follow-up passed `git diff --check` and `bun run format:check -- specs/security-moderation.md`.

## Limits

- The PR changes exactly four files: production `+43/-42`, tests `+198/-27`, and specification `+4/-0`.
- No release-note or changelog changes remain.
- No schema, generated API, public response shape, or worker implementation changes.
- Untargeted `preferRetry` queue behavior is unchanged.
- The worker regression test file remains unchanged; its existing null-claim test is included above as integration proof.

## Contribution Note

AI-assisted implementation under maintainer-directed scope; the behavior and focused validation above were reviewed before submission.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,6 @@
 
 ### Fixes
 
-- Workers: treat targeted pre-publication attempts with active leases as no work instead of retrying another claim path.
 - GitHub Actions/CLI/API: require v2 authorization for every OpenClaw OIDC publish, bind large-artifact upload tickets and package mutations to the exact authorization transaction, consume scoped capabilities with non-public staging, then require an immutable successful parent attempt and atomic durable authorization revalidation before final publication.
 - GitHub Actions: revalidate versioned trusted tooling workflow identity immediately before package publication, including exact protected lightweight tags, so approval waits cannot authorize moved tooling.
 - CI: authenticate the scheduled project-skill updater's pull request mutations with the Barnacle GitHub App while retaining the workflow token for provenance lookup and branch publication.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ### Fixes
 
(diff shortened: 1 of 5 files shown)


vincentkoc:
@clawsweeper re-review

vincentkoc:
@clawsweeper re-review

vincentkoc:
@clawsweeper re-review

PR #3512, opened 2026-08-22
fix: surface a retryable error when the skills catalog fails to load

### LINKED ISSUE #3473: Skills list: first-page fetch failure leaves no error signal, only a bare "Load more" card
Follow-up from #3457 (review by @Yigtwxx,
https://github.com/openclaw/clawhub/pull/3457#pullrequestreview-4919932147).

After a first-page fetch failure on the New, Featured, or Official catalog tabs, the page shows
no failure signal at all. #3457 moved the failure state from `done` to `idle` so the tab stays
retryable, which removed the false `No skills found` — but nothing replaced it:

- The empty state is gated on `listDoneLoading` (`src/routes/skills/index.tsx`), computed as
  `!isLoadingSkills && !canLoadMore && !isLoadingMore`. With status `idle`, `canLoadMore` is
  true, so `listDoneLoading` is false and the empty block is skipped.
- The result list is empty, so nothing renders there either.
- What remains on screen is a bare `Load more` card.

A user who hits a 503 sees an empty page whose only hint that anything went wrong is a button.

`listDoneLoading` cannot distinguish "nothing matched" from "the request failed", because
`canLoadMore` is true in both. The likely shape of a fix is a distinct error state in the browse
model instead of folding the failure into `idle` — deliberately out of scope for #3457, which is
a retryability fix and should not absorb a UI decision.

Filing so it doesn't die in the PR thread. Happy to take it if the maintainers want it.

Yigtwxx:
Closes #3473

## What Problem This Solves

Fixes an issue where users browsing the Skills catalog would see an empty page whose only
content is a bare "Load more" card when the first page request fails on the New, Featured,
or Official tabs.

Nothing on the page says the request failed. The result list is empty, and the "No skills
found" empty state is gated on `listDoneLoading` (`!isLoadingSkills && !canLoadMore &&
!isLoadingMore`), which stays false while the tab is retryable. A user who hits a 503 gets a
button and no explanation.

## Why This Change Was Made

#3457 made a failed first page retryable by leaving the list status at `idle`, deliberately
leaving the failure presentation to a follow-up. `listDoneLoading` cannot tell "nothing
matched" from "the request failed", because `canLoadMore` is true in both cases, so the fix
is a distinct state rather than more conditions on the existing ones.

The browse model's `ListStatus` gains an `error` member, set only for a cursor-less failure
on the non-Trending tabs. The results component branches on it and renders an inline failure
message with a `Try again` action that replays the first page, matching the failure copy
already used by the home listing surface and the retry affordance already used by the plugin
versions panel.

`Try again` is single-flight. Two activations can reach the retry callback before React
rerenders the failure away, and both would replay the first page under the same fetch
generation, leaving the later-to-settle reply free to overwrite the earlier one and restore
the error after a successful recovery. An in-flight ref guards the dispatch and is released
once the request settles, mirroring the guard the adjacent `Load more` path already uses.

Boundaries kept intact:

- Canonical Trending keeps its own `canonicalTrendingUnavailable` / `trendingState =
  "unavailable"` path; it never reaches the new state.
- A failure on a later page still records the retry cursor and stays `idle`, so pagination
  resumes through `Load more` exactly as before.
- The empty-page auto-advance path (a filtered scan that advanced a cursor before failing)
  also keeps its existing `idle` behavior, since it has a real cursor to resume from.
- Search results are unaffected; `listFailed` is false whenever a query is active.

## User Impact

When a catalog request fails, the Skills page now says so and offers a `Try again` control
that reloads the tab in place, instead of showing an unexplained "Load more" button. The
failure message is announced to assistive technology (`role="alert"`). Recovery from a
transient failure takes one click and no longer depends on the user guessing that "Load
more" is really a retry.

## Evidence

Browser proof, local dev server on `http://localhost:3000` against the public read Convex
deployment (`wry-manatee-359`), failure forced with DevTools offline emulation on the
`New` tab:

**Before (current `main`) - the reported state, a bare "Load more" card:**

before

**After - explicit failure with a retry action:**

after

**After clicking `Try again` with the network restored - the New tab loads normally:**

after retry

Tests. The existing parameterized first-page-failure test in
`src/__tests__/skills-index.test.tsx` is updated: it still asserts that `No skills found`
never appears, and now asserts the failure message is present, the bare `Load more` card is
gone, and `Try again` replays the first page with no cursor and recovers. Three tests are
added: a repeated failure keeps the failure state (it must not fall back to the false empty
state), a canonical Trending failure never renders the new state, and two `Try again`
activations that land before the first reply settles dispatch a single first-page request
and leave the recovered list intact.

```
$ bunx vitest run src/__tests__/skills-index.test.tsx \
    src/__tests__/skills-index-load-more.test.tsx \
    src/__tests__/skills-index.claw591.test.tsx
Test Files  3 passed (3)
     Tests  77 passed (77)

$ bunx tsc --noEmit
(clean)

$ bun run lint
(clean)

$ bunx oxfmt --check <touched files>
All matched files use the correct format.
```

`bun run ci:unit` on this branch: 16 files / 27 tests fail. All of them are Windows-local
environment failures also present on `main` and untouched by this diff (worker/CLI suites
that shell out to `bun`, a Mermaid render timeout, and a number-format assertion); none are
in `src/routes/skills/**` or `src/__tests__/skills-index*`. `bun run format:check` fails on
`CLAUDE.md` and `.agents/skills/autoreview/CLAUDE.md`, which is pre-existing on `main` and
not part of this change.

`$autoreview` could not run in this environment: it requires TruffleHog, which is not
installed here and which it refuses to auto-install. The diff was reviewed manually against
the surrounding code paths instead.

### DIFF
--- a/src/__tests__/skills-index.test.tsx
+++ b/src/__tests__/skills-index.test.tsx
@@ -1087,6 +1087,51 @@ describe("SkillsIndex", () => {
     },
   );
 
+  it("dispatches one first-page request when Try again is activated twice before it settles", async () => {
+    vi.stubGlobal("IntersectionObserver", undefined);
+    const consoleErrorSpy = vi.spyOn(console, "error").mockImplementation(() => {});
+    try {
+      let settleRetry: (page: unknown) => void = () => {};
+      convexHttpMock.query
+        .mockRejectedValueOnce(new Error("temporary failure"))
+        .mockImplementationOnce(
+          () =>
+            new Promise((resolve) => {
+              settleRetry = resolve;
+            }),
+        )
+        .mockRejectedValue(new Error("duplicate retry dispatched"));
+
+      render(<SkillsIndex />);
+      await act(async () => {});
+
+      const retryButton = screen.getByRole("button", { name: "Try again" });
+      await act(async () => {
+        fireEvent.click(retryButton);
+        fireEvent.click(retryButton);
+      });
+
+      // Both activations land before React rerenders, so only the guard can keep the second
+      // one from starting a rival request that outlives the first.
+      expect(convexHttpMock.query).toHaveBeenCalledTimes(2);
+
+      await act(async () => {
+        settleRetry({
+          page: [makeListResult("recovered-skill", "Recovered Skill")],
+          hasMore: false,
+          nextCursor: null,
+        });
+      });
+
+      expect(convexHttpMock.query).toHaveBeenCalledTimes(2);
+      expect(screen.getByText("Recovered Skill")).toBeTruthy();
+      expect(screen.queryByRole("alert")).toBeNull();
+      expect(screen.queryByRole("button", { name: "Try again" })).toBeNull();
+    } finally {
+      consoleErrorSpy.mockRestore();
+    }
+  });
+
   it("keeps the first-page failure state when the retry fails again", async () => {
     vi.stubGlobal("IntersectionObserver", undefined);
     const consoleErrorSpy = vi.spyOn(console, "error").mockImplementation(() => {});
--- a/src/__tests__/skills-index.test.tsx
+++ b/src/__tests__/skills-index.test.tsx
@@ -1051,7 +1051,7 @@ describe("SkillsIndex", () => {
   });
 
   it.each(["new", "featured", "official"] as const)(
-    "keeps the %s first page retryable after a temporary fetch failure",
+    "surfaces a retryable failure on the %s first page after a temporary fetch failure",
     async (tab) => {
       searchMock = { tab };
       vi.stubGlobal("IntersectionObserver", undefined);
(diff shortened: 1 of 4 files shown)


Yigtwxx:
Status check after a week, in case this looks like it is waiting on something from me.

ClawSweeper's latest review of the current head (`4f39dc34`) reports no findings and no
contributor-facing blocker, and CI is green. The one red check is the Vercel deployment, which is
waiting on team authorization for a fork commit rather than on anything in the diff. The branch is
still mergeable against `main`; it is 24 commits behind but conflict-free, and I am happy to rebase
it if you would rather have it caught up before review.

One update to what I wrote on #3473 when I opened this. I said the `clawsweeper:needs-product-decision`
label was still the gate on the copy and the affordance. That label has been added and removed a few
times since, and it is currently off the issue (removed Aug 28); the issue now carries
`clawsweeper:linked-pr-open` and `clawsweeper:no-new-fix-pr`. So as far as I can tell the only thing
left here is an ordinary maintainer review of the error copy and the `Try again` affordance.

@pacocartones the offer from #3473 still stands: if you had already started on this, say so and I
will close mine.

Yigtwxx:
Rebased onto current `main` (`2d029cf8`) and force-pushed; head is now `a4b40e6e`.

This is for the one open item in ClawSweeper's latest review, which asks for focused Skills
regression validation on the exact current-main-plus-head merge rather than on the stale
test-merge commit.

The rebase was conflict-free, and the diff is byte-for-byte what it was before: 4 files,
+136/-6. That is expected rather than lucky - none of the four files this PR touches has
changed upstream since the old base `fb0ef1d2`:

```
$ git log fb0ef1d2..origin/main -- src/routes/skills/-SkillsResults.tsx \
    src/routes/skills/-useSkillsBrowseModel.ts src/routes/skills/index.tsx \
    src/__tests__/skills-index.test.tsx
(no output)
```

Validation on the rebased head:

- `bunx tsc --noEmit` - clean.
- `bunx vitest run` over every test file that touches the Skills browse surface -
  7 files, 137 tests, all passing: `skills-index`, `skills-index-load-more`,
  `skills-index.claw591`, `skills-route-default-sort`, `skills-sh-route`,
  `skills-publish-route`, `ui-design-contract`.

The browser proof in the description still stands - the production diff is unchanged, so the
before/after/recovery screenshots describe the same code paths.

One correction to my comment from Aug 31. I said the `clawsweeper:needs-product-decision`
label was off the issue. That was true when I wrote it, but the label has gone on and off
#3473 several times since (removed Aug 28, back Sep 1, and the issue rating has flapped
between reviews the same way), so treating its momentary state as the gate was not a useful
thing to report. Ignore that paragraph; the substance is unchanged - what is left here is an
ordinary maintainer review of the error copy and the `Try again` affordance.

Patrick-Erichsen:
Refreshed this PR onto main `3ab0037e1b` with merge commit `e47cad0436608980c8ed7783d0a82170dd796f4c`. The four-file +136/-6 scope is preserved. Conflict resolution retains main’s manual-search error handling alongside this PR’s first-page list error and retry state.

Before/after proof remains applicable and was inspected directly: [bare Load more](https://raw.githubusercontent.com/Yigtwxx/clawhub/proof-3473-skills-first-page-error-state/01-before-main-bare-load-more.png) → [explicit error and Try again](https://raw.githubusercontent.com/Yigtwxx/clawhub/proof-3473-skills-first-page-error-state/02-after-error-state.png) → [recovered listing](https://raw.githubusercontent.com/Yigtwxx/clawhub/proof-3473-skills-first-page-error-state/03-after-retry-recovered.png).

Refreshed validation:

- 102 focused catalog tests and 6,917 full unit tests pass (3 skipped).
- `ci:static`, `ci:types-build`, and both explicit schema/CLI `tsc --noEmit` checks pass.
- Structured autoreview: no actionable findings.
- The formerly failing malicious-publication browser scenario passes against the real isolated local Convex/app on its first attempt, 40.2s total, runner exit 0. Node 24.19.0 / Bun 1.4.0; no assertions or timeouts were relaxed.

The historical failing job recorded one-second Convex timeouts and a hydration failure. This refresh includes main’s later startup fixes and passes the unchanged scenario locally. Fresh hosted CI is running on the new head; the fork’s Vercel authorization result is separate from test validation.

PR #3524, opened 2026-08-25
fix(sync): avoid duplicate skills.sh batches after timeouts

vincentkoc:
## What Problem This Solves

Fixes an issue where the skills.sh synchronization job could retry an exact cursor while the original timed-out batch still held or renewed its durable lease, and where HTTP 408 responses were treated inconsistently across sibling source GET paths.

## Why This Change Was Made

The durable run projection now exposes only the batch lease expiry, never the lease token. After an ambiguous step timeout, the runner polls that exact durable run through live or renewed leases, adopts committed cursor progress, and retries the exact cursor only after authoritative expiry or release. The four bounded source GET loops now share the same retryable status policy for HTTP 408, 429, and 5xx while preserving the existing long-429 durable handoff and four-attempt cap.

No schema, config, migration, or deployment changes are included.

## User Impact

Operators can expect timed-out skills.sh mirror steps to avoid racing a still-running durable batch, and transient HTTP 408 responses to recover consistently without weakening bounded retry behavior.

## Evidence

- Focused regression suite: 3 files passed, 119 tests passed.
- `bun run ci:static`: passed.
- `bun run ci:unit`: 470 files passed, 1 skipped; 6,239 tests passed, 3 skipped.
- `bun run ci:types-build`: passed.
- Convex dry-run codegen/typecheck passed under Node 24.16.0 against the local anonymous deployment.
- `git diff --check`: passed.
- Added lease cases: expired, live then released, renewed then advanced, already advanced, permanently live bound, missing expiry, and malformed expiry.
- Added HTTP 408 recovery and exhaustion coverage across authenticated API, identity page, proof metadata, and controlled pinned-source GET loops.
- Structured autoreview's TruffleHog scan passed. The model review could not start because the required standalone Codex CLI is not installed in the execution environment; the diff received direct source review instead.

Production delta: +88/-18 lines. The added production code establishes the durable reconciliation owner boundary and central retry classification; tests are +534/-187 lines.

### DIFF
--- a/convex/skillsShMirror.test.ts
+++ b/convex/skillsShMirror.test.ts
@@ -1086,19 +1086,26 @@ describe("skills.sh external mirror", () => {
     await configure(t);
     const { runId } = await startRun(t, "snapshot:lease", 1);
 
-    await expect(
-      t.mutation(mirrorLeaseRefs.claimBatchLeaseInternal, {
-        runId,
-        page: 0,
-        offset: 0,
-        leaseToken: "lease:first",
-      }),
-    ).resolves.toMatchObject({
+    const claimed = await t.mutation(mirrorLeaseRefs.claimBatchLeaseInternal, {
+      runId,
+      page: 0,
+      offset: 0,
+      leaseToken: "lease:first",
+    });
+    expect(claimed).toMatchObject({
       runId,
       page: 0,
       offset: 0,
       leaseExpiresAt: expect.any(Number),
     });
+    await expect(t.query(internal.skillsShMirror.getRunInternal, { runId })).resolves.toMatchObject(
+      {
+        batchLeaseExpiresAt: claimed.leaseExpiresAt,
+      },
+    );
+    expect(await t.query(internal.skillsShMirror.getRunInternal, { runId })).not.toHaveProperty(
+      "batchLeaseToken",
+    );
     await expect(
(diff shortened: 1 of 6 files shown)


vincentkoc:
@clawsweeper re-review

vincentkoc:
## Maintainer merge-risk disposition

Exact reviewed head: `1edaa148b394dffcf6da8e3a85889db107617309`.

- **Bounded recovery accepted:** the 5-second poll interval, 360-poll ceiling, and 30-minute maximum wait are the intended availability tradeoff. Waiting on the authoritative live lease is preferable to racing the original timed-out batch with a duplicate exact-cursor request.
- **Scheduled-sync trace reviewed:** `scripts/skills-sh-catalog/sync.test.ts` passed 20/20 with named coverage for live lease release, lease renewal followed by cursor advance, the 360-poll/1,800,000 ms fail-closed bound, missing/malformed lease state, and an already-committed authoritative cursor.
- **Structured security proof satisfied:** both exact-head `Security Gate: Secret Scanning` / TruffleHog checks passed. The unavailable local helper therefore leaves no secret-scan gap.
- **Hosted proof:** CI run `32822126357` attempt 2 passed after the single unrelated Convex-local hydration retry; all PR checks are green with zero pending or failing checks.

No backend deployment was performed or required. The ClawSweeper rank-up move and owner decision are resolved without changing the reviewed head, so no additional re-review round-trip is needed.

PR #3547, opened 2026-08-29
fix: fail closed when a skill archive blob is missing

SebTardif:
## What Problem This Solves

Public skill ZIP download skipped missing storage blobs and still returned HTTP 200. `clawhub install` treats 200 as success and can write a truncated archive plus origin lock. Sibling download paths already fail closed when a stored file is gone.

## Evidence

terminal output from the patched tree:

```console
$ bunx vitest run convex/downloads.test.ts
 Test Files  1 passed (1)
      Tests  32 passed (32)
```

A missing `ctx.storage.get` now returns 410 `Skill archive file missing from storage` and does not emit a ZIP or a success metric.

## Real behavior proof
Behavior addressed: skill archive download fails closed when any declared blob is missing.
Real environment tested: macOS, bun, patched clawhub worktree.
Exact steps or command run after this patch: bunx vitest run convex/downloads.test.ts
Evidence after fix: terminal output copied below.

```console
$ bunx vitest run convex/downloads.test.ts
 Test Files  1 passed (1)
      Tests  32 passed (32)
```

Observed result after fix: missing blob is 410, not a 200 ZIP.
What was not tested: a live production Convex blob store with a deleted file.

## Tracker

Ref #3668

That issue stays open if this PR is closed without landing on main.

### DIFF
--- a/convex/downloads.test.ts
+++ b/convex/downloads.test.ts
@@ -460,7 +460,7 @@ describe("downloads helpers", () => {
     expect(runAfter).toHaveBeenCalledTimes(1);
   });
 
-  it("streams stored file chunks, stays deterministic, and skips a Blob that vanishes", async () => {
+  it("streams stored file chunks and stays deterministic", async () => {
     const firstChunk = new Uint8Array(64 * 1024).fill(0x61);
     const secondChunk = new TextEncoder().encode("streamed body\n");
     const releaseSecondChunk = deferred<void>();
@@ -518,7 +518,6 @@ describe("downloads helpers", () => {
           files: [
             { path: "a.txt", storageId: "_storage:skill" },
             { path: "b.txt", storageId: "_storage:notes" },
-            { path: "missing.txt", storageId: "_storage:missing" },
           ],
           softDeletedAt: undefined,
         };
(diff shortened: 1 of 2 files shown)


Patrick-Erichsen:
Verified the direct download fix against real anonymous local Convex 1.44.0 (`http://127.0.0.1:3450`, HTTP `:3451`) and the actual ClawHub CLI. The fixture stores two real files across two versions, installs v1, then deletes v2's declared `script.txt` blob before downloading/updating. No mocked storage or action context.

| Observation | Main before | This patch after |
| --- | --- | --- |
| Healthy ZIP/install/update | Complete ZIP; both CLI commands exit 0 | Same |
| Missing blob | 200 partial ZIP silently omits script.txt | 410 text; no ZIP |
| CLI update to broken version | Exit 0; replaces working install and loses script.txt | Exit 1; reports missing file |
| Existing files, lock, origin metadata | Changed | Every SHA-256 unchanged |
| Success metric jobs across raw download + CLI update | 1 → 3 | 1 → 1 |

The initial metric job comes from the healthy v1 install. Temporary proof helpers were removed and the clean function set repushed locally with typechecking enabled.

Refreshed head `193dd11a76c83a913196472fa139e16ec5634603` merges main into the contributor history without rewriting it; net change remains two files, +78/-8. The proven tree is identical to `ab4f892edb0a9d01e132e6d95ad73cb59906b2aa`; only the merge commit message changed. All 52 focused tests, 6,936 unit tests, static, packages, types/build, Convex typecheck, and structured review pass. Current-head hosted checks are running.

Scope: this fixes the direct Convex ZIP path. The hosted manifest/Nitro proxy completeness follow-up remains tracked in #3668, so that issue should stay open after this PR lands.

PR #3549, opened 2026-08-30
fix: delete unpublished skill blobs when publish fails

### LINKED ISSUE #3670: Multipart skill publish leaves Convex blobs after a failed publish
## Summary

Authenticated multipart skill publish (`POST /api/v1/skills`) stores each uploaded file in Convex, then can reject the request during parse or later validation. Those stored ids are not deleted. Failed publishes leave unreferenced blobs.

This is a source-level resource leak. I did not run a live Convex publish against production.

## Verification / reproduction

1. Read `convex/httpApiV1/shared.ts` `parseMultipartPublish` (store, then `parsePublishBody`).
2. Follow the handler path that returns 400 after a successful store.
3. Confirm there is no `storage.delete` of those new ids on that failure path.
4. Compare with the sibling scan upload path in the same file, which already deletes on failure.

## Actual

- Failed multipart skill publish can leave stored files that no skill version owns.

## Expected

- Stored ids from the request are deleted unless a published (or otherwise persisted) version owns them.
- Successful publishes still retain their blobs.

## Evidence

Source: `convex/httpApiV1/shared.ts`. Proposed patch: PR #3549.

ClawSweeper also asked that blobs already committed to a version stay retained. That commit-then-keep rule belongs in the same fix.

## Security / privacy

Storage leak for an authenticated publisher. Not a claimed remote unauthenticated vuln. No payload.

## Acceptance criteria

- [ ] Request-local stored ids are deleted when parse or publish fails before a version owns them.
- [ ] Blobs already committed to a version are not deleted by request-local cleanup.
- [ ] Successful publish still attaches and retains the files.

This tracker stays open if PR #3549 is closed without landing on `main`.

SebTardif:
Failed multipart skill publishes now delete request-owned uploads, including partial stores and failures before publication. Once a pending or published version commits, the publishing helper transfers ownership immediately, preserving its files through later scheduler or compensation failures. This completes Sebastien Tardif's existing parser cleanup and fixes the post-commit deletion finding.

Fixes #3670.

Native local Convex proof: invalid metadata, license refusal and rejected insertion went from two orphan blobs to zero; partial storage failure went from one to zero. The previous PR deleted both files after a post-commit scheduler failure (download 404); the repaired version retained both files (download 200, 989 and 43 bytes). Pending success, successful compensation, failed compensation and normal publication were also exercised using actual storage/version mutations and explicit fault injection. No production data or hosted deployment was used.

Validation: 522 focused tests; 6,972 full-suite tests passed; static, full types/build, explicit schema/CLI/Convex typechecks passed; structured repair review returned no findings.

### DIFF
--- a/convex/httpApiV1.handlers.test.ts
+++ b/convex/httpApiV1.handlers.test.ts
@@ -8277,6 +8277,46 @@ describe("httpApiV1 handlers", () => {
     expect(publishVersionForUser).not.toHaveBeenCalled();
   });
 
+  it("publish multipart deletes stored blobs when owner resolution fails", async () => {
+    vi.mocked(requireApiTokenUser).mockResolvedValueOnce({
+      userId: "users:1",
+      user: { handle: "p" },
+    } as never);
+    const runMutation = vi.fn(async (_mutation: unknown, args: Record<string, unknown>) => {
+      if (isRateLimitArgs(args)) return okRate();
+      throw new Error("Publisher not found");
+    });
+    const form = new FormData();
+    form.set(
+      "payload",
+      JSON.stringify({
+        slug: "demo",
+        displayName: "Demo",
+        ownerHandle: "@missing",
+        version: "1.0.0",
+        changelog: "",
+        acceptLicenseTerms: true,
+        tags: ["latest"],
+      }),
+    );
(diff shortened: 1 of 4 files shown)


SebTardif:
@clawsweeper re-review

Issue #3575, opened 2026-09-02
API v1 skill catalog loses owner identity and has latestVersion schema drift


HwangBae:
## Summary

`GET /api/v1/skills` still exposes only `slug` as the public item identity after skills became owner-scoped. Two valid skills owned by different publishers can therefore produce the same public identity while carrying different metadata.

There is also a contract mismatch for installability/version state: the list handler and OpenAPI allow `latestVersion: null`, while the package ArkType schema (`ApiV1SkillListResponseSchema`) and strict consumers do not accept `null` for that field.

## Verification / reproduction

Use a fixed, synthetic backend fixture only; do not use production identities or payloads.

1. Create two public owners, `fixture-owner-a` and `fixture-owner-b`.
2. Give each owner a public skill with slug `shared-fixture-slug`. Set the same primary value for the selected list sort on both records (for example, the same `updatedAt` for `sort=updated`) and give them distinguishable display names/summaries.
3. Give both skills public latest versions with exact SemVer values, including one fixture such as `1.2.3+fixture.01` to prove exact version text is preserved.
4. Add enough deterministic filler skills to force at least four cursor pages with a small `limit`.
5. Add one more public skill whose public latest version is absent, so its list projection produces `latestVersion: null`.
6. Request `GET /api/v1/skills?sort=updated&limit=<small fixed limit>`, then follow every returned `nextCursor` for at least four consecutive pages.
7. Inspect the raw HTTP JSON and validate each page with `ApiV1SkillListResponseSchema` (and any strict consumer contract), preserving the distinction between an omitted field and an explicit JSON `null`.

The fixture should assert the complete public identity and metadata of every item, not only cursor uniqueness.

## Actual

- Both owner-scoped records are emitted with the same slug-only public identity even though their metadata differs.
- The list response can emit `latestVersion: null`.
- `ApiV1SkillListResponseSchema` permits an omitted `latestVersion`, but not an explicit `null`, so a valid handler response can fail strict package/consumer validation.
- The cursor implementation's stable internal tie-breaker can prevent pagination instability, but it does not resolve the collision in the identity exposed to API consumers.

## Expected

- Every list item exposes a stable `ownerHandle`; the public skill identity is `ownerHandle + slug`.
- `latestVersion` nullability has one explicit contract across the handler, OpenAPI, ArkType packages, and consumers. Either:
  - filter non-installable items from this catalog response, or
[...]

PR #3580, opened 2026-09-03
fix: preserve skill catalog owner identity and null versions

goutamadwant:
Ref #3575.

## What Problem This Solves

Fixes catalog consumers losing the distinction between two owners publishing the same skill slug, and rejecting valid entries whose public latest version is null.

## User Impact

Current catalog entries expose owner-qualified identity, and CLI explore displays `ownerHandle/slug`. Canonical responses always include an explicit nullable latest version. The CLI still accepts the earlier response shape from custom registries configured with `--registry`.

## Why This Change Was Made

The catalog query already resolves publisher identity but dropped it from the API projection. The shared schema also disagreed with the server's null response. This preserves identity through the query, HTTP response, schema, and CLI display, while keeping the named legacy custom-registry decoder compatible.

## Evidence

- Registered Convex HTTP fixtures walk ten tied-sort records across at least four cursor pages, checking complete identities, metadata, exact SemVer build text, null versions, and trending output. Removing identity projection or restoring the old null-rejecting schema reproduces two failures each.
- The built CLI runs against real loopback current and legacy registries, proving duplicate-owner output, null versions, exact version text, and legacy omissions.
- 550 focused backend/schema tests and full `bun run ci:pr` pass. On the final base: 6,918 coverage tests, package/artifact checks, static checks, TypeScript/build, and HTTP integration all pass.
- The added local-auth assertion validates catalog output from a real published skill. Two Testbox attempts hit existing one-second auth/publish/scan-claim deadlines before reaching it; the Testbox reported two CPUs despite the eight-vCPU workflow label. The normal PR browser CI has now passed this assertion on the exact updated head. No timeout or error filtering was changed.
- Independent review through P2 is scoped-clean. Maintainer follow-up and added proof were AI-assisted.


Patrick-Erichsen:
Supplemental native Convex before/after verification completed for merged head `137bf726c00ff3dcf30de83bf28ab14924714017` (merge `9bed2f09dd943bcb27cb90a05f7477cdb6df290a`).

Ran the actual HTTP handlers and built CLI against anonymous local Convex at `http://127.0.0.1:35800` / HTTP `:35801`. The same database held two real publisher records and ten skills with tied sort values, including two identical slugs, one null latest version, metadata/tags, and exact version `1.2.3+fixture.01`. The baseline restored the exact pre-PR production server/schema/CLI files; candidate used the merged PR source.

| Scenario | Before | After |
| --- | --- | --- |
| Updated and newest sorting | Five pages each; owner omitted | Five pages each; ten distinct owner/slug identities retained |
| Canonical decoding with null latest version | Exit 1 | Exit 0 |
| CLI against native registry | Exit 1 on null latestVersion | Exit 0; both same-slug owners printed; null version renders v? |
| Exact SemVer build metadata | Retained | Retained as 1.2.3+fixture.01 |
| Trending identity | Owner omitted | Ten owner/slug identities retained |
| CLI against a live legacy-shape loopback registry | Exit 0 | Exit 0 with bare-slug fallback |

Metadata and latest tags were also preserved. The complete runner passed. Temporary fixture/source changes were restored, the worktree is clean, and the disposable runtime stopped. Static, explicit schema/CLI/Convex typechecks, and independent source review passed; the exact merged head's executable GitHub CI was already green. This supplements the author's existing HTTP/CLI tests with native backend evidence.

PR #3586, opened 2026-09-04
fix: large ClawPack publishes time out after the upload succeeds

steipete:
## What Problem This Solves

Fixes an issue where publishing a large ClawPack (thousands of files, several MB) through `clawhub package publish` fails with `Network request failed: the request timed out. curl: (28) Operation timed out after 120002 milliseconds with 0 bytes received` even though the upload itself succeeds. After #3584 routed `@openclaw/matrix@2026.9.1` (5.5 MB, 1,843 files) through the upload-url flow, the OpenClaw release publish reached ClawHub but the publish request timed out (openclaw/openclaw run 33833703657, job 100904763980).

## Why This Change Was Made

`POST /api/v1/packages` unpacks the staged tarball and stores every extracted file with `ctx.storage.store` one at a time inside the Convex HTTP action before answering, so a 1,843-file package takes longer than the CLI's generic 120 s `UPLOAD_TIMEOUT_MS`. Two owner-level changes: `storeClawPackFiles` now stores files in batches bounded by 8 MiB and 16 files (order preserved), keeping the memory ceiling the old sequential comment protected while cutting wall time by roughly the batch width; and CLI form requests accept an optional `timeoutMs`, with the package publish POST using a 5 minute budget (the Vercel function budget in front of Convex) instead of the generic upload timeout. Staging uploads and other form requests keep 120 s.

## User Impact

Large plugin packages publish reliably from the CLI and from the reusable `package-publish.yml` workflow; smaller packages are unaffected. No API contract changes.

## Evidence

- `convex/httpApiV1.handlers.test.ts` "multipart ClawPack publish stores extracted files in bounded concurrent batches": on the sequential loop it fails with `expected 1 to be greater than 1`; with the fix it passes, with max in-flight ≤ 16 and file order preserved.
- `packages/clawhub/src/http.test.ts` and `http.bun.test.ts`: `timeoutMs` honored on the fetch runtime (300 s timer) and curl runtime (`--max-time 300`), default 120 s unchanged.
- `packages/clawhub/src/cli/commands/packages.test.ts`: the publish POST passes `timeoutMs: 300000`.
- Local: `bun run ci:static`, root `tsc --noEmit`, package `tsc` for `packages/schema` and `packages/clawhub`, full `convex/httpApiV1.handlers.test.ts`, `bun run --cwd packages/clawhub verify`, and Codex autoreview (results in the PR comment thread).
[...]

PR #3591, opened 2026-09-04
fix(publish): run the static moderation scan in the Node runtime

steipete:
## What Problem This Solves

Publishing `@openclaw/matrix@2026.9.1` (1,843 files after unpacking) now gets past the Vercel body cap (#3584) and the storage staging budget (#3586), but the trusted-publisher publish action still dies after ~220 s with a redacted Convex error. Live `convex logs` for the failing runs show the isolate for `packages:publishPackageForTrustedPublisherInternal` being killed while `runStaticPublishScan` reads and decodes every file of the ClawPack for the moderation scan. `publishPackageImpl` runs in the default Convex runtime, whose actions are capped at 64 MiB of memory; the full decoded text of a large plugin plus the scan's working set exceeds that, so the publish never reaches `insertReleaseInternal` and the CLI only sees an opaque failure.

## Why This Change Was Made

The scan itself is fine; it is running in the wrong runtime. Convex's Node.js runtime gives actions 512 MiB, so this moves the scan across that boundary and nothing else:

- new `convex/staticPublishScanNode.ts` (`"use node"`) exposes `runStaticPublishScanInternal`, which calls the unchanged `runStaticPublishScan` against the same storage ids.
- `convex/packages.ts` routes both callers (`publishPackageImpl` and `scanPackageReleaseStaticallyInternal`) through `runStaticPublishScanInNode`, a thin `runAction` hop using the existing `runActionRef`/`stripUndefinedForStoredAttempt` conventions. Scan semantics, reason codes, and the stored `staticScan` shape are untouched.
- `convex/_generated/api.d.ts` registers the new module (codegen needs a deployment; the edit matches what `convex codegen` emits).

The action hop is the smallest change that fixes the owner: the publish action stays in the Convex runtime (it needs `ctx.storage`, mutations, and scheduling exactly as before), and only the memory-heavy step crosses over, mirroring how `packageInspectorNode` already runs the package inspector.

## User Impact

Large plugins (thousands of files, tens of MB of source) publish again instead of failing with a redacted server error after several minutes. No CLI, API, or schema change; small packages take the same path with one extra internal action call.

## Evidence

- `convex/staticPublishScanNode.test.ts`: the Node action runs the real scan on the handed files and returns the verdict (reason code + engine version), proving the hop preserves scan output.
- `convex/packages.public.test.ts`: the `runAction` mock now dispatches by function name, running the real scan for the Node ref and answering the inspector otherwise. The pre-existing "scans plugin publishes and forwards scan status" case fails on the old mock (it returned the inspector result as the scan result) and passes with the dispatching mock, so the publish → Node scan → `insertReleaseInternal` wiring is covered end to end.
- Local: `bun x tsc --noEmit`, `vitest run convex/packages.public.test.ts convex/lib/staticPublishScan.test.ts convex/staticPublishScanNode.test.ts` (369 passed), `oxlint`/`oxfmt --check` on touched files, `bun run ci:static`.
- Live proof follows after deploy: re-dispatch of the `@openclaw/matrix@2026.9.1` publish from the openclaw release tag; the versions endpoint must return 200.

Production LOC: +~30 (new Node action module + the `runStaticPublishScanInNode` hop); justified by the runtime boundary (Convex vs Node memory ceiling) that cannot be expressed inside the existing action.

### DIFF
--- a/convex/_generated/api.d.ts
+++ b/convex/_generated/api.d.ts
@@ -212,6 +212,7 @@ import type * as skillsShMirrorPublic from "../skillsShMirrorPublic.js";
 import type * as skillsShMirrorVisibility from "../skillsShMirrorVisibility.js";
 import type * as skillsShPublicTestFixtures from "../skillsShPublicTestFixtures.js";
 import type * as stars from "../stars.js";
+import type * as staticPublishScanNode from "../staticPublishScanNode.js";
 import type * as statsMaintenance from "../statsMaintenance.js";
 import type * as telemetry from "../telemetry.js";
 import type * as tokens from "../tokens.js";
@@ -431,6 +432,7 @@ declare const fullApi: ApiFromModules<{
   skillsShMirrorVisibility: typeof skillsShMirrorVisibility;
   skillsShPublicTestFixtures: typeof skillsShPublicTestFixtures;
   stars: typeof stars;
+  staticPublishScanNode: typeof staticPublishScanNode;
   statsMaintenance: typeof statsMaintenance;
   telemetry: typeof telemetry;
   tokens: typeof tokens;
--- a/convex/packages.public.test.ts
+++ b/convex/packages.public.test.ts
@@ -1,6 +1,7 @@
 /* @vitest-environment node */
 
 import { getAuthUserId } from "@convex-dev/auth/server";
+import { type FunctionReference, getFunctionName } from "convex/server";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { sha256Hex } from "./lib/clawpack";
 import { verifyOpenClawPublishAuthorization } from "./lib/openClawPublishAuthorization";
@@ -11,6 +12,7 @@ import {
 } from "./lib/recommendationScore";
 import { buildPackageInventoryDigest } from "./lib/skills";
 import { buildDeterministicPackageZip } from "./lib/skillZip";
+import { runStaticPublishScan } from "./lib/staticPublishScan";
 import {
   backfillLatestPackageScanStatusInternal,
   backfillPackageReleaseScansInternal,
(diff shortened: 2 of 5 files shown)


steipete:
## Landing evidence

**Local proof (head d0b47830, Bun 1.4.0):**
- `bun x tsc --noEmit` clean.
- `vitest run convex/packages.public.test.ts convex/lib/staticPublishScan.test.ts convex/staticPublishScanNode.test.ts` → 3 files, 369 tests passed. The new `staticPublishScanNode.test.ts` runs the real moderation scan through the Node action and asserts the `suspicious.dynamic_code_execution` verdict comes back unchanged; the publish tests' `runAction` mock now dispatches the Node ref to the real scan, so `publishPackageImpl → runStaticPublishScanInNode → insertReleaseInternal` is exercised end to end.
- `bun run ci:static` (peers, audit, llms, format, oxlint type-aware, knip, workflow pins) clean.
- Independent Codex review of the diff: scoped-clean, no P0/P1 findings.
- Local run of the CI `moderation-malicious` lane (`bun run test:pw:local-auth -- --project=chromium e2e/local-auth/malicious-skill-ban-flow.pw.test.ts`, local Convex backend with Node 24 so `"use node"` actions deploy): `1 passed (50.6s)` on this head. The same lane on `main` (fa653d46) also passes locally; one earlier local attempt on this head hit the identical `start_push 408 → ENOENT build_deps/node_modules.zip` sequence seen in CI, confirming it is a nondeterministic local-deploy flake rather than the new module (the Node bundle adds no external packages; `convex.json` still lists only `sharp`).

**CI:** on 1125778f (run 33854679879) `unit`, `static`, `e2e-http`, `playwright-smoke`, and the Vercel preview are green; on d0b47830 the same set was green after one `static` rerun (first attempt was `bun audit` → `ConnectionClosed`, the registry outage also hitting other repos today). The `playwright-local-auth / *` lanes (`publish-generated-card`, `profile-context` on this head) fail on the pre-existing local Convex deploy flake (`start_push 408` → retry → `ENOENT build_deps/node_modules.zip`); the identical sequence fails on current `main` (run 33846708408, attempts 1–3) without this change. Follow-up to stabilize that lane is queued separately.

**Vercel preview:** the first deployment on d0b47830 failed (log not accessible from here; `main` fa653d46 and the previous merge had succeeded). The frontend half of the build passes locally, and the empty retrigger commit 1125778f got a successful preview deployment, which runs the same `convex deploy --preview-create` of this Node module against Convex cloud plus the frontend build, so the first failure was transient.

**Runtime proof before prod:** the merge SHA's `Deploy Test` deploys to the real Convex test deployment (Node runtime included) and runs the proof jobs; prod deploy is gated on its success. Final live proof: re-dispatch of the `@openclaw/matrix@2026.9.1` ClawHub publish from the openclaw release tag, verified via `GET /api/v1/packages/%40openclaw%2Fmatrix/versions/2026.9.1` → 200.

PR #3592, opened 2026-09-04
fix(ci): retry bun audit on transient registry failures

steipete:
## What Problem This Solves

The `static` CI job runs `bun audit` first. On 2026-09-04 the npm advisory endpoint dropped the connection (`ConnectionClosed: audit request failed`) on four separate runs (PR #3591 run 33849286983, main runs 33846708408 and 33856468453 attempts 1–2) and, from a developer machine, hung for over 90 s. Each time the whole 5-minute `static` job failed, `Deploy Test` never fired, and the production deploy of a release fix stayed blocked until a manual rerun happened to hit a good moment.

## Why This Change Was Made

A dropped or hung advisory request is not an advisory finding. `scripts/ci-audit.ts` now owns the audit step:

- one accepted-advisory list (the same 13 `--ignore` ids that lived inline in `package.json`);
- each attempt capped at 60 s via `spawnSync` `timeout`, with a killed child reported as `ETIMEDOUT`;
- up to 3 attempts with 5 s / 10 s backoff, retrying only when the output matches network-class failures (`ConnectionClosed`, `audit request failed`, `fetch failed`, `ECONN*`/`ETIMEDOUT`/`EAI_AGAIN`/`EPIPE`);
- real findings, or any other non-zero exit, fail immediately with the audit's own exit code; exhausted retries fail with the last exit code.

Worst case is about 3.3 minutes, inside the job's 5-minute timeout. The gate stays fail-closed: an outage that lasts through all retries still fails the job.

## User Impact

CI only. Fewer red `static` jobs from registry blips; no change to what the audit accepts.

## Evidence

- `scripts/ci-audit.test.ts`: classifier (connection drop and timeout → transient; findings → fatal), ignore-list wiring, retry-until-success with backoff timings, immediate failure on findings, exhausted budget returns the last exit code (5 tests, all passing locally).
- Local `bun x vitest run scripts/ci-audit.test.ts`, `oxlint --type-aware` on both files, `bun run deadcode:ci` clean; independent Codex review of the diff: scoped-clean.
- The same outage will exercise this on the PR's own `static` job.


steipete:
## Landing evidence

**Local (head fa46e703):** `bun x vitest run scripts/ci-audit.test.ts` → 5 passed; `bun run ci:types-build` clean (the first head, 04a0f9e6, failed `types-build` on an untyped `sleep` mock parameter, fixed in fa46e703); `oxlint --type-aware` and `oxfmt --check` on both files clean; `bun run deadcode:ci` clean; independent Codex review of the diff: scoped-clean.

**CI (run 33861954740):** `static` passed on the first attempt with the wrapper in place, `types-build`, `e2e-http`, `playwright-smoke`, and Vercel preview green. Attempt 1 `unit` failed on the pre-existing `scripts/ui-proof-backend.test.mjs` process-tree flake (same assertion failed on `main` run 33846708408 earlier today, unrelated to this change; follow-up queued); attempt 2 green. The `playwright-local-auth / *` lanes hit the known local Convex deploy flake and passed on rerun.

**Behavior:** a dropped or hung advisory request is retried up to 3 times with a 60 s cap per attempt (worst case ≈ 3.3 min, inside the job's 5-minute timeout); real findings and any other non-zero exit fail immediately with the audit's own exit code, so the gate stays fail-closed.

PR #3593, opened 2026-09-04
fix: sanitize static publish scan metadata

morajabi:
## Summary

- normalize package metadata before crossing the Convex action boundary into the Node static scanner
- add a regression assertion for a valid plugin manifest containing a JSON Schema `$schema` key

## Why

Publishing an exact ClawPack with a channel config schema currently fails before release insertion with:

`Field name $schema starts with a '$', which is reserved.`

The package publish path already sanitizes extracted manifests before database storage, but the new Node-runtime scan hop receives the raw metadata first. This applies the same existing `toConvexSafeJsonValue` normalization at that boundary. Package files remain unchanged and are still scanned from storage; only the duplicate structured metadata projection is normalized.

## Tests

- `bunx vitest run convex/packages.public.test.ts --testNamePattern 'scans plugin publishes and forwards scan status to insertReleaseInternal'`
- `bun run ci:static`
- `bun run ci:types-build`
- `npx --yes -p node@24 -c 'bun run ci:unit'` (6,334 passed, 3 skipped)

### DIFF
--- a/convex/packages.public.test.ts
+++ b/convex/packages.public.test.ts
@@ -12935,6 +12935,14 @@ describe("packages public queries", () => {
           categories: ["security"],
           topics: ["Manifest Topic"],
           contracts: { tools: ["demoTool"] },
+          channelConfigs: {
+            inline: {
+              schema: {
+                $schema: "http://json-schema.org/draft-07/schema#",
+                type: "object",
+              },
+            },
+          },
         }),
       ],
       [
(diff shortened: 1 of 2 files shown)


steipete:
Thanks for identifying the scan metadata boundary. This is covered on main by #3594 (3aba7d1b0088e4e5bc072cbda4407c2248e7862e): metadata crosses the Convex-to-Node boundary as JSON text and is decoded before scanning, preserving `$schema` and nested channel configuration keys without renaming them.

I checked the current caller, Node action validator/decoder, and regression coverage. #3594 also records a successful production package publication through this path. Closing this superseded candidate; no additional metadata-key sanitization is needed.

PR #3664, opened 2026-09-11
feat(api): add plugin marketplace overview snapshot

Patrick-Erichsen:
## Summary

- add a public, cacheable `GET /api/v1/plugins/overview` endpoint for the marketplace home
- return canonical categories plus the deduplicated top 8 Featured, Trending, and official-first/download-sorted plugins per category
- keep search, category expansion, and full traversal on the existing paginated endpoint

## Why

The Control UI home previously had to issue one request per shelf and traverse catalog cursors before ordinary categories appeared. The overview endpoint moves that bounded fanout behind one cacheable HTTP response without downloading the complete catalog.

## Performance proof

Using the same local Convex backend and fresh isolated browser contexts:

| Control UI | Median until ordinary categories render | Initial catalog RPCs |
| --- | ---: | ---: |
| unchanged main | 1,479 ms | 8+ |
| overview consumer | 509 ms | 1 |

That is 65.6% lower latency / 2.91x faster in the controlled comparison. The original production-backed failure took 110.108 s to expose the first ordinary category because it traversed 20 cursor pages.

## Validation

- `VITE_CONVEX_URL=https://example.invalid bun run ci:unit` (6,611 passed, 3 skipped)
- `bun run ci:types-build`
- `bun run ci:packages`
- `bun run ci:static`
- live local HTTP response: 200 with canonical categories and shared-cache headers
- structured review: clean


Patrick-Erichsen:
Addressed the latest ClawSweeper findings at `e9ddacda1d`:

- Sparse category shelves now use the existing family/official/category/download indexes through a bounded overview query, so the snapshot no longer drops matches beyond the generic scan window.
- Deduplicated entries carry zero-based `featuredRank` and `trendingRank`, preserving each shelf's independent order.
- Added regressions for sparse official/category selection and overlapping Featured/Trending membership.

Validation on the exact local head: focused catalog suite 827/827, `ci:static`, `ci:types-build`, `git diff --check`, and structured autoreview clean.

Patrick-Erichsen:
Addressed the latest ClawSweeper findings at `9a0bfa1641`: family-less official-first pages now stay on the stable multi-family cursor path, and Trending filters to plugin families before consuming its page limit. Added failing-before/passing-after regressions for both cases. Exact local validation: 829 focused tests, static, types/build, format/diff, and structured autoreview all clean.

Issue #3678, opened 2026-09-11
GitHub import unzips the archive before size limits


SebTardif:
## Summary

Authenticated GitHub import unzips the downloaded repo archive with `unzipSync` and no filter, then checks file count, per-file size, and total unzipped size after inflate. The download cap is 25MB compressed. A highly compressible 25MB zip can expand far past the 80MB unzipped budget inside the Convex action before those post-checks run.

This is a source-level DoS on an authenticated import action. I did not upload a zip bomb to production.

## Verification / reproduction

1. Read `convex/githubImport.ts` `unzipToEntries`.
2. Confirm `unzipSync(zipBytes)` has no `filter`.
3. Compare with `convex/githubSkillSync.ts` `createZipEntryLimitFilter`, which enforces `originalSize` before inflate.
4. `fetchGitHubZipBytes` only caps the compressed download.

A local `bun` probe of `__test.unzipToEntries` against a zip whose uncompressed `originalSize` exceeds the single-file cap is on branch `SebTardif:fix/github-import-unzip-limits`.

## Actual

- Size limits run after fflate inflates the archive.
- A 25MB compressed GitHub zip can exceed the 80MB unzipped budget in the action.

## Expected

- File count and `originalSize` limits apply in an unzip filter before inflate, matching skill-sync.
- Invalid / mac-junk paths stay skipped as they are today.

## Evidence

Source: `convex/githubImport.ts` `unzipToEntries`. Sibling already fixed in `convex/githubSkillSync.ts`.

Proposed patch (no upstream PR yet; barnacle author cap is 10):

https://github.com/openclaw/clawhub/compare/main...SebTardif:fix/github-import-unzip-limits

This is not the same bug as GitHub import blob cleanup (PR #3612).

## Security / privacy

Authenticated zip-bomb DoS of a Convex action. Not a claimed unauthenticated remote vuln. No exploit archive is attached.

## Acceptance criteria

- [ ] `unzipToEntries` rejects oversized `originalSize` / file count before inflate.
- [ ] Valid small archives still import.
- [ ] Skill-sync filter behavior (mac junk, path normalize) is preserved.

This tracker stays open until the filter lands on `main`, including if a later PR is closed.

SebTardif:
Proposed patch is going up now that PR #3613 is closed (barnacle slot). This issue stays open if that new PR is closed without landing.

Patrick-Erichsen:
[...]

PR #3679, opened 2026-09-11
fix(import): cap GitHub zip expansion before inflate

SebTardif:
## What Problem This Solves

Authenticated Convex GitHub import (`previewGitHubImport`, `previewGitHubImportCandidate`, `importGitHubSkill`) calls `unzipSync(zipBytes)` with no filter, then checks file count / single-file / total size after the archive has already been inflated.

`fetchGitHubZipBytes` only caps the compressed download at 25MB. A highly compressible 25MB GitHub zip can expand far past the 80MB unzipped budget inside the Convex action before those post-checks run.

Sibling `convex/githubSkillSync.ts` already uses `createZipEntryLimitFilter` so `originalSize` is enforced before inflate. This change applies the same filter to `githubImport.ts` `unzipToEntries`.

## Evidence

Live `bun` against production `unzipToEntries` with a zip whose uncompressed `originalSize` is 10MB+1:

```console
$ bun -e "import { zipSync } from 'fflate'; import { __test } from './convex/githubImport.ts'; ..."
THREW ConvexError: Repo archive contains a file that is too large
```

Same-repo: `githubSkillSync` unzip filter (`createZipEntryLimitFilter`). F006 [#3612](https://github.com/openclaw/clawhub/pull/3612) is the same file's blob-store leak, not unzip bounds.

## Real behavior proof

- **Behavior or issue addressed:** GitHub import unzip rejects oversized entries from zip headers before fflate inflates them.
- **Real environment tested:** Windows 11, bun 1.4.1, worktree `C:\tmp\wt-ch-f012` at `786ce7d`.
- **Exact steps or command run after this patch:** Built an in-memory zip with `zipSync` containing `model.bin` of length `10 * 1024 * 1024 + 1` and called `__test.unzipToEntries`.
- **Evidence after fix:** terminal output `THREW ConvexError: Repo archive contains a file that is too large`.
- **Observed result after fix:** Import unzip fails closed on the single-file cap during the filter callback instead of inflating first and skipping later.
- **What was not tested:** A 25MB highly compressible archive on live Convex. File-count and 80MB total caps use the same filter and are not separately exercised in this live command.

## Tracker

Ref #3678

That issue stays open if this PR is closed without landing on main.

### DIFF
--- a/convex/githubImport.test.ts
+++ b/convex/githubImport.test.ts
@@ -1,5 +1,6 @@
 /* @vitest-environment node */
 import { generateKeyPairSync } from "node:crypto";
+import { zipSync } from "fflate";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { internal } from "./_generated/api";
 import { __test } from "./githubImport";
@@ -84,6 +85,15 @@ describe("githubImport", () => {
       "demo-repo/skill/SKILL.md",
       "demo-repo/skill/notes.md",
     ]);
+  });
+
+  it("rejects oversized files so they are not inflated past import limits", () => {
+    const zip = zipSync({
+      "demo-repo/skill/SKILL.md": new TextEncoder().encode("# Demo\n"),
+      "demo-repo/skill/model.bin": new Uint8Array(10 * 1024 * 1024 + 1),
+    });
+
+    expect(() => __test.unzipToEntries(zip)).toThrow(/file that is too large/i);
   });
 
   it("rejects a public repo owned by another GitHub account before repo lookup", async () => {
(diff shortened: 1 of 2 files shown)


Patrick-Erichsen:
Closing as superseded by the merged [#3711](https://github.com/openclaw/clawhub/pull/3711), commit `6f92d9e12b33717d16101ca4d3776d9878da3846`.

Thank you for the pre-inflation hardening proposal. The landed follow-up retains that approach and fixes the compatibility regression where an unrelated oversized file caused rejection of an otherwise valid skill import. The same three real ZIP regressions were compared against main, this proposal, and the landed fix; current hosted CI is green. Your branch has not been modified.

PR #3691, opened 2026-09-14
fix: show published versions after pending releases

goutamadwant:
Closes #3687. Refs #3087. CI startup repair: #3710.

## What Problem This Solves

Fixes an issue where users opening plugin version history receive HTTP 500 when newer releases are unpublished. Production logs confirm the failing query for Brave, Slack, and Tokenjuice.

## Why This Change Was Made

The shared release query filters before one native Convex pagination call. It retains the 1–200 item target, existing scan limits, legacy releases, and native cursor metadata. This read-only change adds no state writer.

## User Impact

Published version history loads while pending, blocked, and deleted releases stay hidden. Both direct queries and HTTP clients use the same fixed owner.

## Evidence

[Real Convex HTTP and browser proof](https://github.com/openclaw/clawhub/blob/qa-artifacts/clawhub-ui-proof/pr-3691/2026-09-15/native-public-report.md): baseline HTTP 500 becomes 200 with the same synthetic fixture. Cursor continuation recovers every published and legacy version. A direct 250-item request returns 200, then two remaining releases.

Checks: 836 targeted tests; 6,700 unit tests plus coverage; `ci:static`, `ci:types-build`, and `ci:e2e-http`. Independent source review found no actionable issues.

## Compatibility

No schema or response-shape change. Native continuation and split metadata survive. Convex owns subscription invalidation; no new cache.

Disclosure: AI was used to understand the codebase and review the fix.

PR #3710, opened 2026-09-15
fix: disable scheduled jobs before local browser tests start

obviyus:
Prerequisite for #3691.

## What Problem This Solves

Fixes an issue where local browser tests start production background jobs before their isolated environment is configured, adding work during builds and tests.

## Why This Change Was Made

Start the backend without publishing, configure its environment, then publish once. A persistent launcher owns the backend process group through cleanup; a separate log tail preserves diagnostics.

## User Impact

Local browser tests start without production cron jobs and stop their backend descendants after success, bootstrap failure, or parent disconnection.

## Evidence

[Native startup and lifecycle proof](https://github.com/openclaw/clawhub/blob/qa-artifacts/clawhub-ui-proof/ci-startup-e0e82417/public-report.md): 36 registered jobs become zero. The unchanged moderation test passes in 36.6 seconds, with exit 0 and no surviving processes.

Three lifecycle cases pass; restoring early launcher exit makes the regression fail. Also passed: 29 focused tests, 6,695 unit tests with coverage, `ci:static`, and `ci:types-build`.

## Compatibility

No production behavior or timeout changes. This test runner requires Linux/macOS process groups and rejects Windows before changing local state. Hosted runner sizing is unchanged.


obviyus:
Maintainer decision: Linux/macOS-only support is accepted for this local-auth browser-test runner. Hosted CI uses Ubuntu, and native Windows now receives the documented error before dependency installation or local-state isolation. This decision applies to the test runner in this PR.

@clawsweeper re-review

obviyus:
**Proof:** The actual local browser-test runner starts without production cron jobs, passes the unchanged moderation scenario, and stops its backend processes cleanly.

- **Bug:** Background jobs started before the isolated test environment was configured.
- **Fix:** Configure the backend before its first publication and retain ownership through cleanup.

| Before | After |
| --- | --- |
| 36 scheduled jobs; 35 executions before build | Zero scheduled jobs and executions; clean runner exit |

Issue #3733, opened 2026-09-16
Skill --force install rolls back after leftover backup delete fails


SebTardif:
## What happens

`clawhub skill install --force` (and update) stages the new tree, swaps it into place, then deletes the old backup. That last delete sits in the same `try` as the swap. `force: true` only swallows a missing path. If delete throws `EBUSY` or `EPERM` (Windows antivirus, a locked file), the catch removes the new install and restores the backup.

The upgrade already succeeded. Cleanup of the leftover backup then undoes it.

## How to see it

1. Install a skill so the target directory exists.
2. Run `clawhub skill install <slug> --force`.
3. If deleting `.slug.backup-*` fails after the swap, the command errors and the previous tree is restored.

From the code: `installSkillWithOptionalStaging` in `packages/clawhub/src/cli/commands/skills.ts`. Introduced in [`7d69b337c7a7`](https://github.com/openclaw/clawhub/commit/7d69b337c7a744224e800b7d92a86d57f757b883) (`fix: make skill updates rollback safe`).

## Expected

Once the new tree is in place, backup delete is best-effort. A leftover backup directory is fine. Rolling back a finished swap is not.

SebTardif:
Opened PR https://github.com/openclaw/clawhub/pull/3734 for this tracker (`Ref` only).

PR #3734, opened 2026-09-16
fix: skill --force install rolls back when leftover backup delete fails

SebTardif:
Ref #3733

## What Problem This Solves

Fixes: a finished `clawhub skill install --force` (or update) is rolled back when leftover backup delete fails.

## User Impact

User impact: a successful skill upgrade stays installed even if the old backup directory cannot be deleted.

## Why This Change Was Made

Staging still rolls back when the swap itself fails. After `rename(stage, target)` succeeds, backup delete is best-effort. `force: true` only ignores a missing path, so `EBUSY`/`EPERM` used to enter the rollback branch and replace the new tree with the old one.

Introduced in [`7d69b337c7a7`](https://github.com/openclaw/clawhub/commit/7d69b337c7a744224e800b7d92a86d57f757b883).

## Evidence

Public command: `clawhub skill install <slug> --force` when the target already exists.

Before: leftover backup delete lived in the same `try` as the swap. A locked backup made the command reject and restore the old tree.

After: the same command keeps the new tree. A Node script that performs the production swap against real directories, then throws `EBUSY` on backup delete, leaves `SKILL.md` as `new`.

```console
$ node /tmp/f013-live-proof.mjs
target_body="new\n"
OK: finished swap kept after backup delete failed
```

## Real behavior proof

- **Behavior or issue addressed:** After a finished skill install swap, a leftover backup that cannot be deleted no longer rolls the new tree back.
- **Real environment tested:** macOS, Node, patched `packages/clawhub` at `/tmp/clawhub-f013`.
- **Exact steps or command run after this patch:**

  ```bash
  node /tmp/f013-live-proof.mjs
  ```

- **Evidence after fix:** terminal output from the patched tree:

  ```console
  $ node /tmp/f013-live-proof.mjs
  target_body="new\n"
  OK: finished swap kept after backup delete failed
  ```

  Production helper now deletes the backup after the swap try, and rollback requires `backupCreated && !stageMoved`.

- **Observed result after fix:** The new `SKILL.md` stays `new` when backup delete throws. The install command no longer renames the backup back onto the target.
- **What was not tested:** A live registry install on Windows with a real antivirus lock on the backup directory.


steipete:
The updated head `37d3485cc646a029305f7b541cf0dae92e2d2c53` passes static, unit, packages, types/build and HTTP e2e. Its public browser smoke failure is the same pre-existing missing skills.sh fixture failure seen on main, confirmed in run 35688489751.

#3795 fixes that CI root cause by moving the existing layout regression to a seeded local backend. Its exact head has green static, unit, package, type/build, HTTP, public-browser and local-auth checks, and the matching remote gates passed. It is awaiting the requested owner review. This install fix remains open until the CI repair can land and the required checks are green.

PR #3750, opened 2026-09-16
fix(convex): restore comments query compatibility

Patrick-Erichsen:
## Summary

- restore the historical public `comments:listBySkill` Convex query as a read-only compatibility tombstone
- return the empty list stale clients already understand, without reading retired comment data
- keep comment writes and the retired comments UI/data path disabled
- document the compatibility boundary and cover the exact public function name through `convex-test`

Fixes [CLAW-858](https://linear.app/my-openclaw/issue/CLAW-858/restore-the-commentslistbyskill-function-expected-by-clawhub-clients).
Sentry: [CLAWHUB-CONVEX-5](https://openclaw-ba.sentry.io/issues/CLAWHUB-CONVEX-5)

## Root cause

PR #2658 intentionally removed the comments module after the UI had been disabled, including the public `comments:listBySkill` function. Production still receives calls to that exact historical name. The deleted client treated an empty array as the normal no-comments state, so a bounded query tombstone removes the missing-function failure without reviving comment storage, reads, or writes.

## Best-fix verdict

**Best.** The compatibility function sits at the missing public API boundary named by the production error and preserves the retired product decision.

Alternatives considered:

- restore the old comments query/data reads: rejected because the tables/data path are retired
- restore `add`, `remove`, and `report`: rejected because no production evidence implicates those writes and restoring them would revive unsupported behavior
- update current clients only: rejected because current source and the current production bundle no longer call this API; this failure comes from stale callers

## Verification

- `bun run test convex/comments.test.ts` — 2 passed
- `env -u HTTP_PROXY -u HTTPS_PROXY -u ALL_PROXY NO_PROXY=127.0.0.1,localhost bun run ci:unit` — 534 files passed, 1 skipped; 7,057 tests passed, 3 skipped; coverage 86.69% statements / 78.28% branches / 87.97% functions / 89.33% lines
- `env -u HTTP_PROXY -u HTTPS_PROXY -u ALL_PROXY NO_PROXY=127.0.0.1,localhost bun run ci:types-build` — passed
- `bunx tsc --noEmit` — passed
- `.agents/skills/autoreview/scripts/autoreview --mode local` — clean, no accepted/actionable findings (0.97 confidence)

`bunx convex codegen` could not run in this isolated worktree because it has no `CONVEX_DEPLOYMENT`; the deterministic two-line generated API entry was updated directly and validated by the focused function-name test, full typecheck, and production build.

No UI proof: backend-only compatibility query; current clients expose no comments UI.


Patrick-Erichsen:
Closing this approach: skill comments are intentionally retired. We will not restore comments:listBySkill or a compatibility endpoint. Follow-up will investigate and remove any remaining supported callers or feature remnants; stale external calls are not a reason to reintroduce the API.

PR #3777, opened 2026-09-17
fix(ci): coalesce pending skills.sh synchronization runs

vincentkoc:
## What Problem This Solves

Fixes: hourly skills.sh synchronization requests accumulate in the production deployment queue while an earlier deployment or sync holds the shared lock.

## User Impact

Each ref keeps one sync workflow active or waiting for production, plus only the latest pending sync request. Scheduled and same-ref manual requests both coalesce; a newer request replaces an older pending request without cancelling the active sync or replacing a queued manual deployment.

This applies to runs using the new workflow definition. Already queued runs keep their original definition.

## Why This Change Was Made

The [sync workflow](https://github.com/openclaw/clawhub/blob/c6d3d937d4a159150b6493dc93cb423d60a04afd/.github/workflows/skills-sh-sync.yml) takes a distinct per-ref workflow group with `queue: single`, then the entire sync job takes the existing `deploy-production` group with `queue: max`. Both use `cancel-in-progress: false`. The [Deploy workflow](https://github.com/openclaw/clawhub/blob/96430c6f924c69d3083d33d8ef374c3c3f804531/.github/workflows/deploy.yml) is unchanged and never waits on the outer sync group, so the lock order has no cycle. Cleanup and proof upload remain inside the shared production lock.

This uses [GitHub's native concurrency contract](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency): one replaceable pending request for `single`, up to 100 pending jobs/workflows for `max`. Shared-group ordering follows when work starts waiting on that group, not dispatch time. Net operational YAML growth is five lines; application runtime code is unchanged.

## Evidence

Validated from main `96430c6f924c69d3083d33d8ef374c3c3f804531`; published head `c6d3d937d4a159150b6493dc93cb423d60a04afd` has the exact reviewed and tested four-file diff.

- The changed workflow regression fails against the original workflow, then passes with this fix.
- `bun run test scripts/skills-sh-sync-workflow.test.ts scripts/skills-sh-catalog/sync.test.ts scripts/skills-sh-catalog/sync.cli.test.ts`: 26 passed.
- `bun run ci:static`: passed.
- `bunx tsc -p packages/schema/tsconfig.json --noEmit` and `bunx tsc -p packages/clawhub/tsconfig.json --noEmit`: passed.
- `git diff --check` and the native autoreview secret scan: passed. Two independent source reviews passed under the equivalent manual-review exception in `AGENTS.md`; native model review could not start because its CLI launcher failed under the isolated environment (engine exit 127).
- Proof used a fresh frozen install with Node 24.18.0 and Bun 1.3.14. Hosted CI uses Bun 1.3.10 and remains the merge gate. No production workflow was dispatched for this change.

AI-assisted implementation; I reviewed the lock ordering and regression coverage and can maintain this change.


vincentkoc:
Maintainer decision on the remaining hosted-concurrency proof gap in [the exact-head review](https://github.com/openclaw/clawhub/pull/3777#issuecomment-5708308082): accepted for merge at `c6d3d937d4a159150b6493dc93cb423d60a04afd`.

[GitHub's documented concurrency contract](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency) supports one active workflow plus one replaceable pending request per ref in the outer sync group. The reviewed inner `deploy-production` group covers the entire sync job, including cleanup and proof upload, uses `queue: max` and `cancel-in-progress: false`, and remains shared with the unchanged deployment workflow. Deployment never acquires the outer sync group.

There is no hosted trace of these nested groups yet. The mitigation is to observe the first naturally scheduled post-merge sync and its production-lock behavior. We will not dispatch an artificial production run or cancel active work to manufacture that evidence. The per-ref bound applies to runs using the new definition; already queued runs retain their original definition.

The exact-head [CI run, attempt 1](https://github.com/openclaw/clawhub/actions/runs/35180458743) passed all 16 jobs. The review found no code or security defects. This decision resolves its remaining merge-risk item without changing the reviewed head or PR body.

